PR #12, opened 2019-10-02
starting USSD endpoint and starting models refactor

wangfowen:
super barebones USSD endpoint just to get started, and extracted out IpAddress and utils (based on http://exploreflask.com/en/latest/organizing.html) from models.py to start making it not just one huge file

am I doing this stuff right? what's the general work flow for making sure i'm not breaking stuff? i currently just load the admin dashboard then click around and see if anything crashes... doesn't seem optimal

### DIFF
--- a/app/server/api/ussd_api.py
+++ b/app/server/api/ussd_api.py
@@ -1,0 +1,55 @@
+from flask import Blueprint, request, make_response, jsonify
+from flask.views import MethodView
+
+ussd_blueprint = Blueprint('ussd', __name__)
+
+class ProcessKenyaUssd(MethodView):
+
+    """
+        USSD Entry method
+        Method: POST
+
+        @params sessionId - USSD session ID
+        @params phoneNumber - MSISDN initiating the USSD request
+        @params serviceCode - USSD code dialled to access the service eg. *701#
+        @params text - User input. The initial input is blank
+
+        Method returns a text response to be displayed to the user's screen
+        All responses mid-session begin with CON. All responses that terminate the session begin with the word END
+    """
+    def post(self):
+        post_data = request.get_json()
+
+        session_id = post_data.get('sessionId')
+        phone_number = post_data.get('phoneNumber')
+        user_input = post_data.get('text')
+
+        if phone_number:
+            """
+                TODO:
+                user = User.find_by(phone: UssdApi::Utilities.internationalize_phone(params[:phoneNumber], 'KE'))
+                ussd_request = UssdRequest.new(build_ussd_request_hash)
+                if ussd_request.valid?
+                    current_menu = UssdApi::MenuProcessor.process_request(ussd_request.attributes)
+                    ussd_session = create_or_update_session(ussd_request.attributes, current_menu)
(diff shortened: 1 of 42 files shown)


app/server/api/ussd_api.py
```diff
@@ -0,0 +1,55 @@
...
+        Method: POST
+
+        @params sessionId - USSD session ID
+        @params phoneNumber - MSISDN initiating the USSD request
+        @params serviceCode - USSD code dialled to access the service eg. *701#
+        @params text - User input. The initial input is blank
+
+        Method returns a text response to be displayed to the user's screen
+        All responses mid-session begin with CON. All responses that terminate the session begin with the word END
+    """
```
wangfowen: public API so I don't put a `required_auth` right?

app/server/models/ip_address.py
```diff
@@ -0,0 +1,43 @@
+from server import db, sentry, celery_app
```
wangfowen: extracted out of models.py

app/server/models/utils.py
```diff
@@ -0,0 +1,19 @@
+import datetime
```
wangfowen: extracted out of models.py

app/server/api/ussd_api.py
```diff
@@ -0,0 +1,55 @@
...
+            text = "test"
+        else:
+            #TODO: render plain: UssdMenu.find_by(name: 'exit_invalid_request').display_text
+            text = "test2"
+
+        return make_response(text, 200)
+
+
+ussd_blueprint.add_url_rule(
+    '/v1/ussd/kenya',
```
wangfowen: do I need to put v1 or is there already some prepending of api/v1, etc?

wangfowen: and looks like if I curl the endpoint I get 404:
```
127.0.0.1 - - [03/Oct/2019 00:17:29] "POST /api/v1/ussd/kenya HTTP/1.1" 404 -
```
am I missing something?

wangfowen: if i curl the endpoint i get 404:

```
127.0.0.1 - - [03/Oct/2019 00:17:29] "POST /api/v1/ussd/kenya HTTP/1.1" 404 -
```

am I missing something?

enjeyw:
> i currently just load the admin dashboard then click around and see if anything crashes... doesn't seem optimal

unfortunately that's probably the best way _for now_, as test coverage isn't quite there yet. A few more core tests and i think we'll be able to swap to running test watchers... does that seem right to u?

enjeyw:
> models.py to start making it not just one huge file

Ur structure is looking _soooo_ much cleaner. hopefully it doesn't kill the python imports lol

app/server/api/ussd_api.py
enjeyw: that doesn't sound right. what stops a malicious actor just POSTING with a fake session_id or phone number? Does Africa's talking have some sort of field to supply auth headers?

enjeyw:
CI/CD pipeline tests are failing, but i don't think that a problem with this specific PR. I'll look over today.

PR #13, opened 2019-10-03
Getting CI/CD Working

enjeyw:
This should fix the Circle CI build failure


enjeyw:
Hmmm not ready to merge yet, half of circle ci is failing

enjeyw:
[Get CI/CD Test working for new system](https://trello.com/c/MdiJHgXX/48-clean-up-aws-environment)

app/migrations/versions/9c233f12de2c_.py
```diff
@@ -29,6 +29,6 @@ def downgrade():
     # ### commands auto generated by Alembic - please adjust! ###
     op.add_column('transfer_account', sa.Column('owning_organisation_id', sa.INTEGER(), autoincrement=False, nullable=True))
     op.create_foreign_key('transfer_account_owning_organisation_id_fkey', 'transfer_account', 'organisation', ['owning_organisation_id'], ['id'])
-    op.drop_constraint(None, 'organisation', type_='foreignkey')
+    op.drop_constraint("fk_org_level_account", 'organisation', type_='foreignkey')
```
enjeyw: Not ideal, but had to go back and modify a migration file because it kept on dying during the docker-compose launch

test/unit/app/utils/test_user.py
```diff
@@ -4,17 +4,18 @@
...
-    GIVEN create_transfer_account_user function
-    WHEN called with first_name, last_name, phone
-    THEN assert one_time_code, transfer account exists,
-        transfer_account approval is equal to config
-    """
-    import config
-    assert create_transfer_account_user.one_time_code is not None
-    assert create_transfer_account_user.transfer_account is not None
-    assert create_transfer_account_user.transfer_account.is_approved is config.AUTO_APPROVE_TRANSFER_ACCOUNTS
+# REDACTED: USERS NOW HAVE MULTIPLE TRANSFER ACCOUNTS
```
enjeyw: this whole test doesn't really work on the new system because there's now multiple transfer accounts per user, though we should probably re-create a new equivalent test

enjeyw:
yasssssss

wangfowen:
nicee :D

app/manage.py
```diff
@@ -18,12 +18,8 @@ def run(self):
 
             print("~~~~~~~~~~ Searching for Master Address ~~~~~~~~~~")
 
-            if app.config['IS_USING_BITCOIN']:
-                master_address_type = "BITCOIN_MASTER"
-                master_address = app.config['BITCOIN_MASTER_WALLET_ADDRESS']
-            else:
-                master_address_type = "MASTER"
-                master_address = app.config['MASTER_WALLET_ADDRESS']
```
tristanhcole: @enjeyw Have we removed support for Bitcoin?

app/server/utils/user.py
```diff
@@ -219,7 +219,7 @@ def save_device_info(device_info, user):
     if device_info['serialNumber'] and not DeviceInfo.query.filter_by(serial_number=device_info['serialNumber']).first():
         # Add the device if the serial number is defined, and isn't already in db
         add_device = True
-    elif not device_info['serialNumber'] and not DeviceInfo.query.filter_by(unique_id=device_info['uniqueID']).first():
+    elif not device_info['serialNumber'] and not DeviceInfo.query.filter_by(unique_id=device_info['uniqueId']).first():
```
tristanhcole: @enjeyw Mobile still passes param as `uniqueID`. Need to update.

https://github.com/teamsempo/SempoMobileApp/blob/2d8f42b3e40ba173a6929b2c47838d3e2a7e1c1a/src/sagas/authSaga.js#L144

app/migrations/versions/9c233f12de2c_.py
wangfowen: how will that work for deploying to prod? can't rollback and re-migrate since you'd lose data right?

app/server/utils/blockchain_tasks.py
```diff
@@ -86,7 +114,11 @@ def get_token_decimals(token):
                                                       'function': 'decimals'
                                          })
 
-    return _execute_synchronous_task(decimals_sig)
+    if current_app.config['IS_TEST']:
```
wangfowen: 😬

yeahh very non-ideal to test this way. where is this run? unit tests, or integration tests?
if it's integration tests, it should be running its default behavior to really see the integration. if it's unit tests, you can make a mock of the worker at the point of usage to customize the return.

in ruby you can do something like: `expect(worker).to receive(:get_token_decimals).and_return(18)`
from a quick search, the closest thing i see is: https://realpython.com/python-mock-library/#managing-a-mocks-return-value
but you'd probably know better than me if there's something better

app/migrations/versions/9c233f12de2c_.py
wangfowen: also how do these work for migrating locally? do i have to drop my db and recreate?

app/manage.py
enjeyw: not properly yet, but I think we should deprecate - it doesn't fit our current strategic plans?

app/server/utils/blockchain_tasks.py
enjeyw: it's for integration tests. Currently it doesn't work because the response is supposed to be returned from the celery worker, but given all the tests are actually inside docker-compose, there's probably no reason we couldn't have them run.  I'll create a task for myself

Issue #11, opened 2019-10-02
Simplify external resource storage


tristanhcole:
**Problem:** The model is currently spitting out `UploadedDocument` and `UploadedImage` into two close to identical tables.

**Solution:** Merge into a single model (i.e. `UploadedResource`). 

New table should also probably be associated with an organization ID(s) or have a private base model for added security from accidental data sharing between organizations/users.

PR #14, opened 2019-10-03
refactoring out things from the mega models.py

wangfowen:
tried to extract out User from models.py and then one thing led to another to fix dependencies... and now here we are... this should be merged asap if it looks okay, otherwise will easily result in merge conflicts
aside from the comments below, this is entirely moving files around and adding/fixing imports


app/server/models/transfer.py
```diff
@@ -0,0 +1,611 @@
+import datetime, enum
```
wangfowen: this file is still massive... couldn't split it out because of circular dependencies. how do we fix that? ideally each model has its own file so it's easier to think through it

app/server/models/transfer.py
```diff
@@ -0,0 +1,611 @@
...
+        if not self.is_approved:
+            self.is_approved = True
+
+            if self.is_beneficiary:
+                disbursement = self.make_initial_disbursement()
+                return disbursement
+
+    def make_initial_disbursement(self, initial_balance=None):
+        #TODO(refactor): circular dependency
+        from server.utils.credit_transfers import make_disbursement_transfer
```
wangfowen: what's the purpose of utils.user and utils.credit_transfer? what distinguishes helper functions in those vs functions on the models for them respectively? those were points of contention for circular dependencies. see the other "TODO(refactor)"s. is this the correct way to fix them?

app/server/models/transfer.py
wangfowen: is this an okay name / grouping for what is the relation between these various models in the meantime?

app/server/models/upload.py
```diff
@@ -0,0 +1,33 @@
...
+    user_id = db.Column(db.Integer, db.ForeignKey(User.id))
+
+    @hybrid_property
+    def image_url(self):
+        return get_file_url(self.filename)
+
+class UploadedDocument(ModelBase):
+    __tablename__               = 'uploaded_document'
+
+    kyc_application_id = db.Column(db.Integer, db.ForeignKey('kyc_application.id'))
```
wangfowen: how come here the foreign key is a string whereas in `UploadedImage` it's a class reference? is class reference preferable for any reason? they seem like potential for circular dependencies otherwise

app/server/models/user.py
```diff
@@ -0,0 +1,466 @@
...
+    RoleNotFoundException,
+    TierNotFoundException,
+    NoTransferCardError
+)
+from server.constants import (
+    ACCESS_ROLES
+)
+
+# TODO(refactor): circular dependency when in util.user
+def get_transfer_card(public_serial_number):
```
wangfowen: don't know if this is correct fix. this was in util.user originally, but then user was importing util.user which was importing user...

app/server/utils/credit_transfers.py
```diff
@@ -530,18 +532,19 @@ def transfer_credit_via_phone(send_phone, receive_phone, transfer_amount):
...
+    receive_user = User.query.filter_by(phone=receive_phone).first()
     if receive_user is None:
         return {'status': 'Fail', 'message': "Can't send to phone number: " + send_phone}
 
     if send_user.transfer_account.balance < transfer_amount:
         return {'status': 'Fail', 'message': "Insufficient Funds"}
 
-    transfer = make_payment_transfer(transfer_amount, token, send_user, receive_user)
+    # TODO(refactor): removed token as second argument since it doesn't exist... where was it before?
+    transfer = make_payment_transfer(transfer_amount, send_user, receive_user)
```
wangfowen: this code was breaking because `token` doesn't exist. i didn't touch that code, don't know how it was working before

app/server/models/transfer.py
wangfowen: actually, i'm kinda confused now when an import causes an error for dependency vs is okay with it. models.py before was working just fine but there must have also been loops with the utils right?

enjeyw:
Hahaha yeah welcome to the import hell of python.

I’ll look over this properly in the morning, but the string vs class references that u commented on earlier are basically designed to work around the reference errors: you can only reference “class B” inside a method of “class A” if class B is declared before A. OTOH u can reference a string anywhere. So where we use the class reference, it’s generally where the class has been declared higher up in the code. So things will break/work depending on ur order of imports in the init file. 

Util files tend to be particularly bad if they reference other classes in the module, because they totally spaghetti-fy ur import order.

app/server/models/transfer.py
enjeyw: This is looking good. In my mind, the core models are:
- User
- TransferAccount
- CreditTransfer
- (Maybe) Organisation

Most other models are here to support those four.
So ideally I think we'd seperate out "TransferAccount" and "CreditTransfer" into two different files if dependencies permit, though i can imagine it's non trivial

app/server/models/transfer.py
enjeyw: To add to my comments at the bottom of the thread:

> What distinguishes helper functions in those vs static functions on the models? 

In short, I built them out there to stop the models file from getting way huge, but your refactor is a better solution.
I don't _love_ doing database actions (such as db.session.add() )from inside database models themselves, but tbh I'm not convinced there's any good logic behind that; the only db action that I know should definitely be extracted out is session commit.
It may actually be reasonable to shift those into the models themselves, now especially if they reduce import errors.

> adding an import inside the function instead of at the top. that's what i read somewhere but will this cause an error when the function itself is run?

Yeah in-function imports are pretty much a last resort. My Pycharm Linter actually yells at me if i try and do them lol.

app/server/models/user.py
enjeyw: This is the best overview i've ever seen on this issue:
https://stackoverflow.com/questions/22187279/python-circular-importing

I've used that process in a few places, for example in /utils/credit_transfers.py

`    transfer = models.CreditTransfer(amount=transfer_amount,
                                     token=token,
                                     sender_transfer_account=outbound_transfer_account,
                                     recipient_user=receive_account,
                                     transfer_type=models.TransferTypeEnum.DISBURSEMENT, uuid=uuid)`

That being said, i do wonder if minimising the use of helper function files now that we have a few seperate models files may actually help this in a far easier manner.

app/server/utils/credit_transfers.py
enjeyw: weird, must have slipped through. my bad

app/server/models/upload.py
tristanhcole: No reason. Generally, a string is used if it is referencing a class before it is created in the model file.

`UploadedDocument` and `UploadedImage` should be merged into a single class. Issue here: https://github.com/teamsempo/SempoBlockchain/issues/11

app/server/models/user.py
wangfowen: got it, so should we start using `import x` instead of `from x import y`?

PR #15, opened 2019-10-04
Fiat ramp

tristanhcole:
FiatRamp model to handle cash in and cash out ability.


tristanhcole:
[POLi Payments](https://trello.com/c/lHi4uzqx/46-poli-payments)

tristanhcole:
@enjeyw using `make_disbursement_transfer()` as the internal payment. Would be keen to know your thoughts on how we label and set up a float wallet.

POLi payments flow works except for our internal payment.

TODO's: https://trello.com/c/lHi4uzqx/46-fiatramp-poli-payments

tristanhcole:
Core models and flow built.

wangfowen:
@tristanhcole is this still a WIP or does that mean this PR is ready for review?

app/migrations/versions/90b392e41d9f_.py
```diff
@@ -0,0 +1,24 @@
+"""empty message
```
wangfowen: can these be removed?

app/server/api/poli_payments_api.py
```diff
@@ -0,0 +1,87 @@
...
+                )
+
+            except PoliPaymentsError as e:
+                response_object = {
+                    'message': str(e)
+                }
+                return make_response(jsonify(response_object)), 400
+
+            fiat_ramp = db.session.query(FiatRamp).filter(
+                FiatRamp.payment_metadata["poli_link"].astext == 'https://poli.to/'+poli_link
```
wangfowen: this seems a bit fragile. if poli ever changes their url in any way this would break. also it's a bit confusing that fiat ramp's metadata's "poli_link" is not the same as "poli_link" from the response

app/server/me_api/misc.py
```diff
@@ -282,4 +286,106 @@ def post(self):
...
+        reference = request.args.get('reference')
+        if reference:
+
+            fiat_ramp = FiatRamp.query.filter_by(payment_reference=reference).first()
+            if fiat_ramp is None:
+                return make_response(jsonify({'message': 'Could not find payment for reference {}'.format(reference)})), 400
+
+            poli_link = fiat_ramp.payment_metadata['poli_link']
+            try:
+                get_poli_link_status_response = get_poli_link_status(poli_link=poli_link.split('/')[-1])
```
wangfowen: again here, a bit confusing where poli_link is different from poli_link. seems we only ever use that last part, so maybe we should just store that as poli_link? or name the last section "poli_link_id"?

wangfowen:
so this doesn't hook up poli to withdrawal/deposit yet, right? it's just the core of how a fiat ramp would look and the APIs for poli?

app/migrations/versions/90b392e41d9f_.py
tristanhcole: I don't think so. It's a merge of two separate database migration heads (i.e. created when running `python manage.py db merge heads`)

app/server/api/poli_payments_api.py
tristanhcole: Good point. Will review.

I'll change the variable names at least.

app/server/me_api/misc.py
tristanhcole: Yes, will update

tristanhcole:
> so this doesn't hook up poli to withdrawal/deposit yet, right? it's just the core of how a fiat ramp would look and the APIs for poli?

ahhhh it's mostly done. the deposit function works and model is built. @enjeyw needs to build or advise on how we are planning on setting up internal Sempo Float wallets to ensure the deposit/withdrawal is correct

https://github.com/teamsempo/SempoBlockchain/pull/15/files/e9b935cb8f0189318ef153e8d3d3efc0c1082c01#diff-8431bc6a24b5ece5522f43428e78709fR55

app/server/api/poli_payments_api.py
```diff
@@ -0,0 +1,87 @@
...
+            if fiat_ramp is None:
+                return make_response(jsonify({'message': 'Could not find payment for POLi Link: {}'.format(poli_link)})), 400
+
+            status = get_poli_link_status_response['status']
+
+            if fiat_ramp.payment_status == FiatRampStatusEnum.PENDING:
+                if status == 'Completed':
+                    # Complete: "The full amount has been paid"
+                    individual_recipient_user = User.query.execution_options(show_all=True).get(fiat_ramp.authorising_user_id)
+                    deposit = make_deposit_transfer(
```
tristanhcole: @wangfowen This is a deposit transfer on successful PoLI payment

tristanhcole:
> @tristanhcole is this still a WIP or does that mean this PR is ready for review?

Core FiatRamp model is built and won't change. PoLI payments works in demo env.

Float wallets are WIP but I don't think should stop this being reviewed/merged.

app/client/components/transferAccount/transferAccountManager.jsx
```diff
@@ -35,7 +35,7 @@ class TransferAccountManager extends React.Component {
     this.state = {
         action: 'select',
         transfer_type: 'ALL',
-        create_transfer_type: 'WITHDRAWAL',
+        create_transfer_type: 'RECLAMATION',
```
enjeyw: isn't it weird how "reclaim" is spelt with an "i" but "reclamation" isn't?! stoopid english

app/server/utils/poli_payments.py
```diff
@@ -0,0 +1,106 @@
+import random, string, datetime, requests, base64
+from flask import current_app
+
+
+class PoliPaymentsError(Exception):
+    pass
+
+
+def generate_poli_host(path=None):
```
enjeyw: as per owen's notes, perhaps `generate_poli_endpoint` given it extends the host?

enjeyw: or `generate_complete_poli_path`?

app/server/utils/poli_payments.py
```diff
@@ -0,0 +1,106 @@
...
+
+    response = requests.get(url=url, headers=headers)
+
+    if response.status_code == 200:
+        result = dict(status=response.json())
+
+    else:
+        raise PoliPaymentsError(response.json()['ErrorMessage'])
+
+    return result
```
enjeyw: does this return a dict even when there's a non 200 status code?

app/server/api/poli_payments_api.py
```diff
@@ -0,0 +1,87 @@
...
+    def post(self):
+        """
+        POLi Payments Webhook that provides a token to lookup a POLi link and check status.
+
+        :return:
+        """
+
+        if request.data:
+            try:
+                response = get_poli_link_from_token(token=request.data.decode())
```
enjeyw: I'm confused as to what this is doing... it's the poli webhook that then loops back to make a request from.... the original poli server? why do they need this?

app/server/me_api/misc.py
```diff
@@ -282,4 +286,106 @@ def post(self):
...
+                }
+                return make_response(jsonify(response_object)), 400
+
+            status = get_poli_link_status_response['status']
+
+            if fiat_ramp.payment_status == FiatRampStatusEnum.PENDING:
+                if status == 'Completed':
+                    # "The full amount has been paid"
+                    individual_recipient_user = User.query.execution_options(show_all=True).get(fiat_ramp.authorising_user_id)
+                    deposit = make_deposit_transfer(
```
enjeyw: Given it's not idempotent, I think this should be a `POST` or a `PUT` request rather than a `GET`...

enjeyw: oo the below is a `post`, so this definitely feels like a `put`...

app/server/me_api/misc.py
```diff
@@ -282,4 +286,106 @@ def post(self):
...
+                        transfer_amount=fiat_ramp.payment_amount,
+                        token=fiat_ramp.token,
+                        receive_account=individual_recipient_user,
```
enjeyw: `receive_user` is probably less ambiguous - aware that this one is a hangover from my bad naming lol!

app/server/utils/credit_transfers.py
```diff
@@ -361,24 +364,63 @@ def make_blockchain_transfer(transfer_amount,
...
                           transfer_use=None,
                           transfer_mode=None,
                           require_sender_approved=True,
                           require_recipient_approved=True,
                           require_sufficient_balance=True,
                           automatically_resolve_complete=True,
-                          uuid=None):
+                          uuid=None,
+                          is_disbursement=False,
+                          is_reclaim=False):
```
enjeyw: given that disbursement and reclaim can't be both true, wondering if maybe this "type" should be a string config? not certain though.

enjeyw:
few changes to make, but overall this is looking good. clleeeeaaan codddeeeee

app/client/components/transferAccount/transferAccountManager.jsx
tristanhcole: Haha yeah

app/server/utils/poli_payments.py
tristanhcole: agh good point, need to check otherwise will raise an `AttributeError`

app/server/api/poli_payments_api.py
tristanhcole: Yes, it's an authing process.

They send us a unique token to our webhook, we use that to request a Poli link, which we can request the status from.

app/server/me_api/misc.py
tristanhcole: ok

app/server/utils/credit_transfers.py
tristanhcole: mhm need to think about that

app/server/utils/poli_payments.py
tristanhcole: Yes returns a dict for non 200's

PR #16, opened 2019-10-04
split up credit transfer and transfer account, move out organisation

wangfowen:
last model.py refactor i promise (for a while). same as before, earlier the merge the less likely for conflicts

changes:
- credit transfer and transfer account are in separate files. kept blockchain address in transfer account and blockchain transaction in credit transfer since those seemed related. could move them into their own file, blockchain instead? what makes the most sense?
- organisation extracted from models.py
- various transfer related enums moved into its own utils file since it's used in a lot of different places
- transfer card moved into its own file. the random method that was hanging out in user.py i've moved into a static method on transfer card
- changed a couple of foreign keys from class references to strings to reduce dependency issues


app/server/models/credit_transfer.py
```diff
@@ -0,0 +1,269 @@
...
+                    required_task_dict[task_type]['status'] = proposed_new_status
+                    required_task_dict[task_type]['hash'] = transaction.hash
+            except ValueError:
+                pass
+
+        return required_task_dict
+
+    @hybrid_property
+    def pending_blockchain_tasks(self):
+        return self._find_blockchain_tasks_with_status_of('PENDING')
```
wangfowen: these should also be an enum

app/server/models/organisation.py
```diff
@@ -0,0 +1,124 @@
...
+from server import db
+from server.exceptions import (
+    OrganisationNotProvidedException
+)
+from server.utils.blockchain_tasks import (
+    create_blockchain_wallet
+)
+from server.models.utils import ModelBase, OneOrgBase, ManyOrgBase, organisation_association_table
+
+@event.listens_for(Query, "before_compile", retval=True)
```
wangfowen: is it okay for this to be in `organisation.py`? wasn't sure how this code works, didn't see it called directly anywhere, but seems relevant to organisation?

app/server/models/transfer_card.py
```diff
@@ -0,0 +1,58 @@
...
+        return self._phone
+
+    @amount_loaded.setter
+    def amount_loaded(self, amount):
+        self._amount_loaded = amount
+        message = '{}{}{}'.format(self.nfc_serial_number, amount, self.transfer_account.token.symbol)
+        self.amount_loaded_signature = current_app.config['ECDSA_SIGNING_KEY'].sign(message.encode()).hex()
+
+    @staticmethod
+    def get_transfer_card(public_serial_number):
```
wangfowen: moved method from user.py here

enjeyw:
we should be able to get tests to pass this now!

app/server/models/credit_transfer.py
enjeyw: so I've shifted away from using an ENUM class directly in the db, because they blow up in a bad way during migrations. My compromise is to use a string hybrid property with checks on the setter (see eth_worker >sql_persistence > models.py, line 178)

`@hybrid_property
    def status(self):
        return self._status

    @status.setter
    def status(self, value):

        if value not in STATUS_STRING_TO_INT:
            raise ValueError('Status {} not allowed. (Must be {}'.format(value, STATUS_STRING_TO_INT))

        self._status = value
`

does that seem reasonable to u? it's a relatively new thing i've been trying out, only in the last month.

app/server/models/organisation.py
enjeyw: i think this should probably be in utils or models.py. while it checks the org model, it's bound to ALL db queries via an event listener (https://docs.sqlalchemy.org/en/13/core/event.html)

app/server/models/credit_transfer.py
wangfowen: what causes it to blow up? aren't they just strings?

app/server/models/organisation.py
wangfowen: when does the code get run? on compilation do all event listeners anywhere in the code get run? like if i move it to utils, do i have to do something special to ensure it still happens as before? i'd prefer it not to be in models.py, eventually it'd be good if everything was moved out and that file didn't exist anymore

wangfowen:
i rebased in nick/multi-tenant and based this commit off of that branch so it can be merged after that one

app/server/models/credit_transfer.py
enjeyw: >  aren't they just strings?
I'm pretty sure ENUMS are their own datatype in postgres. To be honest it's been ages since i looked at this (we haven't had to upgrade ENUMS in a while), and my understanding of db migrations wasn't great back then, but from memory this was the issue we were running into:

https://stackoverflow.com/questions/14845203/altering-an-enum-field-using-alembic

Might be worth re-exploring?

app/server/models/organisation.py
enjeyw: >  eventually it'd be good if everything was moved out and that file didn't exist anymore

Oh yep good point. Utils would be an excellent place to put it. 

>  on compilation do all event listeners anywhere in the code get run? like if i move it to utils, do i have to do something special to ensure it still happens as before?

I'm 90% sure you can just shift it. I'll give it a shot once I've merged this PR

PR #17, opened 2019-10-04
ussd tables and api entrypoint

wangfowen:
welp, turns out my refactor breaks alembic auto-generation of migrations. here i created the migration by moving my table into models.py, then moving it back out. will need to fix that.


app/server/models/ussd.py
```diff
@@ -0,0 +1,32 @@
...
+        but this system is really bad since we always need 
+        all languages defined for all new menu items, no graceful fallback 
+    """
+    display_text_en = db.Column(db.String, nullable=False)
+    display_text_sw = db.Column(db.String, nullable=False)
+
+class UssdSession(ModelBase):
+    __tablename__ = 'ussd_sessions'
+
+    session_id = db.Column(db.String, nullable=False, index=True, unique=True)
```
wangfowen: i only saw one other index in all of the codebase. is this correct? that's a bit worrying for performance

enjeyw:
> welp, turns out my refactor breaks alembic auto-generation of migrations. here i created the migration by moving my table into models.py, then moving it back out. will need to fix that.

lol!
https://stackoverflow.com/questions/45919768/how-to-get-alembic-to-recognise-models-from-multiple-model-files-in-flask
perhaps?

app/server/models/ussd.py
enjeyw: good catch. that's probably just me being shit. i'm never really sure when to index and when not to, which leads to a lack of indexing in general.

wangfowen: a super quick primer on indices:
- when you do a db query on a field (`User.query.filter_by(phone=msisdn)` for example), it scans all your rows looking at that entry, meaning O(n) search. when you have millions of rows, that's real bad
- when you add an index, it essentially creates another whole copy of your table, but sorted by whatever you indexed by. this means 2x (roughly, in hand wavy terms) storage costs and slightly slower write time, like plus O(ln n)? more hand wavy tree insertion algorithms, but querying by that field becomes essentially O(1)
- when you do a sort by, find by, basically any sort of sql operation, sql magically knows how to use the appropriate indices for that action (asterisk here, it sometimes messes up if you have too many similar indices on the same table. then you gotta go hunt down what's causing it to use the wrong one and that's sad times)
- if you're doing a search on multiple fields at once, you'll want a multi-field index to account for that (`db.Index(<name>, <field1>, <field2>, etc, <params>)`)

summary is: if you're often searching, sorting, etc, by a specific field or combination of fields, you should index on that to increase performance. storage cost is cheap these days so over-indexing is generally better than under-indexing

enjeyw: oh man that's _super_ helpful. So we should be indexing any field that get's a relatively commonly used filter_by query?

wangfowen: yep! filter_by, order_by, etc. i'm not sure if id needs to be indexed, it was done automagically in rails so idk if that was rails or postgres that did it

wangfowen: it's not as big a deal for tables that stay small, but for tables that'll get big real fast like CreditTransfer, it's important to have indices

app/migrations/env.py
```diff
@@ -20,7 +20,32 @@
 from flask import current_app
 config.set_main_option('sqlalchemy.url',
                        current_app.config.get('SQLALCHEMY_DATABASE_URI'))
-target_metadata = current_app.extensions['migrate'].db.metadata
+#target_metadata = current_app.extensions['migrate'].db.metadata
+from server.models import (credit_transfer, device_info, feedback, ip_address, kyc_application, models, organisation,
```
wangfowen: tried various fixes found online and this is the only one that works... not great since it's so manual and repetitive, open to suggestions to make it better. 

i don't actually really understand how metadata works... the original code recognized a subset of the classes, but not only just the ones in models.py, it recognized some of the other ones in other files but not all of them, very strange. 

this is the closest to an automatic solution: https://gist.github.com/logston/e0ab3e46dd51b27f83b5d8ad359289ad
but it errors out in ways that don't make sense to me

app/migrations/seed.py
```diff
@@ -0,0 +1,604 @@
+"""
```
enjeyw: So if we're not bringing across new GE tokens, we won't need this right?

app/server/api/ussd_api.py
```diff
@@ -23,27 +29,23 @@ def post(self):
         session_id = post_data.get('sessionId')
         phone_number = post_data.get('phoneNumber')
         user_input = post_data.get('text')
+        service_code = post_data.get('serviceCode')
 
         if phone_number:
-            """
-                TODO:
-                user = User.find_by(phone: UssdApi::Utilities.internationalize_phone(params[:phoneNumber], 'KE'))
```
enjeyw: Took a couple of seconds to realise this was Ruby 😅

app/migrations/seed.py
wangfowen: just not the community token part of it, we still need the USSD menu part

app/server/models/ussd.py
```diff
@@ -0,0 +1,46 @@
...
+    """
+    display_text_en = db.Column(db.String, nullable=False)
+    display_text_sw = db.Column(db.String, nullable=False)
+
+    @staticmethod
+    def find_by_name(name: str) -> "UssdMenu":
+        menus = UssdMenu.query.filter_by(name=name)
+        if menus.count() == 0:
+            sentry.captureMessage("No USSD Menu with name {}".format(name))
+            # should handle case if no invalid_request menu?
```
enjeyw: yep just ran into this - fails bad atm. Could probably just have something generic.

wangfowen: this should just not happen once we do the db seed for menus. all devs should run the seed as part of their setup so they at least have some data to work off of

enjeyw:
[Create data tables for USSD](https://trello.com/c/HDOVfJao/52-create-data-tables-for-ussd)

enjeyw:
[Create data tables for USSD](https://trello.com/c/HDOVfJao/52-create-data-tables-for-ussd)

enjeyw:
[USSD API endpoint](https://trello.com/c/01Tqjxob/17-ussd-api-endpoint)

enjeyw:
[USSD API endpoint](https://trello.com/c/01Tqjxob/17-ussd-api-endpoint)

enjeyw:
[Implement Ussd Processor](https://trello.com/c/tXAJUd6C/59-implement-ussd-processor)

PR #21, opened 2019-10-10
Fix for Issue #11

tristanhcole:
Fix for Issue #11. Merging `UploadedImage` with `UploadedDocument` for simplicity.


app/server/schemas.py
```diff
@@ -84,11 +84,20 @@ def get_profile_url(self, obj):
         return processed_profile_pictures
 
 
-class UploadedImageSchema(Schema):
-    id                      = fields.Int(dump_only=True)
-    filename                = fields.Str()
-    image_url               = fields.Function(lambda obj: obj.image_url)
-    credit_transfer_id      = fields.Int()
+class UploadedResourceSchema(Schema):
```
wangfowen: oh whoa i didn't see this file -- are we supposed to manually create an entry into it? my ussd tables didn't automatically generate this. and where does this get used?

wangfowen:
are there already existing uploaded documents/images? how are we going to handle migrating them over to the new table?

wangfowen:
also, it's a bit scary that we have so many different branches off of other un-merged branches at this point. will we have issues with database head merge for example? my understanding was that if the db was at a certain revision and that revision diverged, it'd be able to pick them up. but if we diverge off of an older revision, won't the db not see it?

tristanhcole:
No. Multi-tenant is a hard break on the old version. Not migrating any data over.

We can migrate heads reasonably easy, there isn’t usually to many problems.

tristanhcole:
We should probably merge soon

app/server/schemas.py
enjeyw: this gets used when we want to serialise a SQLAlchemy object into a JSON string. For example in user_api:

`            response_object = {
                'status': 'success',
                'message': 'Successfully Loaded.',
                'data': {
                    'user': user_schema.dump(user).data
                }
            }

            return make_response(jsonify(response_object)), 201
`

It's pretty verbose so it doesn't always get enforced. But yep, need to create them urself, which isn't too bad because it means u can create some pretty specific behaviour. 

As @RubenGeo pointed out, we should probably be doing consistent validation on our API, which this library also supports (https://marshmallow.readthedocs.io/en/stable/)

enjeyw:
> but for example, what if we both start working from A. you merge in several db changes B -> C. so now prod db is at C. at this point if i make changes to A -> D -> E, my revisions are on one prod db long passed. does it still know to pick it up?

In general, yeah it does (I think because it's basically just diff-ing the heads), though I have had a couple of times where it's failed

tristanhcole:
[Issue #11: Simplify external resource storage](https://trello.com/c/tZG99DIB/57-issue-11-simplify-external-resource-storage)

PR #22, opened 2019-10-10
usdd processor code

wangfowen:
ussd processor code translated from ruby to python, still need to write tests for it
copied over state machine code not translated yet


app/migrations/seed.py
```diff
@@ -263,51 +263,6 @@
   u.parent_id = start_menu.id
 end
 
-puts 'Creating convert token menu...'
```
wangfowen: convert is deprecated, we don't need to migrate these menus

app/server/models/ussd.py
```diff
@@ -19,6 +19,19 @@ class UssdMenu(ModelBase):
     display_text_en = db.Column(db.String, nullable=False)
     display_text_sw = db.Column(db.String, nullable=False)
 
+    @staticmethod
+    def find_by_name(name: str) -> "UssdMenu":
+        menus = UssdMenu.query.filter_by(name=name)
+        if menus.count() == 0:
+            sentry.captureMessage("No USSD Menu with name {}".format(name))
```
wangfowen: if we want to gracefully fail and log the issue, should we be using sentry?

enjeyw: Yep, exactly

PR #23, opened 2019-10-11
Added kobo_api & fields in user table & sepparate table for custom user attributes

RubenGeo:
- Added kobo_api as a sepperate end point in which the custom attributes are fist extracted and the standard `UserUtils.proccess_create_or_modify_user_request` function is used to add the user
- Created a custom_attribute_user_storage a related table in which custom attributes are stored
- Added `prefered language` , `bio` and `gender` as a column in the user table
- Functionality to input values in these columns when creating or updating a new user


.gitignore
```diff
@@ -7,6 +7,7 @@
 .DS_STORE
 .vscode
 .idea
+.env
```
RubenGeo: I added .env here to set the python path in vscode. Else vscode will complain about faulty imports

app/migrations/versions/180aa40ff01e_.py
```diff
@@ -0,0 +1,24 @@
+"""empty message
```
RubenGeo: I never worked with these migrations files before. The readme specifcally mentioned to commit these however I ran `python manage.py db migrate`. Which may resulted in multiple migration files?? Is that okay?

app/server/api/user_api.py
```diff
@@ -141,55 +136,57 @@ def post(self, user_id):
...
-            user.first_name = first_name
+    if not user:
+        response_object = {
+            'message': 'User not found'
+        }
+        return make_response(jsonify(response_object)), 400
 
-        if last_name and not last_name == user.last_name:
-            user.last_name = last_name
+    if first_name and not first_name == user.first_name:
```
RubenGeo: I think my linter auto styled some stuff. I think it may be worth it to switch to pycharm for me

app/migrations/versions/180aa40ff01e_.py
wangfowen: you might have run migrate multiple times when nothing changed. you only want to commit the empty ones if it merges the head when it has branched off, which is not the case here. if you delete all the new migrations and run migrate again, it should just generate one clean migration

app/server/models/user.py
```diff
@@ -39,6 +43,10 @@ class User(ManyOrgBase, ModelBase):
 
     first_name = db.Column(db.String())
     last_name = db.Column(db.String())
+    preferred_language = db.Column(db.String(), default='en_au')
```
wangfowen: per what we discussed, we want to remove the default

app/server/api/user_api.py
wangfowen: we may want some sort of linter config file to enforce one set of style guide, so we don't end up with so many of these commits changing styles back and forth (mine also did this)

app/server/models/custom_attribute_user_storage.py
```diff
@@ -0,0 +1,13 @@
+from server import db
+from server.models.utils import ModelBase
+from server.models.user import User
+
+
+class CustomAttributeUserStorage(ModelBase):
```
wangfowen: you'll want to add this to `app/migrations/env.py` so it gets picked up in the migration generation. still hoping we can figure out a better way to do that, very easy to forget to add new files to there

wangfowen:
@RubenGeo fyi I rebased your commits into one and based it off the new branch that has all my changes merged in it

app/server/api/user_api.py
enjeyw: I think linter config is a good idea. @tristanhcole and my pycharm settings might have something really funky with tabs and spaces, because the resulting output isn't just a style change - it's pulling the entire `put` method out of the class! 😱

app/migrations/versions/180aa40ff01e_.py
enjeyw: This migration file is pretty deep in the history, so I'm just gonna leave it - it's not really causing any harm anyways lol

```
46903aca6fe1 -> 8a2e011771d6 (head), empty message
a487024cdea9 -> 46903aca6fe1, empty message
8aa8a0606156 -> a487024cdea9, empty message
180aa40ff01e -> 8aa8a0606156, empty message
1a8a9213e85d -> 180aa40ff01e, empty message
1a8a9213e85d -> af906681b8b2 (head), empty message
132ea53022f1 -> 1a8a9213e85d (branchpoint), empty message
5c5b65647bed -> 132ea53022f1, empty message
ae98d36420a1 -> 5c5b65647bed, empty message

```

enjeyw: 1a8a9213e85d and a487024cdea9 are identical and causing migrations to break, while 180aa40ff01e and 8aa8a0606156 are empty:

`46903aca6fe1 -> 8a2e011771d6 (head),
a487024cdea9 -> 46903aca6fe1, empty message
8aa8a0606156 -> a487024cdea9, BREAKS
180aa40ff01e -> 8aa8a0606156, NULL
1a8a9213e85d -> 180aa40ff01e, NULL
132ea53022f1 -> 1a8a9213e85d (branchpoint), empty message
`

so gonna manually re-route 1a8a9213e85d  to 46903aca6fe1...

wangfowen: hm, won't deleting all the migrations from this PR and then just generating one new one fix it?

app/server/models/user.py
wangfowen: sharing context here -- turns out we don't want a default because USSD state machine sees if preferred language is set. if it's not, then it sends user through initial language configuration. so instead of a default preferred language, what would be better is having a default fallback language perhaps for sempo's system (could be en_AU) that we use if user has no preferred language

app/migrations/versions/180aa40ff01e_.py
enjeyw: yeah, but then u 2 would have to mess with ur db version. probably would have been smarter actually

app/server/models/user.py
enjeyw: that makes sense to me

app/server/api/user_api.py
```diff
@@ -20,16 +20,19 @@ class UserAPI(MethodView):
     @requires_auth
     def get(self, user_id):
 
-        can_see_full_details = AccessControl.has_suffient_role(g.user.roles,{'ADMIN': 'admin'})
+        can_see_full_details = AccessControl.has_suffient_role(
```
enjeyw: I get the reasoning behind splitting this over multiple lines (PEP8 and all that), but my instinctive response is that unless we're at the 121 line limit, this actually harms readability rather than improves it. open to other opinions though @tristanhcole @RubenGeo @wangfowen

app/server/utils/user.py
```diff
@@ -243,29 +262,45 @@ def save_device_info(device_info, user):
...
 
     attachments = attribute_dict.get('_attachments', [])
 
     for attachment in attachments:
         submitted_filename = attachment['filename'].split('/')[-1]
-        for attribute in custom_attributes.keys():
-            if submitted_filename == custom_attributes[attribute]['value']:
+        for attribute in custom_attributes:
+            if submitted_filename == attribute.value:
+                print('hier')
```
enjeyw: delete

enjeyw: (the print statement that is)

app/server/api/user_api.py
RubenGeo: I do not have strong opinion about this. I think this was messed up by my linter somewhere

app/server/models/custom_attribute_user_storage.py
RubenGeo: Thanks!

app/migrations/versions/180aa40ff01e_.py
RubenGeo: So I did this: 
- fetch your commits
- `git reset --hard upstream/ruben/uss`
- changed the user model : bio & gender --> enrolled_by & referred_by

How do I now migrate to the new version?

enjeyw:
@RubenGeo for some reason it's not letting me respond to ur question about migration! You do:

```
cd app

python manage.py db upgrade

```

and it should bring ur database up to date with what's currently on the branch. slack me if ur having troubles :)

app/migrations/versions/180aa40ff01e_.py
wangfowen: k i bombed all the migrations and created just one new one. @RubenGeo you'll probably have to drop your database and recreate it now

wangfowen:
should be good to merge if it looks good to you guys @RubenGeo @enjeyw

PR #24, opened 2019-10-14
Merging in all PRs based off nick/multi-tenant into main multi-tenant branch

enjeyw:


config.py
```diff
@@ -90,7 +90,9 @@
 
 
 IS_TEST = specific_parser['APP'].getboolean('IS_TEST', False)
-IS_PRODUCTION = specific_parser['APP'].getboolean('IS_PRODUCTION', True)
+IS_PRODUCTION = specific_parser['APP'].getboolean('IS_PRODUCTION')
```
enjeyw: This should prevent Sentry from automatically logging errors on local machines

enjeyw: though we definitely need to get a better way to share configs

enjeyw:
most of this is refactoring

PR #27, opened 2019-10-17
ussd menu migrations

wangfowen:
- seed for menus
- add in more test cases since preferred_language fixed

### DIFF
--- a/app/migrations/seed.py
+++ b/app/migrations/seed.py
@@ -8,18 +8,22 @@ from server import db, create_app
 from server.models.ussd import UssdMenu
 
 
-def get_or_create_menu(name, **kwargs):
+def get_or_create_menu(name, description, display_text_en, display_text_sw, parent_id=None):
     instance = UssdMenu.query.filter_by(name=name).first()
     if instance:
         return instance
     else:
-        instance = UssdMenu(name=name)
+        instance = UssdMenu(
+            name=name,
+            description=description,
+            display_text_en=display_text_en,
+            display_text_sw=display_text_sw,
+            parent_id=parent_id
+        )
+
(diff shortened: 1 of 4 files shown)


app/migrations/seed.py
```diff
@@ -1,182 +1,125 @@
...
-  c.contract_address = '0x8a890d89cd3585b06a48a174dd847d7818d3074f'
-end
 
-CommunityToken.where(name: 'Kangemi').first_or_create! do |c|
-  c.name = 'Kangemi'
-  c.token_alias = 'KANG'
-  c.token_identifier = '5b9cccbc81df57e455340ee8'
-  c.contract_address = '0x22e9839aeff9818a8b3617586167e8b92f2ed3c4'
-end
+def get_or_create_menu(name, **kwargs):
```
enjeyw: wondering why ur setting kwargs rather than explicitly specifying the arguments allowed? 
Given it's a factory for a single class, we should be able to list them all, which would help prevent incorrect args creeping in no?

wangfowen: good idea. fixed.
for another PR -- can we also add a seed for creating a user and a token? i dropped my database to test this and lost my user and token that we created that other time through the test python script. would be good if we populated our local dbs with enough stuff to be able to run from the get go

wangfowen: i guess that's a bit different actually. there's a seed that we need to run everywhere -- which is the case for this script since we need the menus on prod as well. and then there's one we need to run locally to have a good dev setup like user and token

enjeyw: I'll just create different triggers for them, but it make sense to have them both as seed files in migrations probs...

PR #28, opened 2019-10-17
every model moved out of models/models.py, env.py fixed

wangfowen:
from some preliminary testing, this looks like it _should_ no longer require adding each new model file into the env.py file. my understanding is they do in fact all have the same metadata? 

also ended up moving everything else out into their own models files so that we don't have a partial models.py anymore. we can combine things into single files if they make sense to group together, but i personally like having them in separate files so i can easily search by file name for the model i'm looking for


enjeyw:
Agreed, pulling this in right away.

PR #29, opened 2019-10-18
Store and check reset password token to prevent their reusebility

RubenGeo:
Added an array of password_reset_tokens to the user model. And functions to
- Save a password reset token
- Check whether a token is in the array
- Clear all tokens (after succesfull pass reset)
- Clear up any expired tokens in the password_reset_tokens field


app/server/models/user.py
```diff
@@ -408,6 +410,30 @@ def decode_single_use_JWS(cls, token, required_type):
 
             return {'success': False, 'message': e}
 
+    def save_password_reset_token(self, password_reset_token):
+        # make a "clone" of the existing token list
+        self.clear_expired_tokens()
+        current_reset_tokens = self.password_reset_tokens[:]
+        current_reset_tokens.append(password_reset_token)
+        # set db value
+        self.password_reset_tokens = current_reset_tokens
```
wangfowen: we don't have to do `db.session.commit()` for these kinds of things?

wangfowen: oh i see one at the end of the reset password api call, but not one at the end of request password reset email one. have you tested this works as expected?

RubenGeo: That's what i thought too. But it get stored in the db without it

app/server/models/user.py
```diff
@@ -487,9 +487,17 @@ def set_pin(self, supplied_pin=None, is_activated=False):
 
         self.hash_password(pin)
 
-    # TODO(ussd): change to a field once we figure out what's the deal with reset_token
     def is_resetting(self):
```
RubenGeo: @wangfowen  This is what you meant?

app/server/models/user.py
```diff
@@ -461,9 +487,17 @@ def set_pin(self, supplied_pin=None, is_activated=False):
...
-        return False
+        self.clear_expired_tokens()
+        valid_tokens = []
+        for token in self.password_reset_tokens:
```
wangfowen: is it still necessary to do this loop if we've cleared the expired tokens? all that remains should be valid ones right?

wangfowen: could just do `self.password_reset_tokens.count() > 0` after clearing expired i think

RubenGeo: Changed this, thanks :+1:

app/server/models/user.py
tristanhcole: That's because of this function that is executed after a successful request

https://github.com/teamsempo/SempoBlockchain/blob/381f82424b1eec8fa1d6446f84a1099e5b039575/app/server/__init__.py#L158-L163

wangfowen:
@RubenGeo yeahh in writing the tests, would be good to cover that None case for all the methods. I'm worried that in the future we'll change the implementation and accidentally leave it out. especially since it implicitly happens in the core methods via the clear, could be very easy to accidentally change the behavior

PR #30, opened 2019-10-21
Exchange

enjeyw:
Not ready to merge in yet, just wanted to get the PR out there


app/manage.py
```diff
@@ -1,42 +1,51 @@
...
 parent_dir = os.path.abspath(os.path.join(os.getcwd(), ".."))
 sys.path.append(parent_dir)
 sys.path.append(os.getcwd())
 
-from server import create_app, db
-from server.models.blockchain_address import BlockchainAddress
 
 class UpdateData(Command):
 
     def run(self):
```
wangfowen: is this the seed? should we be adding this to the instructions for what to run in dev setup?

app/server/api/token_api.py
```diff
@@ -61,8 +62,62 @@ def post(self):
...
+
+        if exchange_contract_address and exchange_contract_id:
+            return make_response(jsonify(
+                {'message': 'Must ONLY specify one of exchange contract id or address'}))
+
+        if exchange_contract_id:
+            exchange_contract = ExchangeContract.query.get(exchange_contract_id)
+
+        elif exchange_contract_address:
+            exchange_contract = ExchangeContract.query.filter_by(blockchain_address=exchange_contract_address).first()
```
wangfowen: should probably put an index on this. should be happening fairly often right?

also, i found out anything that's a primary key automatically gets an index.

app/server/api/token_api.py
```diff
@@ -61,8 +62,62 @@ def post(self):
...
+
+        else:
+            return make_response(jsonify(
+                {'message': 'Must specify one of exchange contract id or address'}))
+
+        if exchange_contract is None:
+            return make_response(jsonify(
+                {'message': 'No Exchange Contract found for ID: {}'.format(exchange_contract_id)}))
+
+        token.add_exchange_contract(exchange_contract)
```
wangfowen: hmm so a token has an exchange contract and exchange contract has a `reserve_token_id`. is that the token you exchange to, and the contract defines the rate?

app/server/api/token_api.py
```diff
@@ -61,8 +62,62 @@ def post(self):
...
+        elif exchange_contract_address:
+            exchange_contract = ExchangeContract.query.filter_by(blockchain_address=exchange_contract_address).first()
+
+        else:
+            return make_response(jsonify(
+                {'message': 'Must specify one of exchange contract id or address'}))
+
+        if exchange_contract is None:
+            return make_response(jsonify(
+                {'message': 'No Exchange Contract found for ID: {}'.format(exchange_contract_id)}))
```
wangfowen: none found for id nor address?

app/server/models/credit_transfer.py
```diff
@@ -163,8 +174,8 @@ def resolve_as_completed(self, existing_blockchain_txn=None):
         self.resolved_date = datetime.datetime.utcnow()
         self.transfer_status = TransferStatusEnum.COMPLETE
 
-        self.sender_transfer_account.balance -= self.transfer_amount
-        self.recipient_transfer_account.balance += self.transfer_amount
+        self.sender_transfer_account.add_to_balance(-1 * self.transfer_amount)
```
wangfowen: seems like there should be tests for this kind of stuff :|

app/server/models/token.py
```diff
@@ -1,9 +1,15 @@
...
 )
-from server.models.utils import ModelBase
 
+from server.models.utils import (
+    ModelBase,
+    exchange_contract_token_association_table
+)
+
+from server.models.exchange import ExchangeContract
+import server.models.transfer_account
```
wangfowen: circular dependency or why do it this way?

app/server/models/token.py
```diff
@@ -43,5 +63,27 @@ def token_amount_to_system(self, token_amount):
         return int(token_amount) * 100 / 10**self.decimals
 
     def system_amount_to_token(self, system_amount):
-        return int(float(system_amount) / 100 * 10**self.decimals)
+        return int(float(system_amount) * 10**self.decimals / 100)
+
+    def add_exchange_contract(self, exchange_contract: ExchangeContract):
+
+        exchange_transfer_account = (server.models.transfer_account.TransferAccount.query
```
wangfowen: are exchange transfer accounts like a special thing? they're not associated with any particular user right?

app/server/models/token.py
```diff
@@ -43,5 +63,27 @@ def token_amount_to_system(self, token_amount):
...
+                                     .filter_by(token=self)
+                                     .filter_by(exchange_contract=exchange_contract)
+                                     .first())
+
+        if not exchange_transfer_account:
+            exchange_transfer_account = server.models.transfer_account.TransferAccount(
```
wangfowen: huh? why would this exist if the previous one didn't?

app/server/models/transfer_account.py
```diff
@@ -86,17 +95,17 @@ def get_approval(self, receiving_address):
     @hybrid_property
     def total_sent(self):
         return int(
-            db.session.query(func.sum(CreditTransfer.transfer_amount).label('total')).execution_options(show_all=True)
-            .filter(CreditTransfer.transfer_status == TransferStatusEnum.COMPLETE)
-            .filter(CreditTransfer.sender_transfer_account_id == self.id).first().total or 0
+            db.session.query(func.sum(server.models.credit_transfer.CreditTransfer.transfer_amount).label('total')).execution_options(show_all=True)
```
wangfowen: are we organizing our code incorrectly? this circular dependency business feels like we're doing something wrong...

wangfowen:
@enjeyw i rebased your branch so all your commits are at the top. you might have to reset to origin when you pull. if you have other changes you made since this PR, make sure to store them somewhere so they don't get lost

app/manage.py
enjeyw: yeah my plan was:
- move most of the actual logic of this into seed.py, and just call the function from here
- do the same with the other seed data.

Add the call to the readme.

app/server/api/token_api.py
enjeyw: oh yep will do.

app/server/models/transfer_account.py
enjeyw: Hmmm yeah doesn't feel right  does it?
I tried the accepted answer here:
https://stackoverflow.com/questions/29217076/flask-sqlalchemy-query-have-to-import-all-models/29218728

But couldn't even get the imports in `__init.py__` to recognise

app/server/models/token.py
enjeyw: I had a case when I was dev-ing this, though tbh it's probably not something that would occur in general without something really weird happening. Should I just remove?

app/server/models/token.py
enjeyw: Yep. Transfer accounts can be associated with anything that can send/receive tokens. Ie:
- users
- organisations (which are in turn bound to admin users)
- exchange contracts

app/server/models/token.py
enjeyw: yeah circular dep. `import server.models.transfer_account` style imports didn't even consistently work, so there was a bunch of trial and error :|

app/server/models/credit_transfer.py
enjeyw: Yep will add

app/server/api/token_api.py
enjeyw: Ohhh good catch. The responses were originally seperate for each case. I'll split up again so we can have properly informative messaging

app/server/api/token_api.py
enjeyw: Close. Even if you're exchanging between to CIC tokens (aka ones defined by a bonding curve WRT a reserve), you still go through the reserve token to make the full exchange. so you could have an exchange that is `CIC1 > RESERVE` but you could also have one that is `CIC1 > RESERVE > CIC2`

PR #32, opened 2019-10-23
Philip/ussd

PhilipWafula:
Not ready for merge.
Some functions necessary for state machine are not yet available.

### DIFF
--- a/app/server/utils/ussd/kenya_ussd_state_machine.py
+++ b/app/server/utils/ussd/kenya_ussd_state_machine.py
@@ -1,189 +1,548 @@
 """
-    TODO(ussd): state machine code from GE - need to translate to python
-
-    state :initial_language_selection, initial: true
-    state :initial_pin_entry, initial: true
-    state :initial_pin_confirmation
-
-    state :start, initial: true
-    state :send_enter_recipient
-    state :send_token_amount
-    state :send_token_reason
-    state :send_token_reason_other
-    state :send_token_pin_authorization
-    state :send_token_confirmation
-    state :account_management
-    state :balance_inquiry_pin_authorization
-    state :choose_language
-    state :pin_change
-    state :current_pin
-    state :new_pin
-    state :new_pin_confirmation
-    state :directory_listing
-    state :exchange_token
-
-    state :exchange_rate_pin_authorization
-    state :exchange_token_agent_number_entry
-    state :exchange_token_amount_entry
-    state :exchange_token_pin_authorization
-    state :exchange_token_confirmation
-    state :help
-    state :my_business
-    state :about_my_business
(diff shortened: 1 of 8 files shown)


test/unit/app/utils/test_kenya_state_machine.py
```diff
@@ -0,0 +1,137 @@
...
+                             (initial_pin_confirmation_state, standard_user, "1212", "exit_pin_mismatch"),
+                             # start state transition tests
+                             (start_state, standard_user, "1", "send_enter_recipient"),
+                             (start_state, standard_user, "2", "account_management"),
+                             (start_state, standard_user, "3", "directory_listing"),
+                             (start_state, standard_user, "4", "exchange_token"),
+                             (start_state, standard_user, "5", "help"),
+                             (start_state, standard_user, "6", "exit_invalid_menu_option"),
+                             # send_enter_recipient state tests
+                             # (send_enter_recipient_state, standard_user, "0712345678", "send_token_amount"),
```
wangfowen: i'd put TODOs around these so it's more clear which ones are comments vs things that need to be fixed later

app/server/utils/ussd/kenya_ussd_state_machine.py
```diff
@@ -1,189 +1,548 @@
...
+    def is_authorized_pin(self, user_input):
+        pass
+
+    def is_blocked_pin(self, user_input):
+        pass
+
+    def is_valid_new_pin(self, user_input):
+        pass
+
+    def save_business_directory_info(self, user_input):
```
wangfowen: should put TODOs on top of these methods so we know we need to fix these

app/server/utils/ussd/kenya_ussd_state_machine.py
```diff
@@ -1,189 +1,548 @@
...
+
+from server import db
+from server.models.user import User
+from server.models.ussd import UssdSession
+from server.utils.phone import proccess_phone_number, send_generic_message
+
+
+def get_user(query_filter):
+    user = User.query.filter_by(phone=proccess_phone_number(phone_number=query_filter, region="KE")).first()
+    if user is not None:
```
wangfowen: what happens if user is None then?

test/unit/app/utils/test_kenya_state_machine.py
PhilipWafula: I'll add the TODOs on top of these and push

app/server/utils/ussd/kenya_ussd_state_machine.py
PhilipWafula: Should have had an exception there, I'll add one.

PR #33, opened 2019-10-23
refactoring create user form

wangfowen:
- using redux form (muuuch cleaner forms - validation, state handling, submission)
- using css modules (showcase what this could be like)
- split up create user and vendor (they're already very different and about to become more so)
- fix broken logic and accidental global vars


app/client/components/user/createVendor.jsx
```diff
@@ -0,0 +1,329 @@
+import React  from 'react';
+import { connect } from 'react-redux';
+
+import AsyncButton from '../AsyncButton.jsx'
+import QrReadingModal from "../qrReadingModal.jsx"
+import { createUser, RESET_CREATE_USER } from '../../reducers/userReducer'
+import { StyledButton, ErrorMessage, ModuleHeader } from '../styledElements'
+import { ManagerInput, InputObject, InputLabel, Code } from './userElements'
+
+class CreateVendorForm extends React.Component {
```
wangfowen: left the way create vendor works alone for comparison, but did fix the logic and accidental globals here also

app/client/components/user/createVendor.jsx
```diff
@@ -0,0 +1,357 @@
+import React  from 'react';
+import { connect } from 'react-redux';
+import styled from "styled-components";
+
+import AsyncButton from '../AsyncButton.jsx'
+import QrReadingModal from "../qrReadingModal.jsx"
+import { createUser, RESET_CREATE_USER } from '../../reducers/userReducer'
+import { StyledButton, ErrorMessage, ModuleHeader } from '../styledElements'
+
+class CreateVendorForm extends React.Component {
```
wangfowen: left the way create vendor works alone for comparison, but did fix the logic around one_time_code and accidental globals here also

app/package.json
```diff
@@ -63,8 +64,8 @@
     "moment-locales-webpack-plugin": "^1.0.7",
     "react-test-renderer": "^16.4.1",
     "redux-mock-store": "^1.5.3",
+    "style-loader": "^1.0.0",
     "ts-jest": "^23.0.0",
-    "typescript": "^2.9.2",
```
wangfowen: oh huh, i added typescript, then changed my mind and removed it. didn't realize it was already in here. is there an intention to move to typescript?? i'd be more than happy to incrementally add it in

app/client/components/user/createUser.jsx
```diff
@@ -1,422 +1,200 @@
...
-    this.handleInputChange = this.handleInputChange.bind(this);
-    this.handleCashierChange = this.handleCashierChange.bind(this);
-  }
+    return (
+        <div>
+            <ModuleHeader>Create a {props.transferAccountType} account</ModuleHeader>
+
+            <div style={{padding: '1em'}}>
+                <form onSubmit={handleSubmit}>
+                    <Field name="firstName" component={renderField} type="text" label={'First Name'} />
```
wangfowen: before i just add it as `<Field component="input" className={styles.class} />` -- and just styled the field component itself. would something like that be possible in styled-components? styling custom components

app/package.json
enjeyw: Yeah I'd like to start typing things incrementally on both the backend and the frontend. Especially in cases where the required type is super ambiguous.

app/client/components/user/createUser.jsx
enjeyw: do u mean like this?
https://www.styled-components.com/docs/basics#styling-any-component

wangfowen:
checkk it outtt some typescript woooo

i've switched CreateVendor to be a separate form but sharing the same logic as CreateUser now. we can merge the forms if we do decide to use all the same fields from user in vendor, but for now i'm leaving it the same

still TODO: fix the checkbox and QR code populating field

wangfowen:
welp, typescript errors will block builds now it looks like. a good and bad thing. error comes from styled-components... might be a sign to change it up ;)

```

File: app/client/components/user/CreateUser.tsx [hunk 5]
```diff
@@ -0,0 +1,143 @@
...
+    IS_USING_BITCOIN: boolean
+    MAXIMUM_CUSTOM_INITIAL_DISBURSEMENT: number
+    CURRENCY_NAME: string
+  }
+}
+
+type Form = ICreateUser & ICreateVendor
+type Props = DispatchProps & StateProps & OuterProps
+
+class CreateUser extends React.Component<Props> {
```
wangfowen: fyi this is how you can tell what props there are, and whether they're from redux state/dispatch or passed into the component (outer)

wangfowen:
k compile errors fixed and TODOs i left before have been addressed, should have same behavior as before now. i tested that it hits `createUser` with the expected fields when you fill out the form, but i'm erroring with "Transfer card not found". is that because my environment isn't set up right?
one of the params is `require_transfer_card_exists: true,`: does that actually need to change now for GE customers we create through admin interface?

if it looks good then we can merge this

enjeyw:
> k compile errors fixed and TODOs i left before have been addressed, should have same behavior as before now. i tested that it hits `createUser` with the expected fields when you fill out the form, but i'm erroring with "Transfer card not found". is that because my environment isn't set up right?

Yeah effectively- when u enter a 6 digit code, it's trying to find the physical nfc card with the same number, which doesn't existed. The field accepts both 6 digit transfer cards, or 7+ digit phone numbers - if u enter a phone number it shouldn't error.

> one of the params is `require_transfer_card_exists: true,`: does that actually need to change now for GE customers we create through admin interface?

Oooof is that hard-coded?

app/migrations/versions/1a852ec3bc90_.py
```diff
@@ -0,0 +1,30 @@
...
+# revision identifiers, used by Alembic.
+revision = '1a852ec3bc90'
+down_revision = '8d672062eb10'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('user', sa.Column('default_transfer_account_id', sa.Integer(), nullable=True))
```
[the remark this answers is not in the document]
wangfowen: for new admin user creation flow i'm thinking admin sets a default token from a dropdown. then user creation will create transfer account with token_id being that token, then set default_transfer_account to be the new transfer account

wangfowen:
@enjeyw 

> Yeah effectively- when u enter a 6 digit code, it's trying to find the physical nfc card with the same number, which doesn't existed. The field accepts both 6 digit transfer cards, or 7+ digit phone numbers - if u enter a phone number it shouldn't error.
should we add some sort of validation then? if it's more than 7 digits, validate it's a proper phone number. should just be numerals right? no + at the beginning necessary or - between digits?

> Oooof is that hard-coded?
yeahh require_transfer_card_exists=true is hard-coded. how should it work instead?

PR #34, opened 2019-10-26
new admin user creation

wangfowen:


app/client/components/user/CreateUserForm.tsx
```diff
@@ -46,32 +51,39 @@ class CreateUserForm extends React.Component<InjectedFormProps<ICreateUser, Oute
...
-        </Field>
+        </InputField>
       )
     } else {
-      uniqueIdentifierInput = <Field name="blockchainAddress" component={TextInput} type="text" label='Bitcoin Address' />
+      uniqueIdentifierInput = <InputField name="blockchainAddress" label='Bitcoin Address' isRequired />
     }
 
-
+    //TODO(admin_create): set a default for it to be easier for staff? should it be 400 or 40000 if we're giving 400 CICs?
```
wangfowen: @enjeyw if we want to give 400 tokens, is it initial disbursement of 400 or 40000? and should we just set it default to be that instead of having them type it every time?

app/server/api/user_api.py
```diff
@@ -126,6 +126,7 @@ def post(self, user_id):
 
         post_data = request.get_json()
 
+        # TODO(admin_create): pass in current user
         response_object, response_code = UserUtils.proccess_create_or_modify_user_request(
```
wangfowen: are we able to pull info about the user making the API call? or do we just have their token to validate they have the right permissions?

wangfowen: we need it for `enrolled_by`

app/server/utils/user.py
```diff
@@ -504,6 +504,9 @@ def proccess_create_or_modify_user_request(attribute_dict,
 
         return response_object, 200
 
+    #TODO(admin_create): add enrolled by if passed in
+    #TODO(admin_create): add bio, gender
+    #TODO(admin_create): if passed in default token, use that token for default transfer account
```
wangfowen: is it possible to set the token on the transfer account created if it's being done by the eth worker?

app/client/components/user/CreateUserForm.tsx
```diff
@@ -46,32 +51,39 @@ class CreateUserForm extends React.Component<InjectedFormProps<ICreateUser, Oute
...
         <div style={{padding: '1em'}}>
           <form onSubmit={this.props.handleSubmit}>
-            <Field name="firstName" component={TextInput} type="text" label={'First Name'} />
-            <Field name="lastName" component={TextInput} type="text" label={'Last Name'} />
             {uniqueIdentifierInput}
+            <InputField name="firstName" label='Given Name(s)' isRequired />
+            <InputField name="lastName" label='Family/Surname' />
+            <InputField name="bio" label='Directory Entry' />
+            <InputField name="location" label='Location' />
+            <SelectField name="gender" label='Gender' options={["Male", "Female", "Other"]} />
```
wangfowen: @RubenGeo this is where we'd put the business categories in the admin user creation interface, could be done similarly to gender

app/client/components/user/CreateUserForm.tsx
enjeyw: so on the _user interface_ it's 400, but it's stored everywhere as 400,00 (aka the amounts in cents). There's actually already a default setting, thats triggered through "make_initial_disbursement" when the user is approved. It's currently set through the config file via "STARTING_BALANCE", though it should probably be migrated to a setting associated with an organisation.

app/server/api/user_api.py
enjeyw: oh yep we already do this! There's a column on the base model class called "authorising_user_id"  which is by default set to the user id of the user making the api call, as determined by their auth token:

```
    authorising_user_id = db.Column(db.Integer, default=get_authorising_user_id)
.
.
.
def get_authorising_user_id():
    if hasattr(g, 'user'):
        return g.user.id
    elif hasattr(g, 'authorising_user_id'):
        return g.authorising_user_id
    else:
        return None

```

so u can either just use that as `enrolled_by`, or set it using a similar process, or run `enrolled_by` as a hybrid property.

app/server/utils/user.py
enjeyw: what do u mean by "being done by the eth worker"? In general eth worker is pretty passive... it takes blockchain txn requests from the main app and tries to do them.

app/client/components/form/FormValidation.ts
```diff
@@ -0,0 +1,5 @@
+const FormValidation = {
+  required: (value: string | number) => (value !== undefined && value !== "") ? undefined : "This field is required"
```
enjeyw: 😍

app/client/components/user/CreateVendorForm.tsx
```diff
@@ -60,33 +44,33 @@ class CreateVendorForm extends React.Component<InjectedFormProps<ICreateUser, Pr
...
+        <InputField name="publicSerialNumber" label={isCashierAccountValue ? 'Cashier Phone Number' : 'Phone Number or ID'} isRequired>
           <QrReadingModal
             updateData={ (data: string) =>  this.setSerialNumber(data) }
           />
-        </Field>
+        </InputField>
       )
     } else {
-      uniqueIdentifierInput = <Field name="blockchainAddress" component={TextInput} type="text" label='Bitcoin Address' />
+      uniqueIdentifierInput = <InputField name="blockchainAddress" label='Bitcoin Address' isRequired />
```
enjeyw: we're actually deprecating support for bitcoin, so we could probably just kill these bits off.

app/server/utils/user.py
wangfowen: ohh okay, i haven't dug into that code, but this means transfer account is created on the app side and not on eth worker? is it created after eth worker finishes running then? it sets the blockchain address on the transfer account based on what's returned from the eth worker right?

app/client/components/user/CreateUserForm.tsx
wangfowen: eh so STARTING_BALANCE is added automatically? do we still need the initial disbursement amount input field then? and agreed, starting balance should be moved to the organisation level instead of being a global

wangfowen: huh... i've looked through the user creation code now. so there's both starting balance AND custom initial disbursement. are they used differently? should we just remove the custom initial disbursement from the admin user creation form?

app/server/api/user_api.py
wangfowen: huh, i see that authorising_user_id is on every model, but i don't see where it gets set. where is that happening?

app/server/utils/user.py
wangfowen: okay i see the code where transfer account is created as part of the user creation flow, but i don't see where it calls the eth worker. if we provide a public serial number instead of a blockchain address, it just creates a transfer account without an address...? where would it update that transfer account to the newly created wallet address?

wangfowen: oh wait i see it. on transfer account init, if no blockchain address is created it kicks off a celery worker that sync will return the blockchain address from wallet creation. is that correct?

app/client/components/user/CreateUserForm.tsx
enjeyw: custom initial disbursement is useful if the admin wants to set up a user with an amount that isn't the default, which is pretty often, though the UI should probably do a better job of suggesting it's not default

app/server/api/user_api.py
enjeyw: in the column defintion itself:
`    authorising_user_id = db.Column(db.Integer, default=get_authorising_user_id)`

The default part calls the function

app/server/utils/user.py
enjeyw: ^^ yep that's correct. Creating a blockchain wallet is (effectively) instant on the eth worker's side, which is why the transfer_account waits for the result

enjeyw: so to draw a conclusion under the whole thing, yes `transfer_account.token = token` is a valid operation

app/client/components/user/CreateUserForm.tsx
wangfowen: @enjeyw got it, going to change it to "additional_initial_disbursement" in that case

app/server/api/user_api.py
wangfowen: coool!

PR #35, opened 2019-10-29
Customer Support via Intercom

tristanhcole:
Adding customer support via intercom.

Groups `organisations` into companies on intercom.

Added identity verification using HMAC for extra security


tristanhcole:
[Customer Support](https://trello.com/c/7AF19elG/62-customer-support)

app/server/utils/intercom.py
```diff
@@ -2,8 +2,23 @@
...
+    :param user_id: Sempo UserID to generate a unique hash.
+    :param device_type: Web, Android or iOS
+    :return:
+    """
+    if device_type is 'WEB':
+        secret = current_app.config['INTERCOM_WEB_SECRET']
+    elif device_type is 'ANDROID':
+        secret = current_app.config['INTERCOM_ANDROID_SECRET']
+    else:
+        return Exception('No device_type provided')
```
enjeyw: `return Exception(f'device_type {device_type} provided not recognised. Allowed device types are {ALLOWED_INTERCOM_DEVICE_TYPES}')

enjeyw:
> should we just punt the decision for later

My instinct is to punt until we can properly review what to do about it all lol

PR #36, opened 2019-10-29
new sms utils

wangfowen:
- Nuke message processor
- Create util for determining how to send messages
- Move out phone utils for sending specific messages since they're only used in one place each and not general
- Add Africa's Talking client

Testing done:
- new unit test for if choosing channel is correct
- ran a local test of sending message to Africa's Talking sandbox, message was successful


app/server/utils/user.py
```diff
@@ -564,3 +564,29 @@ def proccess_create_or_modify_user_request(attribute_dict,
     }
 
     return response_object, 200
+
+
+def send_onboarding_message(to_phone, first_name, credits, one_time_code):
```
wangfowen: moved out from phone.py. the other ones didn't seem to be used at all so i just deleted them

app/server/utils/phone.py
```diff
@@ -41,68 +36,35 @@ def proccess_phone_number(phone_number, region=None, ignore_region=False):
...
-                .format(
-            current_app.config['PROGRAM_NAME'],
-            credits/100,
-            current_app.config['CURRENCY_NAME'],
-            pin,
-            current_app.config['CURRENCY_NAME']
-        )
 
-        # send_generic_message(to_phone, receiver_message)
+# just checking by area code may break down one day since multiple countries share the same country codes...
```
tristanhcole: I think it's pretty scalable for time being

test/unit/app/utils/test_phone.py
```diff
@@ -3,15 +3,36 @@
 """
 import pytest
 
+from server.utils.phone import ChannelType
 
-@pytest.mark.parametrize("phone,expected", [
-    ("0401391419", "+61401391419"),
-    ("+961401391419", "+961401391419"),
+
+@pytest.fixture(scope='function')
```
tristanhcole: Nice. It's a lot cleaner having the fixtures in the file they're actually used in (given many are only used in one file)

app/server/utils/chatbot_controller.py
```diff
@@ -1,461 +0,0 @@
-import datetime, re
```
enjeyw: this was the first code we ever wrote on the new system. RIP

app/server/utils/phone.py
```diff
@@ -41,68 +36,35 @@ def proccess_phone_number(phone_number, region=None, ignore_region=False):
...
+def channel_for_number(phone):
```
enjeyw: might be good to use `phonenumbers.parse(number)` here to extract the country code? they've probably got better edge case management. no strong opinion though.

wangfowen:
> So did you just delete the WhatsApp and chatbot controller as we don't use them?

yeahh we're sure they're not going to be used anymore right?

app/server/utils/phone.py
wangfowen: i played around with `phonenumber.parse` but it seemed like you needed to pass in the region as well? otherwise i got an empty region in the returned object. or did i do it wrong?

wangfowen: ah i was trying to print `phone_obj.region` and that was empty string, but country_code comes out correctly. will switch to that

wangfowen: hmm just tried it with a fake trinidadian phone number and parse outputs country code as just 1. in which case matching by string might actually give us better granularity

enjeyw:
> > So did you just delete the WhatsApp and chatbot controller as we don't use them?
> 
> yeahh we're sure they're not going to be used anymore right?

yeah, because if we were to re-implement a chatbot, I think we'd almost certainly use the new ussd state-machine

app/server/utils/phone.py
enjeyw: oh weird! all g then

PR #37, opened 2019-10-29
adding i18n and using it for ussd menus

wangfowen:


app/server/utils/ussd/kenya_ussd_processor.py
```diff
@@ -68,11 +68,12 @@ def replace_vars(menu: UssdMenu, ussd_session: UssdSession, display_text: str, u
             replacements.append(['%exchange_amount%', ussd_session.session_data['exchange_amount']])
         elif 'pin_authorization' in menu.name or 'current_pin' in menu.name:
             if user.pin_failed_attempts() > 0:
-                # TODO: not a great way to do i18n...
+                # TODO(ussd): replace this with i18n placeholders
```
enjeyw: this todo being to swap things into the YAML file?

enjeyw:
hmmm this is now failing for some other reason

app/server/utils/ussd/kenya_ussd_processor.py
wangfowen: swap into the yaml file and use the built in #{} placeholders for variables instead of the final replace_vars business

wangfowen:
oo mobile does i18n with json? i found yml to be more concise so i opted for it. we can switch to json

app/server/__init__.py
```diff
@@ -39,6 +40,9 @@
                     backend=config.REDIS_URL,
                     task_serializer='json')
 
+i18n.load_path.append(os.path.abspath(os.path.join(os.getcwd(), "app/server/locale")))
```
wangfowen: @enjeyw this was what was causing test failures. had to change to abs path instead of just the string for docker. so locally when i invoke tests, i'm doing it from root folder. will this cause issues for when we run the server, which happens from the /app folder?

PR #39, opened 2019-10-30
new admin user creation flow

wangfowen:
- generalize fields better
- make validation better
- hook in new admin fields
- organisation sms


app/server/models/organisation.py
```diff
@@ -47,6 +48,8 @@ class Organisation(ModelBase):
     email_whitelists    = db.relationship('EmailWhitelist', backref='organisation',
                                           lazy=True, foreign_keys='EmailWhitelist.organisation_id')
 
+    welcome_message     = db.Column(db.String)
```
wangfowen: changing this in i18n commit

app/server/models/organisation.py
```diff
@@ -55,3 +59,9 @@ def __init__(self, **kwargs):
             wei_target_balance=current_app.config['SYSTEM_WALLET_TARGET_BALANCE'],
             wei_topup_threshold=current_app.config['SYSTEM_WALLET_TOPUP_THRESHOLD'],
         )
+
+    def send_welcome_sms(self, to_user):
+        if self.custom_welcome_message_key:
+            message = i18n_for(to_user, "organisation.{}".format(self.custom_welcome_message_key))
+            send_message(to_user.phone, message)
```
wangfowen: any idea how i can mock out the `send_message` method here for a test? given it's not a method on the class itself but from elsewhere. if you look at the test below i did for kenya state machine, i was hoping to do the same for a test for this, but i feel like there's some pythonfoo that i'm lacking here

app/server/utils/ussd/kenya_ussd_state_machine.py
```diff
@@ -86,7 +86,6 @@ def change_preferred_language_to_en(self, user_input):
 
     def change_preferred_language_to(self, language):
         self.user.preferred_language = language
-        db.session.commit()
```
wangfowen: if i have this line in here, the tests give me errors saying:
```
sqlalchemy.exc.IntegrityError: (psycopg2.IntegrityError) null value in column "session_id" violates not-null constraint
```

I'm guessing because it's trying to create the factories..? if i don't do this line, the user change still sticks for the test. does that mean it also stuck for good? i'm still a bit confused when we need to commit given there's that magic code somewhere that makes committing not necessary

PhilipWafula: I am running into this very error with my tests.

PhilipWafula: @enjeyw any idea what the way out here would be ?

app/server/utils/i18n.py
```diff
@@ -1,11 +1,8 @@
 import i18n
-from typing import Optional
-
-from server.models.user import User
 
 
 # kwargs are for placeholders
-def i18n_for(user: Optional[User], key: str, **kwargs) -> str:
```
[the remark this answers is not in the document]
enjeyw: answer: stop using python :P

in all seriousness, if u check the stack trace to see where the circular import is, i've often found that changing the OTHER import helps...

app/server/utils/ussd/kenya_ussd_state_machine.py
enjeyw: >  i'm still a bit confused when we need to commit given there's that magic code somewhere that makes committing not necessary

Oh that magic code is only for requests going the the REST API - these changes still need a commit. 

> does that mean it also stuck for good?

Nope, just for the duration of that test module. The following code in conftest tears down all ur changes:
```

@pytest.fixture(scope='module')
def init_database():
    # Create the database and the database table

    with current_app.app_context():
        db.create_all()  # todo- use manage.py

    yield db  # this is where the testing happens!

    with current_app.app_context():
        db.session.close_all()  # DO NOT DELETE THIS LINE. We need to close sessions before dropping tables.
        db.drop_all()
```

enjeyw: at it's most basic, this is failing because the column session_id needs a value. @PhilipWafula have u tried just inserting a random string to make it happy?

PhilipWafula: Yeah I fed it a value same error.

app/client/components/form/SelectField.tsx
```diff
@@ -0,0 +1,51 @@
+import React from 'react'
+import {Field} from 'redux-form'
+
+import {WrappedInput} from './WrappedInput'
+import styles from "./styles.module.css";
+import FormValidation from "./FormValidation";
+
+interface SelectFieldJson {
```
tristanhcole: Are we exclusively using TypeScript now?

app/client/components/user/CreateUserForm.tsx
```diff
@@ -1,66 +1,73 @@
...
-    } else {
-      uniqueIdentifierInput = <Field name="blockchainAddress" component={TextInput} type="text" label='Bitcoin Address' />
-    }
-
+    let businessUsage;
 
     if (window.MAXIMUM_CUSTOM_INITIAL_DISBURSEMENT > 0) {
-      initialDisbursementAmount = <Field name="customInitialDisbursement" component={TextInput} type="text" label={'Initial Disbursement Amount'}>
+      initialDisbursementAmount = <InputField name="additionalInitialDisbursement" label={'Additional Initial Disbursement Amount'}>
         {window.CURRENCY_NAME}
```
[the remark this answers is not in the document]
enjeyw: weird, I thought it automatically detected the currency based on config. Can u swap it over to reference the token symbol?

app/server/utils/ussd/kenya_ussd_state_machine.py
enjeyw: errrgg that's annoying. ummmm i'm at a loss unfortunately.

app/server/models/organisation.py
enjeyw: Perhaps we'd have a sms sender class that we inject as a dependency on app buildup, which we can swap to a mock for tests?

app/server/utils/ussd/kenya_ussd_state_machine.py
wangfowen: @PhilipWafula per https://stackoverflow.com/a/19664562, we need to do UssdSessionFactory.build instead of UssdSessionFactory()
i've fixed mine in this PR

wangfowen: build gives it an auto-incremented id so it doesn't error. but this begs the question how do we test if a db change actually stuck around vs we forgot to db.session.commit?

app/client/components/form/SelectField.tsx
wangfowen: i'd say incrementally get into it. if you're working on an existing file that'll be hard to convert, don't bother or wait til you get more experience in it :) but if you're doing something simple from scratch, would be worth giving it a shot. the more typescript the safer the code

app/server/utils/ussd/kenya_ussd_state_machine.py
wangfowen: oh wait i mis-read. we're supposed to use `create` and not `build`. `build` does NOT save it to the database. i think the problem is we didn't set session_id as the primary key. will fix right now

wangfowen: OKAY IGNORE EVERYTHING I SAID. it's an early morning, my brain hasn't fully booted. session_id is not a primary key, it's passed in from the USSD API but needs to be not null and needs to be unique, so i've defined it as a sequence on the factory (see `UssdSessionFactory`). there were other fields that had to be not null as well so i also faked those. NOTE @PhilipWafula: i've just defined `ussd_menu_id` as 1. for later more complicated tests we may need that to be tied to actual ussd menus, but it'll do for now.
what i don't get though is according to the docs, `UssdSessionFactory()` is the same as `UssdSessionFactory.create()`. in that case, it should be saving to the db already. but then how come the null constraint errors didn't happen until we had code that manually did a `db.session.create()` in the test?

wangfowen: just tried it with UssdSessionFactory.create... also doesn't error until our explicit db commit. whyyyy? their documentation says "returns a saved instance", but it clearly did not save properly if it had violated null constraint errors?? or is this some sqlalchemy or something else black magic where initial saves are okay to violate constraints...

enjeyw: Ohhhhh I think i can actually answer this one!

basically the state that u store in an SQLAlchemy session isn't actually validated against db constraints until you commit to db. it's not so much black magic of SQLA telling u it's ok, so much as it is SQLAbeing totally ignorant of what the db has done till then. The reasoning being it's basically impossible for SQLA to know for certain whether something violates constraints without talking to the db ("sure this key was unique when i last check the db state, but how do i know another agent hasn't added a row with the same key since then").

wangfowen: ooo makes sense!

app/client/components/user/CreateUserForm.tsx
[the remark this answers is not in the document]
wangfowen: it's based on a window variable which comes from a global config. yet another things to make org level i guess lol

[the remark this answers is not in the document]
enjeyw: #OrgLevelAllTheThings

app/migrations/versions/ac8716ecdc21_.py
```diff
@@ -0,0 +1,28 @@
...
+# revision identifiers, used by Alembic.
+revision = 'ac8716ecdc21'
+down_revision = 'e2af63b3a35e'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_unique_constraint(None, 'user', ['_phone'])
```
wangfowen: making `phone` unique per user. will this cause any issues? @enjeyw @tristanhcole

enjeyw: nope this seems reasonable to me!

app/server/models/organisation.py
enjeyw: @wangfowen Something like this?
https://github.com/teamsempo/SempoBlockchain/blob/44e21e44ece12e6ec8c797eef7eeb0ead9e8fb14/test/unit/conftest.py#L92
(I believe requires wrapping the sms utils in a class, as monkeypatching raw functions inside a module didn't work for me) 

Alternatively something like this?
https://github.com/teamsempo/SempoBlockchain/blob/88a01d45cf9d275bc002b6e3ff89051e8e8d6223/test/functional/app/me_api/test_me_misc.py#L61

PR #40, opened 2019-10-30
Added pin auth and related features.

PhilipWafula:


app/migrations/versions/d0088b4df549_.py
```diff
@@ -0,0 +1,30 @@
...
+revision = 'd0088b4df549'
+down_revision = 'b40055ad9587'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('user', sa.Column('_encrypted_pin', sa.String(length=128), nullable=True))
+    op.add_column('user', sa.Column('failed_pin_attempts', sa.Integer(), nullable=True))
```
wangfowen: hm, do defaults not register in `add_column`?

app/server/models/user.py
```diff
@@ -499,9 +502,16 @@ def is_resetting(self):
         is_resetting = len(self.password_reset_tokens) > 0
         return is_resetting
 
-    # TODO(ussd): change to a field once we figure out what's the deal with resetting
     def pin_failed_attempts(self):
```
wangfowen: we can just remove this method and change to the field in places that use it

app/server/utils/user.py
```diff
@@ -590,3 +590,36 @@ def send_phone_verification_message(to_phone, one_time_code):
...
+    # TODO [Philip]: Find out whether reset token should be deleted.
+    db.session.commit()
+    send_pin_change_messages(user, new_pin)
+
+
+def send_pin_change_messages(user: User, new_pin):
+    # TODO [Philip]: Find out if there is a better way to check whether commit succeeded
+    if user.pin == new_pin:
+        if user.preferred_language == 'sw_KE':
+            send_message(user.phone, 'PIN yako ya Sarafu Network haijabadilishwa. Piga simu 0799 100 200 kwa usaidizi.')
```
wangfowen: we now can internationalize messages with i18n. see the translations under /locale, and look for usages of `i18n_for`

app/server/models/user.py
PhilipWafula: Should be a field. I think there's some redundancy I need to remove here.

app/server/models/user.py
```diff
@@ -499,9 +502,16 @@ def is_resetting(self):
...
-        return 0
+        return self.failed_pin_attempts
+
+    @property
+    def pin(self):
+        return decrypt_string(self._encrypted_pin)
```
wangfowen: hmm per the discussion here: https://trello.com/c/ei0HPPNB/32-implement-change-users-pin
we should be storing it salted (see `hash_password`). we should not be able to decrypt someone's pin

app/server/utils/user.py
```diff
@@ -590,3 +590,36 @@ def send_phone_verification_message(to_phone, one_time_code):
...
+    user.pin = new_pin
+    user.is_activated = True
+    # TODO [Philip]: Find out whether reset token should be deleted.
+    db.session.commit()
+    send_pin_change_messages(user, new_pin)
+
+
+def send_pin_change_messages(user: User, new_pin):
+    # TODO [Philip]: Find out if there is a better way to check whether commit succeeded
+    if user.pin == new_pin:
```
wangfowen: hmm i don't think if commit failed they wouldn't match. i think it'd probably error out, so we'd want to rescue? @enjeyw probably knows better
also if it's salted, string compare like this wouldn't work

app/migrations/versions/d0088b4df549_.py
PhilipWafula: Just checked up on this and found that it does present an error.
So weed need to run some SQL there before setting nullable to false but I am not so sure how clean its is.
It would end up being something like:
`
op.add_column(
        'users', 
        sa.Column('failed_pin_attempts', sa.Integer(), nullable=True)
    )
    op.execute("UPDATE users SET failed_pin_attempts = false")
    op.alter_column('users', 'failed_pin_attempts', nullable=False)
`

app/server/utils/ussd/kenya_ussd_state_machine.py
```diff
@@ -124,7 +130,9 @@ def is_token_agent(self, user_input):
 
     # TODO: [Philip] Add community token when available
     def upsell_unregistered_recipient(self, user_input):
-        upsell_message = f"{self.user.first_name} {self.user.last_name} amejaribu kukutumia {self.user} lakini hujasajili. Tuma information yako: jina, nambari ya simu, kitambulisho, eneo, na aina ya biashara yako kwa 0757628885."
+        upsell_message = f"{self.user.first_name} {self.user.last_name} amejaribu kukutumia {self.user} lakini " \
```
wangfowen: we can also make this i18n now. see the part about placeholders: https://pypi.org/project/python-i18n/

app/server/utils/ussd/kenya_ussd_processor.py
```diff
@@ -71,10 +71,31 @@ def replace_vars(menu: UssdMenu, ussd_session: UssdSession, display_text: str, u
...
-                    replacements.append(['%remaining_attempts%', "You have {} attempts remaining.".format(3 - user.pin_failed_attempts)])
+                    replacements.append(['%remaining_attempts%',
+                                         "You have {} attempts remaining.".format(3 - user.pin_failed_attempts)])
             else:
                 replacements.append(['%remaining_attempts%', ''])
 
         return reduce(lambda text, r: text.replace(r[0], r[1]), replacements, display_text)
+
+    @staticmethod
+    def authorize_pin(user: User, pin):
```
wangfowen: is this used in the state machine? why is it here and not in the state machine?

app/server/utils/ussd/kenya_ussd_processor.py
```diff
@@ -71,10 +71,31 @@ def replace_vars(menu: UssdMenu, ussd_session: UssdSession, display_text: str, u
...
+        authorized = False
+        if user.pin_failed_attempts == 3:
+            return False
+        if user.pin_failed_attempts < 3:
+            authorized = user.pin = pin
```
wangfowen: `user.pin == pin`? but also this would change if we salt instead of encrypt

app/server/utils/ussd/kenya_ussd_processor.py
wangfowen: or it might make more sense in User if we plan on using it in places beyond just the state machine

wangfowen:
please implement the usage of the pin check as part of this ticket. all states that are gated by pin should work after this. tests also should pass

app/server/utils/ussd/kenya_ussd_processor.py
PhilipWafula: Might need to find out why it was put there in the first place but yes your suggestion does make sense. In the meantime let me work on the pin storage as salted and refactor a couple of the things you've brought up so that we can close on the pin auth feature.

app/migrations/versions/d0088b4df549_.py
wangfowen: https://www.programcreek.com/python/example/87970/alembic.op.add_column
looks like you can pass in a default. don't know why the auto-generated migration didn't have it but should be able to modify and add it in

app/server/utils/ussd/kenya_ussd_processor.py
wangfowen: sounds good, you can leave the auth / actual usage of the pin for the next ticket. as part of this one just do setting pin and seeing that it works in the state machine, and being able to change the pin and seeing that works in the state machine.

app/server/models/user.py
PhilipWafula: So considering that hash_password returns no value and basically just updates the db's `password_hash` field. I think it would be more prudent to refactor the following into their own methods to make them reusable:
`bcrypt.hashpw(password.encode(), bcrypt.gensalt()).decode()`
`bcrypt.checkpw(password.encode(), self.password_hash.encode())`
right ?

app/server/utils/user.py
```diff
@@ -222,14 +223,16 @@ def create_transfer_account_user(first_name=None, last_name=None, preferred_lang
         if current_app.config['AUTO_APPROVE_TRANSFER_ACCOUNTS'] and not is_self_sign_up:
             transfer_account.approve()
 
+    user.default_transfer_account_id = transfer_account.id
```
wangfowen: this was copied over from my PR right? it'll break master because the attr is defined on my PR

app/server/utils/user.py
```diff
@@ -514,18 +518,23 @@ def proccess_create_or_modify_user_request(attribute_dict,
...
         use_last_4_digits_of_id_as_initial_pin=use_last_4_digits_of_id_as_initial_pin,
         existing_transfer_account=existing_transfer_account,
-        is_beneficiary=is_beneficiary, is_vendor=is_vendor, is_self_sign_up=is_self_sign_up
+        is_beneficiary=is_beneficiary, is_vendor=is_vendor, is_self_sign_up=is_self_sign_up,
+        token=token
     )
 
-    custom_attributes = set_custom_attributes(
-        attribute_dict, user)
+    if attribute_dict.get('custom_attributes', None) is None:
```
wangfowen: uh, in general this file's changes should not be made...

app/server/utils/user.py
```diff
@@ -592,34 +601,24 @@ def send_phone_verification_message(to_phone, one_time_code):
         send_message(to_phone, reciever_message)
 
 
-def change_pin(user: User, new_pin):
-    user.pin = new_pin
-    db.session.commit()
-    send_pin_change_messages(user, new_pin)
+def send_sms(user, message_key):
+    message = i18n_for(user, "ussd.kenya.{}".format(message_key))
```
wangfowen: this helper function doesn't make sense in utils/user since the key is specific to ussd.kenya

app/server/utils/user.py
```diff
@@ -592,34 +601,24 @@ def send_phone_verification_message(to_phone, one_time_code):
...
-    else:
-        if user.preferred_language == 'sw_KE':
-            send_message(user.phone, 'PIN yako ya Sarafu Network haijabadilishwa. Piga simu 0799 100 200 kwa usaidizi.')
-        else:
-            send_message(user.phone,
-                         'Your Sarafu Network PIN change failed. Please call 0799 100 200 to for assistance.')
 
+def change_initial_pin(user: User, new_pin):
+    user.is_activated = True
+    change_pin(user, new_pin)
```
wangfowen: do we need to db.commit?

app/server/utils/user.py
```diff
@@ -592,34 +601,24 @@ def send_phone_verification_message(to_phone, one_time_code):
...
-    # TODO [Philip]: Find out whether reset token should be deleted.
-    db.session.commit()
-    send_pin_change_messages(user, new_pin)
-
+def change_pin(user, new_pin):
+    try:
+        user.pin = new_pin
+        db.session.commit()
+        send_sms(user, 'successful_pin_change_sms')
+    except InvalidRequestError:
```
wangfowen: this is the error that happens from a failed db.session.commit?

app/server/utils/user.py
PhilipWafula: Has to have been because all I remember doing here was adding the methods to send messages.

app/server/utils/ussd/kenya_ussd_processor.py
```diff
@@ -24,7 +24,7 @@ def process_request(session_id: str, user_input: str, user: User) -> UssdMenu:
         # new session
         else:
             if user.is_resetting():
-                if user.pin_failed_attempts() >= 3:
```
wangfowen: can remove the now useless code for `pin_failed_attempts`

app/server/utils/user.py
PhilipWafula: I am not on my machine atm but I can clean this up in 30 minutes or so.

app/server/utils/ussd/kenya_ussd_state_machine.py
```diff
@@ -116,6 +116,21 @@ def new_pins_match(self, user_input):
             pins_match = True
         return pins_match
 
+    def authorize_pin(self, pin):
+        authorized = False
+        if self.user.failed_pin_attempts == 3:
+            return False
+        if self.user.failed_pin_attempts < 3:
+            authorized = self.user.check_salt_hashed_password(pin, self.user.pin)
```
wangfowen: should have a method on User similar to `verify_password` instead of doing the `check_salt_hashed_password` here. below also

app/server/utils/user.py
PhilipWafula: Really getting sloppy here 🤦🤦. Cleanup should be up in 30-45 tops currently in transit.

app/server/utils/user.py
PhilipWafula: Yes that's the error it throws.

app/server/models/user.py
enjeyw: I really think we should consider just storing pins and PWs in the same field for now - it could keep the concept of "resetting" much simpler - means we don't need to think about whether they're resetting a pin or their password.

The argument against the above is that pins and PWs have different requirements (one is a set of digits, the other is utf-8), so perhaps need to have different treatment anyway.

test/unit/app/utils/test_kenya_state_machine.py
```diff
@@ -153,10 +165,40 @@ def test_change_language(test_client, init_database, session_factory, user_input
...
-    #assert user.custom_attributes.market_enabled = False
+    # state_machine.feed_char("0000")
+    # assert state_machine.state == "complete"
+    # assert user.custom_attributes.market_enabled = False
     # TODO(owen): assert sms received is opt_out_of_market_place_sms
     return
+
+
+@pytest.mark.parametrize("session_factory, user_input, expected, failed_pin_attempts", [
+    (send_token_pin_authorization_state, "1212", "send_token_pin_authorization", 1)
```
wangfowen: could combine this with `test_exit_pin_blocked`, making them different parameters. would also be good to to test User with two failed attempts: what happens if they input correct on third try, what happens if they input incorrect on third try

app/server/utils/user.py
enjeyw: u should just be able to use the `verify_password` method here.

app/server/utils/ussd/kenya_ussd_state_machine.py
PhilipWafula: I saw the method, but felt like what I was doing wasn't exactly checking a password  since the hashed value was a pin. So that's why I went with hashing and checking the hash. 
Added the two methods to complement each other.

test/unit/app/utils/test_kenya_state_machine.py
```diff
@@ -29,97 +32,106 @@
...
+                             # opt_out_of_market_place_pin_authorization state tests
+                             (opt_out_of_market_place_pin_authorization_state, pin_blocked_standard_user, "1111",
+                              "exit_pin_blocked"),
+                             # exchange_token state tests
+                             (exchange_token_state, standard_user, "1", "exchange_rate_pin_authorization"),
+                             (exchange_token_state, standard_user, "2", "exchange_token_agent_number_entry"),
+                             (exchange_token_state, standard_user, "3", "exit_invalid_menu_option"),
+                             # exchange_rate_pin_authorization state tests
+                             (exchange_rate_pin_authorization_state, standard_user, "0000", "complete"),
+                             (exchange_rate_pin_authorization_state, pin_blocked_standard_user, "1111",
```
wangfowen: this is great :D good job on the extensive tests

app/server/utils/ussd/kenya_ussd_state_machine.py
enjeyw: perhaps we sidebar some time for discussion about this on the call? auth is one of those sufficiently critical things that it warrents proper discussion lol

test/unit/app/utils/test_kenya_state_machine.py
PhilipWafula: Yeah sounds like a good idea, I'll add these with the next push.

app/server/utils/ussd/kenya_ussd_state_machine.py
PhilipWafula: Though something like verify pin would make sense. I'll look at this again.

app/server/models/user.py
wangfowen: oh right. i forgot about this: @PhilipWafula if a user is resetting their pin, we also need to create a reset token and `save_password_reset_token`, and when they successfully set the pin, `delete_password_reset_tokens`. this is because in the `KenyaUssdProcessor`, we check if the user is currently resetting via `is_resetting` to send them through the right flow

@enjeyw i do agree that two edge cases could arise from this intertwined reset flow. if user is trying to reset their password, if mid-way through they use ussd, it'll send them down the reset pin flow. similarly, if they do intend to reset their pin mid-way through password reset, it'll void the token from their password reset. these seem veerrrry unlikely though, and i'd argue that if we do combine pin and password, if we ever want to separate them again it'll be pretty difficult if not impossible. and i can see a user having both a pin and a password for TFA

wangfowen: or maybe it just makes sense to split up pin reset from password reset. pin reset being just a bool `is_resetting_pin`

app/server/utils/user.py
```diff
@@ -200,13 +200,14 @@ def create_transfer_account_user(
 
     if organisation:
         user.organisations.append(organisation)
+        user.default_organisation = organisation
```
wangfowen: was this here before?

app/server/models/user.py
wangfowen: @PhilipWafula see this comment

app/server/utils/user.py
PhilipWafula: So I may have done the cherry pickin thing wrong, there were so many breakages from the old owen/ussd_features branch, so I mirrored master's present state into my branch to resolve those issues.

app/server/utils/ussd/kenya_ussd_processor.py
```diff
@@ -23,7 +23,7 @@ def process_request(session_id: str, user_input: str, user: User) -> UssdMenu:
                 return UssdMenu.find_by_name(new_state)
         # new session
         else:
-            if user.is_resetting():
+            if user.is_resetting_pin():
```
wangfowen: hmm i don't think we need reset tokens for pin, since we don't ever use them to verify anything. all we ever need is to check if we're resetting the pin or not. this could just be done via a boolean that we set and unset. is there any situation we'd need a token for pin?

app/server/models/user.py
```diff
@@ -425,21 +426,32 @@ def decode_single_use_JWS(cls, token, required_type):
...
-        self.clear_expired_tokens()
+        self.clear_expired_password_reset_tokens()
         is_valid = password_reset_token in self.password_reset_tokens
         return is_valid
 
     def delete_password_reset_tokens(self):
         self.password_reset_tokens = []
 
-    def clear_expired_tokens(self):
+    def delete_pin_reset_tokens(self):
```
wangfowen: when is this triggered? don't see this used

app/server/models/user.py
```diff
@@ -425,21 +426,32 @@ def decode_single_use_JWS(cls, token, required_type):
...
-        current_reset_tokens = self.password_reset_tokens[:]
-        current_reset_tokens.append(password_reset_token)
+        self.clear_expired_password_reset_tokens()
+        current_password_reset_tokens = self.password_reset_tokens[:]
+        current_password_reset_tokens.append(password_reset_token)
         # set db value
-        self.password_reset_tokens = current_reset_tokens
+        self.password_reset_tokens = current_password_reset_tokens
+
+    def save_pin_reset_token(self, pin_reset_token):
```
wangfowen: this doesn't seem to be being used either right now. it'll be used in the manual admin reset, but a new user should also be set as resetting pin right? otherwise they wouldn't go down the initial pin setting? or am i mistaken

app/server/models/user.py
```diff
@@ -57,6 +57,9 @@ class User(ManyOrgBase, ModelBase):
     secret = db.Column(db.String())
     _TFA_secret = db.Column(db.String(128))
     TFA_enabled = db.Column(db.Boolean, default=False)
+    _encrypted_pin = db.Column(db.String())
```
enjeyw: can we call this `pin_hash` instead? (Because it's hashed rather than encrypted, and it matches the password terminology)

app/server/models/user.py
```diff
@@ -293,12 +297,22 @@ def update_last_seen_ts(self):
         else:
             self._last_seen = cur_time
 
+    @staticmethod
+    def salt_hash_password(password):
```
enjeyw: another renaming one (sorry lol), but perhaps rename to 'salt_hash_secret' and change the `password` fields to `secret` has well, given it now hashes both the pin and the password. (Same with check/verify below)

app/server/utils/ussd/kenya_ussd_processor.py
enjeyw: I can imagine there's a good chance we'd need it for the android app in the future, though:
- right now there's no way to reset a forgotten pin from the android phone. 
- I think normally the flow for reseting a pin on a flow is different from setting a password - for pin u normally just get a new one-time code via sms or something right?
@PhilipWafula was this what u meant?

enjeyw: one thing i'm not clear on @PhilipWafula @wangfowen is how pin reset flow is supposed to work on the USSD interface - is it something the user is able to do at all, or is an admin required to do something? I couldn't find any use of `change_current_pin` in the code (though i'm guessing that might just be because it's a TODO?)

wangfowen: there's a ticket for adding that to admin interface. right now it's not doable

PR #41, opened 2019-11-01
Ruben/business cat

RubenGeo:


app/server/models/user.py
```diff
@@ -506,6 +514,51 @@ def pin_failed_attempts(self):
...
+        defaul business categories
+        :return: list of most relevant transfer usage objects for the usage
+        """
+        '''
+
+        # I spent way too much time failing to figure out how to write
+        # this query using sqlalchemy if someone could point met how to translate
+        # this query to sqlalchemy please do. 
+        # Can it maybe use the paginate_query function?
+        sql = text('''
```
RubenGeo: Like the comment says if. If someone has a suggest of how to convert this to an sqlalchemy query

test/unit/app/test_models.py
```diff
@@ -159,6 +159,11 @@ def test_valid_tfa_token(authed_sempo_admin_user):
             is not None)
 
 
+def test_get_most_relevant_transfer_usage(authed_sempo_admin_user):
```
RubenGeo: This just checks if a list is returned Testing the complete functionality requires a lot of though. (Creating multiple user, transer accounts, credits transfers, rainbows more stuff?).

Is it worth it?

app/migrations/seed.py
```diff
@@ -257,32 +258,32 @@ def update_or_create_menu(name, description, parent_id=None):
...
+        {'name': 'Health', 'icon': 'message'},
+        {'name': 'General shop', 'icon': 'message'},
+        {'name': 'Environment', 'icon': 'message'},
+        {'name': 'Transport', 'icon': 'message'},
+        {'name': 'Labour', 'icon': 'message'},
+        {'name': 'Other', 'icon': 'message'},
+    ]
+    for business_category in business_categories:
+        try:
+            usage = TransferUsage.create_without_duplicate(
```
RubenGeo: Does it matter if you use functions from the app in a seed script?

app/migrations/seed_dev.py
```diff
@@ -0,0 +1,162 @@
...
+                                    recipient_user=random.choice(user_list),
+                                    token=token,
+                                    uuid=randomString())
+        db.session.add(transfer)
+        transfer_list.append(transfer)
+        i += 1
+    return transfer_list
+
+
+if __name__ == '__main__':
```
RubenGeo: So I just added seed for all the stuff that was needed to test the get_most_relevant_transfer_usage function and an admin user. If anyone has other suggestion I can add it to this pr or create a new pr for that later

app/server/models/transfer_usage.py
```diff
@@ -29,3 +31,13 @@ def icon(self, icon):
         if icon not in MATERIAL_COMMUNITY_ICONS:
             raise IconNotSupportedException('Icon {} not supported or found')
         self._icon = icon
+
+    @classmethod
+    def create_without_duplicate(cls, raw_name, icon, priority, default):
```
wangfowen: would making name a unique attribute + having the setter for name strip and upper not achieve the same thing?

app/server/api/transfer_usage_api.py
```diff
@@ -1,14 +1,32 @@
...
 transfer_usage_blueprint = Blueprint('transfer_usage', __name__)
 
 
 class TransferUsageAPI(MethodView):
+    @requires_auth(allowed_roles={'ADMIN': 'admin'})
+    def get(self):
+        transfer_usages = TransferUsage.query.order_by(
+            desc('default')).all()
+
+        print(transfer_usages_schema.dump(transfer_usages))
```
wangfowen: can delete

app/migrations/versions/2c4716931a43_.py
```diff
@@ -0,0 +1,28 @@
...
+# revision identifiers, used by Alembic.
+revision = '2c4716931a43'
+down_revision = '7c0e719eccd9'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('user', sa.Column('business_usage_id', sa.Integer(), nullable=True))
```
wangfowen: uh oh i added the same thing in my PR, we're going to have collisions :|

app/server/models/user.py
```diff
@@ -506,6 +514,51 @@ def pin_failed_attempts(self):
     def user_details(self):
         "{} {} {}".format(self.first_name, self.last_name, self.phone)
 
+    def get_most_relevant_transfer_usage(self):
```
wangfowen: should be plural?

app/server/models/user.py
wangfowen: i won't be of any help on sqlalchemy, but could we have done the rest as part of the sql query? `group_by u.business_usage_id` + `count`?

app/server/models/user.py
```diff
@@ -506,6 +514,51 @@ def pin_failed_attempts(self):
...
+        number_of_wanted_business_id = 9
+        c = Counter(transfer_usage_ids)
+        most_common = [key for key, val in c.most_common(number_of_wanted_business_id)]
+        # If less than the wanted nr of usage is found fill up the list with default
+        if len(most_common) < number_of_wanted_business_id:
+            most_relevant_ids = self.refine_with_default_transfer_usage(
+                most_common, number_of_wanted_business_id)
+        most_relevant_transer_usage = TransferUsage.query.filter(
+            TransferUsage.id.in_(most_relevant_ids)).all()
+        print(most_relevant_transer_usage)
```
wangfowen: can delete

app/migrations/versions/2c4716931a43_.py
wangfowen: oh wait i made it a custom attribute :D so we're good. any reason for it to be in user vs to be a custom attribute?

test/unit/app/test_models.py
wangfowen: what kind of manual testing did you do? i thiiink it's okay to not thoroughly test it since it's not super critical functionality

RubenGeo: I filled the database using seed_dev and checked if it returned what i expected -> So the top most used transfer filled up with the default

And i did a check if it just return the just the default when you have a new user.

app/migrations/versions/2c4716931a43_.py
RubenGeo: Yea, i thought this would be a common attributes so most user would have a business_usage_id and i thought custom attributes is more for the uncommon attributes that user can have. 

I also think it may make querying for business_usage_id faster. May also come in handy if we want to implment Wills solution of the (~4+ layers deep) tranfer usage. 

What do you think?

app/server/models/transfer_usage.py
RubenGeo: Good idea :)

app/migrations/seed.py
enjeyw: i think that's generally a totally fine idea...

app/server/models/user.py
enjeyw: joins are surprisingly annoying in SQLAlchemy. TBH leaving this in SQL is probably fine for the time being IMO

test/unit/app/test_models.py
enjeyw: yeah i think the effort-to-reward ratio for testing this whole thing doesn't make sense just yet. At some point we should probably build a fixture that creates a lot of transfers for testing other aspects, and at that point testing the rest of this will be easy (though not really a unit test) - maybe just throw in a TODO to remind us that this test is lightweight?

test/functional/app/api/test_transfer_usage_api.py
```diff
@@ -30,3 +30,32 @@ def test_transfer_usage_api(test_client, authed_sempo_admin_user,name,icon,trans
...
+    """
+    authed_sempo_admin_user.is_activated = True
+    authed_sempo_admin_user.TFA_enabled = True
+    authed_sempo_admin_user.set_held_role('ADMIN', 'admin')
+    auth_token = authed_sempo_admin_user.encode_auth_token().decode()
+    tfa_token = authed_sempo_admin_user.encode_TFA_token(9999).decode()
+
+    response = test_client.get('/api/transfer_usage/',
+                               headers=dict(
+                                   Authorization=auth_token + '|' + tfa_token, Accept='application/json'),
```
enjeyw: we actually have a helper function makes this a lot easier 
`from server.utils.auth import get_complete_auth_token`

Then all of the lines above, including activating etc _should_ be possible to delete.

app/migrations/seed_dev.py
```diff
@@ -0,0 +1,162 @@
...
+
+    print('Create a list of users with a different business usage id ')
+    user_list = create_users_different_transer_usage(15)
+
+    print('Creating transactions')
+    # User 1 sends to a random choice of user_list
+    create_transfers(user1, user_list, 30)
+
+    db.session.commit()
+    sql = text("""UPDATE credit_transfer SET transfer_status = 'COMPLETE'""")
```
enjeyw: does `credit_transfer.transfer_status = TransferStatusEnum.COMPLETE` (or something to that effect) not work?

RubenGeo: Yes that works I changed it

enjeyw:
^^ those errors are due to my config file changes; the eth_worker is looking for a local blockchain that doesn't exist

app/migrations/seed.py
```diff
@@ -257,32 +259,32 @@ def update_or_create_menu(name, description, parent_id=None):
...
+        {'name': 'Other', 'icon': 'message'},
+    ]
+    for business_category in business_categories:
+        try:
+            usage = TransferUsage(name=business_category['name'], icon=business_category['icon'], 
+            priority=1, default=True)
+            db.session.add(usage)
+            db.session.commit()
+                
+        except IntegrityError as e:
```
enjeyw: Because there's no check to see if the category already exists, this looks like it'd hit a 'duplicate name' IntegrityError over and over again if you run the seed more than once?

Perhaps it'd make more sense to: 
Check if a name already exists, and if so, update the date to reflect that's the in seed. This would make it easy to update translations, icons etc. 

Also, it might be good to include a priority in the seed data?

app/migrations/versions/2c4716931a43_.py
enjeyw: I _think_ this makes slightly more sense as an actual column rather than a custom attribute; while my natural inclination is to keep the user table as constrained to key information rather than supplementary data, extending on @RubenGeo 's point, given that business_id is a foreign key, it probably makes more sense to specify it as such.*

Actually on that note, the column definition above doesn't have it as a foreign key - perhaps this should be updated if we do keep it as a column?

app/server/api/transfer_usage_api.py
```diff
@@ -19,19 +35,23 @@ def post(self):
         icon = post_data.get('icon')
         priority = post_data.get('priority')
         # EG {'fr': 'fromage'}
-        translations = post_data.get('translations')
 
         if name is None or icon is None:
             return make_response(jsonify({'message': 'Name and Icon must not be None'})), 400
 
         try:
-            usage = TransferUsage(name=name, translations=translations, icon=icon, priority=priority)
```
enjeyw: What happened to translations? I don't think the general python i18n module handles them?

app/server/models/transfer_usage.py
```diff
@@ -29,3 +30,11 @@ def icon(self, icon):
...
             raise IconNotSupportedException('Icon {} not supported or found')
         self._icon = icon
+
+    @hybrid_property
+    def name(self):
+        return self._name
+
+    @name.setter
+    def name(self, name):
+        self._name = name.strip().upper()
```
enjeyw: what's the reasoning around why we coerce this into uppercase before storing? That means if we use the name anywhere on the UX it's be necessarily uppercase...

If it's for testing for duplicates, we could just do that using a filter with an extra hybrid property such as name_lower?

test/functional/app/api/test_transfer_usage_api.py
```diff
@@ -30,3 +31,29 @@ def test_transfer_usage_api(test_client, authed_sempo_admin_user,name,icon,trans
                                 content_type='application/json', follow_redirects=True)
 
     assert response.status_code == status_code
+
+    # Ensure that the posting a duplicate transfer ussage results in a 400
+    response = test_client.post('/api/transfer_usage/',
+                                headers=dict(Authorization=auth_token + '|' + tfa_token, Accept='application/json'),
```
enjeyw: i may have commented on this elsewhere, or maybe it was someone else's PR, but there's a utility function for making this easier: `get_complete_auth_token` in utils > auth.py

app/migrations/seed.py
```diff
@@ -257,32 +259,32 @@ def update_or_create_menu(name, description, parent_id=None):
...
+            print(e)
```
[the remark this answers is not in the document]
wangfowen: @RubenGeo if you haven't made this change yet, i wrote some code for find_or_create in this PR: https://github.com/teamsempo/SempoBlockchain/pull/44/commits/88541f8ab5d6f786db2ed3e5007c56806fd82ece#diff-c92355449c362bbdb8e67b8d67d6a8f8R43

app/server/models/transfer_usage.py
wangfowen: right, to prevent dupes when inserting new ones. I think coercing them to all be uniform when setting isn't so bad, but maybe instead of upper, it should be capitalized?

enjeyw: Capitalised would be better, but i'm worried about unnecessarily stripping out information - what happens if a category contained an acronym like "WASH Supplies"? (WAter, Sanitation and Healthcare)

RubenGeo: Okay, shall I remove the capitalizing from this method? Instead I can add something like this in the setter method: `SELECT * 
FROM public.transfer_usage where lower('food') = lower(_name)` and then see if it return one and throw an error if does if it returns nothing insert into the database. So that way we can ensure the the capitalizing remains the way it's intended by the user and still prevent dupes.

RubenGeo: We only take the risk that a user really want to add a WASH as a different category then wash this is prevented. But that really seems like an edge case

app/server/api/transfer_usage_api.py
RubenGeo: Can they be removed?

app/migrations/seed.py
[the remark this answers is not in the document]
RubenGeo: Cool it would  make sense to use that function then. How would I go about integrating it in my branch? Just copy paste it? Or will that create conflicts?

app/migrations/versions/2c4716931a43_.py
RubenGeo: Good point, i updated this to a foreign key :+1:

app/migrations/seed.py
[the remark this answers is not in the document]
RubenGeo: @enjeyw About the priority, to be honest I don't know what it's used for. What should i change it to?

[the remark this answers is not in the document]
wangfowen: yeah, just copy paste it and i'll fix the conflicts later

app/server/api/transfer_usage_api.py
enjeyw: nah we still need them for I18n on the android app

app/server/models/transfer_usage.py
enjeyw: > But that really seems like an edge case

Hahaha yeah agreed lol! It should be possible to code this up without writing raw SQL...

app/server/models/user.py
tristanhcole: `WHERE sender_user_id = 4 AND c.transfer_status = 'COMPLETE'`
Why is there sender_user_id = 4? Seems like an error?

tristanhcole: @RubenGeo @wangfowen @enjeyw

RubenGeo: You are totally right! I a pull to fix this: https://github.com/teamsempo/SempoBlockchain/pull/52

PR #42, opened 2019-11-04
Contract launcher

enjeyw:
Opening up a PR to see if all the new docker tests work!

This kinda turned into a bit of a yak-sheering exercise to get all the blockchain stuff tested properly


app/server/__init__.py
```diff
@@ -206,4 +207,9 @@ def encrypt_string(raw_string):
 africastalking.initialize(config.AT_USERNAME, config.AT_API_KEY)
 africastalking_client = africastalking.SMS
```
enjeyw: shifted all the instantiations down to the bottom of the file, rather having them disbursed throughout function defs

app/server/api/token_api.py
```diff
@@ -46,11 +53,55 @@ def post(self):
...
+                reserve_token_address=exchange_contract.reserve_token.address,
+                reserve_ratio_ppm=reserve_ratio_ppm
+            )
+
+            address = smart_token_result['smart_token_address']
+            subexchange_address = smart_token_result['subexchange_address']
 
+        token = Token(address=address, name=name, symbol=symbol)
         db.session.add(token)
         db.session.commit()
```
enjeyw: this commit appears to be necessary for some reason, otherwise  `exchange_contract.add_token(token, subexchange_address, reserve_ratio_ppm)` fails. not sure why

app/server/me_api/exchange.py
```diff
@@ -52,26 +53,45 @@ def post(self):
...
+
+                response_object = {
+                    'message': str(e),
+                }
+                return make_response(jsonify(response_object)), 400
 
         elif to_desired_amount:
 
-            exchange = Exchange()
+            # TODO: Test this
```
enjeyw: don't know if this works yet. TBC

app/server/models/credit_transfer.py
```diff
@@ -28,7 +24,7 @@ class CreditTransfer(ManyOrgBase, BlockchainTaskableBase):
     uuid            = db.Column(db.String, unique=True)
 
     resolved_date   = db.Column(db.DateTime)
-    transfer_amount = db.Column(db.Integer)
+    _transfer_amount = db.Column(db.BigInteger)
```
enjeyw: transfer amounts on the blockchain have 256 bits of precision - needing to store some sub $0.01 amounts for exchanges now, so this gets us some of the way there. Problem is that even bigints cant handle the full range of the blockchain-native amounts, so we can't just store them across 1:1

app/server/models/exchange.py
```diff
@@ -40,7 +41,27 @@ class ExchangeContract(ModelBase):
         lazy=True, foreign_keys='TransferAccount.exchange_contract_id'
     )
 
-    def add_token(self, token):
+    def get_subexchange_details(self, token_address):
+        if self.subexchange_address_mapping is None:
+            return None
+        return self.subexchange_address_mapping.get(token_address, None)
+
+    def _add_subexchange(self, token_address, subexchange_address, subexchange_reserve_ratio_ppm):
```
enjeyw: given the 'subexchange' is technically another smart contract, we could store these as objects in their own rights, but the relationships would become pretty complicated

app/server/api/contract_api.py
```diff
@@ -0,0 +1,118 @@
...
+
+        smart_token2 = Token(address=smart_token_address2, name='Goop2', symbol='Goop2')
+        smart_token2.decimals = decimals
+        db.session.add(smart_token2)
+
+        exchange_contract.add_token(smart_token2, converter_address2, reserve_ratio_ppm)
+
+        db.session.commit()
+
+        response_object = {
```
enjeyw: should probably properly serialise these at some point

enjeyw:
Yessss passing!

app/server/utils/blockchain_tasks.py
```diff
@@ -1,280 +1,446 @@
...
+    bonding_curve_token1_to_token2
+)
 
-    return response
 
-def execute_synchronous_transaction_task(signature):
-    if current_app.config['IS_TEST']:
-        # TODO: We need a better way of stubbing responses from the blockchain worker during tests
-        return random.randint(0, 1000000000)
+class BlockchainTasker(object):
```
enjeyw: shifted all of these into a class to make it easier to mock during testing with monkeypatching

app/server/utils/misc_tasks.py
```diff
@@ -0,0 +1,8 @@
+from server import celery_app
+
+class MiscTasker(object):
```
enjeyw: moving all async tasks into classes to aid with monkeypatching during tests (as above)

celery_tester.py
```diff
@@ -0,0 +1,38 @@
...
+# engine = create_engine(config.ETH_DATABASE_URI)
+# eth_models.ModelBase.metadata.drop_all(engine)
+# eth_models.ModelBase.metadata.create_all(engine)
+
+p2 = subprocess.Popen(["celery", "-A", "eth_manager", "worker",
+                       "--loglevel=INFO", "--concurrency=4", "--pool=eventlet"], cwd="./eth_worker")
+
+
+p3 = subprocess.Popen(["celery", "-A", "eth_manager", "worker",
+                       "--loglevel=INFO", "--concurrency=1", "--pool=eventlet", "-Q=processor"], cwd="./eth_worker")
```
enjeyw: this worker has a concurrency of 1, so tasks should necessarily be processed sequentially (this is important because tasks need to be submitted to the blockchain (almost) in order), but for some reason occasionally the tasks progress through the worker at the same time, which results in a race condition and i have LITERALLY NO IDEA WHY.

celery_tester.py
```diff
@@ -0,0 +1,38 @@
+import pytest
```
enjeyw: helper function for launching celery for testings. only works some of the time due to this issue below. if u do the two steps manually rather than launching as a subprocess it works fine, but requires manual setup

eth_worker/eth_manager/celery_tasks.py
```diff
@@ -23,40 +23,75 @@ def after_return(self, status, retval, task_id, args, kwargs, einfo):
...
 def topup_wallets(self):
-    higher_order_tasks.topup_wallets()
+    eth_manager.task_interfaces.composite.topup_wallets()
 
 
-@celery_app.task(**task_config)
-def topup_wallet_if_required(self, address, wei_target_balance, wei_topup_threshold):
-    higher_order_tasks.topup_if_required(address, wei_target_balance, wei_topup_threshold)
+@celery_app.task(**base_task_config)
+def topup_wallet_if_required(self, address):
```
enjeyw: given these methods pass straight through to functions of with identical signatures, we could probably just add the celery decorator to the original functions, and refactor them out to be in an another celery tasks module. will do another time

test/unit/conftest.py
```diff
@@ -0,0 +1,137 @@
...
+
+def mock_class(orginal, substitute, monkey):
+    for func_name in dir(orginal):
+        if callable(getattr(orginal, func_name)) and not func_name.startswith("_"):
+            func = getattr(substitute, func_name, None)
+
+            if func is None:
+                raise Exception("Function {} doesn't have a mock definition in {}!".format(func_name, substitute))
+
+            monkey.setattr(orginal, func_name, func)
```
enjeyw: it'd be nice to be able to move this up a directory so that the functional test can access it, but apparently you're not supposed to turn the parent directories in pytest into modules, so i don't know what to do here...

enjeyw:
Ohhh it's worth mentioning that the functional tests now require celery AND a local test blockchain  to be up and running, while the unit tests have some stuff mocked to make them more unit-y. (ideally they wouldn't even need a db?)

Getting the test blockchain and celery to run properly without just using docker is a pain, so i guess two options for deving are:
- skip the functional tests on local machines?
- optionally mock the blockchain components on functional tests that are being run locally?

enjeyw:
[Model for Recording an Exchange of Tokens](https://trello.com/c/uwpuujR2/75-model-for-recording-an-exchange-of-tokens)

enjeyw:
[Model for Recording an Exchange of Tokens](https://trello.com/c/uwpuujR2/75-model-for-recording-an-exchange-of-tokens)

enjeyw:
[Make blockchain worker endpoint for exchanging tokens](https://trello.com/c/ngtnJX34/9-make-blockchain-worker-endpoint-for-exchanging-tokens)

enjeyw:
[Testing Framework for Blockchain Worker](https://trello.com/c/AysHPTZn/107-testing-framework-for-blockchain-worker)

enjeyw:
[Testing Framework for Blockchain Worker](https://trello.com/c/AysHPTZn/107-testing-framework-for-blockchain-worker)

enjeyw:
@tristan my migrations were TOTALLY messed up so I just deleted them all and took the ones in master. The files above might appear empty because there’s no difference between them and the master branch?

tristanhcole:
Ah ok, yes that makes sense now.

app/server/api/contract_api.py
```diff
@@ -0,0 +1,118 @@
...
+deploy_contracts_blueprint = Blueprint('contracts', __name__)
+
+
+class ContractsAPI(MethodView):
+
+    @requires_auth(allowed_roles={'ADMIN': 'sempoadmin'})
+    def post(self):
+        # post_data = request.get_json()
+
+        name = 'CIC Test 2'
```
tristanhcole: Placeholder data?

app/server/models/credit_transfer.py
tristanhcole: [pg-bignum](https://github.com/beargiles/pg-bignum) sounds like a valid option, alternatively, we could store in `numeric` and do a conversion in application code?

https://dba.stackexchange.com/questions/62934/adding-unsigned-256-bit-integers-in-postgresql/192746

app/server/api/contract_api.py
enjeyw: Oh good catch. all of these need to be swapped to pull from POST request.

Didn't get picked up during writing tests because the tests only test the use of the exchange once it's deployed, rather than the creation of the exchange itself.

Also, deploying a token contract is done automatically when u post a token. for consistency i'm thinking i should either:
- delete the contract_api, and move this to a POST on /api/exchange/contract instead. 
- make it so that POSTING to /api/token/ only registers an existing token, which right now is achieved by providing an address when posting (we definitely need this functionality, as a lot of the tokens we use will be external ones like DAI), and moving all contract deployment to /api/contract/token, /api/contract/exchange etc etc....

Thoughts?

app/server/models/credit_transfer.py
enjeyw: mmmmm yeah i did look in `numeric`, though I read somewhere that performance aint fantastic. I guess it would make it FAR easier to not need to store amounts at three levels of precision like we are now:
- int cents on the front end, effectively 2 decimals
- 256 bit int, effectively 18 decimals in contracts
- big int, effectively 6 + 2 = 8 decimals on the db
 🤢

enjeyw: from https://www.postgresql.org/docs/10/datatype-numeric.html
on one hand...
> calculations on numeric values are very slow compared to the integer types, or to the floating-point types described in the next section.

on the other

>  It is especially recommended for storing monetary amounts

enjeyw: thoughts @tristanhcole and @wangfowen? Seems to be a bit of debate about this on the web:
https://dba.stackexchange.com/questions/110880/numeric-vs-integer-for-a-column-size-and-performance

i'm inclined to go towards using numeric, because we've been stung by amounts not adding to 0 on contracts before, but about 49-51 split.

test/unit/conftest.py
tristanhcole: Can you not just move this code into the `conftest.py` file located in the test directory? May need to just update the pytest fixture scope as necessary

tristanhcole: Also, we _can_ turn pytest directories into modules, just not recommended (and we've gotten this far without)...
http://doc.pytest.org/en/latest/goodpractices.html

eth_worker/eth_manager/task_interfaces/composite.py
```diff
@@ -0,0 +1,249 @@
...
+        args=[subexchange_address, 10],
+        gas_limit=100000
+    )
+
+    transaction_task(
+        signing_address=deploying_address,
+        contract_address=reserve_token_address,
+        contract_type='EtherToken',
+        func='approve',
+        args=[subexchange_address, 1000000*10*18],
```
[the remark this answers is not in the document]
enjeyw: ohhh yesss that's very wrong

app/server/api/organisation_api.py
```diff
@@ -76,6 +75,7 @@ def post(self, organisation_id):
 
         new_organisation = Organisation(name=name, token=token)
 
+        # TODO: Shift this all into being created when an org is created
```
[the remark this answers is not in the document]
enjeyw: yeah i actually put it in there, but then moved it back out again because from memory there were some weird side effects i ran across when originally writing this section, and i didn't wanna change it without testing properly

app/server/api/token_api.py
enjeyw: You can normally make all kinds of queries without needing to persist to db though, as it just checks the orm session...? Like you can paste this pretty much anywhere an it works:

```
        u = TransferUsage(name='Foobar')
        db.session.add(u)
        found_u = TransferUsage.query.filter_by(name='Foobar').first()
```

ur right, should probably be a flush

app/server/models/exchange.py
enjeyw: yeah exactly -  token exchanges are actually made up of a network of linked smart contracts, and u need to know some details of those contracts

celery_tester.py
enjeyw: it works fine 99% of the time on CI/CD: because Circle uses docker compose, the CI/CD tests run on a far more sophisticated worker setup that includes dedicated workers for txn processing

test/unit/conftest.py
enjeyw: I couldn't get it to import from conftest.py. Actually fixed it just putting it in the same folder as fixtures

app/server/models/exchange.py
```diff
@@ -28,6 +25,10 @@ class ExchangeContract(ModelBase):
 
     blockchain_address = db.Column(db.String(), index=True)
 
+    contract_registry_blockchain_address = db.Column(db.String(), index=True)
+
+    subexchange_address_mapping = db.Column(JSON)
```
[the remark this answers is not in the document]
enjeyw: due to above reasoning due to complexity of setting up the internal relationship. AFAIK we only need to get the subexchange address from the primary exchange, so this keeps things nice and simple

eth_worker/eth_manager/task_interfaces/composite.py
```diff
@@ -0,0 +1,249 @@
...
+
+        persistence_interface.set_wallet_last_topup_task_id(address, task_id)
+
+        return task_id
+
+    return None
+
+
+def deploy_exchange_network(deploying_address):
+    gasPrice = 250000000000
```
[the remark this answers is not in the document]
enjeyw: yeah good point, though need to remember not do to 25*10*10 like I did before 😆

enjeyw:
ok cool @tristanhcole - ill fix those errors, and I'll modify tests so that the func tests automatically mock the blockchain if the system ISN'T in a docker container - it's already super annoying not being able to use the func tests locally 😠

eth_worker/eth_manager/task_interfaces/composite.py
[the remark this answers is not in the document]
enjeyw: int(8e7) is pretty foolproof, even for me!

app/server/api/contract_api.py
enjeyw: update: took option 2! saves having token registration/creation squeezed into one endpoint

app/migrations/versions/2460fe231511_.py
```diff
@@ -0,0 +1,94 @@
...
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic.
+revision = '2460fe231511'
+down_revision = '18c8798add0d'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
```
wangfowen: thought this PR was for tests? why is exchange stuff also in it?

app/server/models/credit_transfer.py
wangfowen: beyond basic addition and subtraction, would there be extensive calculations? i haven't found any good benchmarks of exactly how much slower numeric is, but i can't imagine the added slowness causing that much of a difference. if storage of the number is less error prone i'm for it

app/server/models/credit_transfer.py
```diff
@@ -63,6 +59,15 @@ class CreditTransfer(ManyOrgBase, BlockchainTaskableBase):
     to_exchange = db.relationship('Exchange', backref='to_transfer', lazy=True, uselist=False,
                                   foreign_keys='Exchange.to_transfer_id')
 
+    # TODO: Apply this to all transfer amounts/balances, work out the correct denominator size
+    @hybrid_property
+    def transfer_amount(self):
+        return self._transfer_amount / 10**6
```
wangfowen: not a fan of this naming. would prefer it being more precise so we know what we're working with. `transfer_amount_cents` instead of just `_transfer_amount` for example, but whatever the terminology is with crypto

app/migrations/versions/2460fe231511_.py
wangfowen: in the future, for bigger PRs, can we add a description of what are the major changes in the PR? easier for people to get context for reviewing

app/server/models/exchange.py
wangfowen: yeahh i didn't understand subexchange either originally. is there a better terminology for it? or maybe just add a comment explaining it

app/server/models/exchange.py
```diff
@@ -40,7 +41,27 @@ class ExchangeContract(ModelBase):
...
+        if self.subexchange_address_mapping is None:
+            return None
+        return self.subexchange_address_mapping.get(token_address, None)
+
+    def _add_subexchange(self, token_address, subexchange_address, subexchange_reserve_ratio_ppm):
+        if self.subexchange_address_mapping is None:
+            self.subexchange_address_mapping = {}
+        self.subexchange_address_mapping[token_address] = {
+            'subexchange_address': subexchange_address,
+            'subexchange_reserve_ratio_ppm': subexchange_reserve_ratio_ppm
```
wangfowen: subexchanges are pegged at certain ratios? thought it'd fluctuate based on changes in the currency's different bonding curves, etc

app/server/models/exchange.py
```diff
@@ -187,7 +219,7 @@ def exchangeable_by_contract(token, contract):
 
     def _get_conversion_function(self, exchange_contract, from_token, to_token):
         return partial(
-            get_conversion_amount,
+            bt.get_conversion_amount,
```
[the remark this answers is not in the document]
wangfowen: +1 to more explicit naming. db i think is pretty common knowledge to being short for database, whereas when i first saw bt i thought bluetooth. i still haven't gotten to the part of the PR defining what bt is yet and i still don't know what it actually stands for, but from what i'm seeing it's methods related to sending stuff to the blockchain worker..?

app/server/models/user.py
```diff
@@ -11,11 +11,13 @@
 import random
 import string
 
+from server import bt
```
wangfowen: already declared above

app/server/utils/blockchain_tasks.py
wangfowen: ah bt = blockchain tasker :)

app/server/utils/misc_tasks.py
wangfowen: not a fan of throwing random things together under the category of misc :/

eth_worker/eth_manager/stripped_compiled_contracts/XTransferRerouter.json
```diff
@@ -0,0 +1 @@
...00604482015290519081900360640190fd5b60018054911515740100000000000000000000000000000000000000000274ff000000000000000000000000000000000000000019909216919091179055565b600154600160a060020a031681565b60015474010000000000000000000000000000000000000000900460ff16151561036d576040805160e560020a62461bcd02815260206004820152600660248201527f3034376636620000000000000000000000000000000000000000000000000000604482015290519081900360640190fd5b6040805183815260208101839052815185927fb5c80f971fb729e469ffa874c60425659ce82cb4adcfba9731af35ef87b6e619928290030190a2505050565b60015474010000000000000000000000000000000000000000900460ff1681565b600054600160a060020a0316331461042f576040805160e560020a62461bcd02815260206004820152600e60248201527f4d757374206265206f776e65722e000000000000000000000000000000000000604482015290519081900360640190fd5b600054600160a060020a0382811691161415610495576040805160e560020a62461bcd02815260206004820152600660248201527f3263616436300000000000000000000000000000000000000000000000000000604482015290519081900360640190fd5b6001805473ffffffffffffffffffffffffffffffffffffffff1916600160a060020a03929092169190911790555600a165627a7a72305820cf1537437b5eb0897e6822948f68f241b8899c3a9bab7b0dfc5ede13b4644ea90029"}
```
wangfowen: much contracts

eth_worker/eth_manager/stripped_compiled_contracts/BancorConverter.json
```diff
@@ -0,0 +1 @@
...54f057600080fd5b505af1158015615504573d6000803e3d6000fd5b505050506040513d602081101561551a57600080fd5b5051604080517faeea10bd000000000000000000000000000000000000000000000000000000008152600160a060020a03898116600483015291519295509085169163aeea10bd916024808201926020929091908290030181600087803b15801561558457600080fd5b505af1158015615598573d6000803e3d6000fd5b505050506040513d60208110156155ae57600080fd5b5051156157595785600160a060020a03166370a08231866040518263ffffffff1660e060020a0281526004018082600160a060020a0316600160a060020a03168152602001915050602060405180830381600087803b15801561561057600080fd5b505af1158015615624573d6000803e3d6000fd5b505050506040513d602081101561563a57600080fd5b5051604080517fa9059cbb000000000000000000000000000000000000000000000000000000008152600160a060020a0388811660048301526024820188905291519294509088169163a9059cbb9160448082019260009290919082900301818387803b1580156156aa57600080fd5b505af11580156156be573d6000803e3d6000fd5b5050505085600160a060020a03166370a08231866040518263ffffffff1660e060020a0281526004018082600160a060020a0316600160a060020a03168152602001915050602060405180830381600087803b15801561571d57600080fd5b505af1158015615731573d6000803e3d6000fd5b505050506040513d60208110156157
```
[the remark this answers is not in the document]
enjeyw: Yeah it's auto-generated based off a some Solidity Code that's in another repo. No need to edit it within codebase.

app/server/models/credit_transfer.py
enjeyw: yeah good point. How about we
- Go to storing full 18 decimal precision data using the `numeric` datatype
- call this `transfer_amount_wei`, because that's unambiguously 1 * 10^-18 USD/ETH

When passing around amounts use one of
a) Wei (will low power android devices die if they have to store 18+ sig figs?)
b) cents
c) float

@wangfowen @tristanhcole

app/server/models/credit_transfer.py
enjeyw: Ok let's do numeric!

app/server/models/exchange.py
enjeyw: the `reserve_ratio` is what determines the extent of fluctuation with respect to changes in demand - i'll link the derivation in slack!

app/server/models/credit_transfer.py
enjeyw: going to swap the amounts stored on the db to full precision right now because that's a relatively small change, and we can work out to do with the other locations later because there's implications for data transfer etc.

app/server/utils/misc_tasks.py
enjeyw: yeah good point. what do you thing makes more sense? it technically means "all async tasks that aren't blockchain ones"

app/migrations/versions/2460fe231511_.py
enjeyw: yeah will do. hopefully there won't be many big prs like this in the future. also this migration file i think just didn't get commited properly last time

enjeyw:
todos:

- [ ] change to numeric datatype
- [ ] add comment explaining subexchange
- [] duplicate bt declaration in models/user

app/server/models/exchange.py
enjeyw: added a comment at the top of the class to explain :)

PR #44, opened 2019-11-06
hook in business category to admin user creation

wangfowen:
- fix accounts not showing up on accounts page
- reducers and saga for business categories
- also reducers for organisations, not used yet
- hook in reducer to user form
- other selection

### DIFF
--- a/app/client/components/transferAccount/transferAccountList.jsx
+++ b/app/client/components/transferAccount/transferAccountList.jsx
@@ -175,10 +175,10 @@ class TransferAccountList extends React.Component {
     if (this.state.account_type !== 'ALL') {
       // a vendor/recipient filter is applied
       if (this.state.account_type === 'VENDORS') {
-        filteredData = filteredData.filter(account => (account.is_vendor === true))
+        filteredData = filteredData.filter(account => account.is_vendor)
       }
       if (this.state.account_type === window.BENEFICIARY_TERM_PLURAL) {
-        filteredData = filteredData.filter(account => (account.is_beneficiary === true))
+        filteredData = filteredData.filter(account => !account.is_vendor)
       }
     }
 
--- a/app/client/components/user/CreateUser.tsx
+++ b/app/client/components/user/CreateUser.tsx
@@ -6,15 +6,19 @@ import { StyledButton, ModuleHeader } from '../styledElements'
 import styles from './styles.module.css';
 import CreateUserForm, {ICreateUser} from './CreateUserForm';
 import CreateVendorForm, {ICreateVendor} from './CreateVendorForm';
+import { loadTransferUsages } from '../../reducers/transferUsage/actions'
+import {TransferUsage} from "../../reducers/transferUsage/types";
+import {ReduxState} from "../../reducers/rootReducer";
 
-//TODO(ts): fix types
 interface DispatchProps {
   createUser: (body: any) => void,
   resetCreateUser: () => void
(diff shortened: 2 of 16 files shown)


app/client/components/transferAccount/transferAccountList.jsx
```diff
@@ -175,10 +175,10 @@ class TransferAccountList extends React.Component {
     if (this.state.account_type !== 'ALL') {
       // a vendor/recipient filter is applied
       if (this.state.account_type === 'VENDORS') {
-        filteredData = filteredData.filter(account => (account.is_vendor === true))
+        filteredData = filteredData.filter(account => account.is_vendor)
       }
       if (this.state.account_type === window.BENEFICIARY_TERM_PLURAL) {
-        filteredData = filteredData.filter(account => (account.is_beneficiary === true))
+        filteredData = filteredData.filter(account => !account.is_vendor)
```
wangfowen: this is the fix for users not showing up correctly on accounts page. have we removed `is_beneficiary`? didn't see it in the code

app/client/components/user/CreateUser.tsx
```diff
@@ -107,23 +116,27 @@ class CreateUser extends React.Component<Props> {
...
           transferAccountType={transferAccountType}
+          transferUsages={this.props.transferUsages}
           onSubmit={(form: Form) => this.onCreateUser(form)}
         />
       }
     }
   }
 }
 
+//TODO: why doesn't ReduxState work correctly
```
wangfowen: might be because there's a mix of ts and js reducers..? but it means we can't get the full benefits of typescript with react/redux here, which is unfortunate

app/client/reducers/organisation/reducers.ts
```diff
@@ -0,0 +1,50 @@
+import {combineReducers} from "redux";
```
wangfowen: for this and the other reducer i added, i changed up how the files are organized, but i used the same reducer+saga format as the other reducers. should i be changing it to something else? you said there was a better format in the mobile code?

app/client/sagas/transferUsageSagas.js
```diff
@@ -0,0 +1,49 @@
+import { put, takeEvery, call, all } from 'redux-saga/effects'
```
wangfowen: i have no idea how to do reducers for typescript. gave it a try, it was messy, gave up for now and left it as js

app/client/components/transferAccount/transferAccountList.jsx
enjeyw: It shouldn't have been removed *yet*, though i don't love that the beneficiary type is defined on both the user and on the transfer_account, and they both have slightly different meanings.  This looks ok for now, and then we can work out a proper treatment later

app/client/reducers/organisation/reducers.ts
enjeyw: oh that's relating to how the Sagas update all the objects - so this is fine as it is right now!

app/client/components/user/CreateUser.tsx
enjeyw: what goes wrong exactly?

app/client/sagas/transferUsageSagas.js
enjeyw: is this helpful at all?
https://github.com/Lemoncode/redux-sagas-typescript-by-example/tree/master/01_hello_saga

PR #45, opened 2019-11-06
send_message changes

wangfowen:
- organisation default welcome sms
- refactor send_message into a class
- tests for organisation send sms


app/server/__init__.py
```diff
@@ -18,6 +18,8 @@
 import sys
 import os
 
+from server.utils.phone import MessageProcessor
```
wangfowen: should it still be in utils like this if we're using it within server/init?

does seem cleaner overall to declare it here with the various clients though, so we're not directly calling the clients in the util. if we want to ever extract message processor into a service, should be very easy

enjeyw: > should it still be in utils like this if we're using it within server/init?

I mean we _could_ create a new module for it, but honestly this seems fine to me.

app/server/utils/slack_controller.py
```diff
@@ -179,7 +178,7 @@ def approve_user(message_blocks, username, message_ts, phone):
     )
 
     if phone:
-        send_message(to_phone=phone,
+        message_processor.send_message(to_phone=phone,
                              message='Hooray! Your identity has been successfully verified and Sempo account limits lifted.')
```
enjeyw: I've added a task to get all of these messages into i18n!

PR #47, opened 2019-11-10
Transaction Limits

tristanhcole:
Sempo limits built, need to work out how to do GE limits plus write tests


tristanhcole:
[Add transaction limit to fiat off ramp](https://trello.com/c/qkHUS7qI/8-add-transaction-limit-to-fiat-off-ramp)

app/server/models/credit_transfer.py
```diff
@@ -172,19 +174,67 @@ def resolve_as_completed(self, existing_blockchain_txn=None):
             if self.recipient_user and self.recipient_user.transfer_card:
                 self.recipient_user.transfer_card.update_transfer_card()
 
-        if self.transfer_type == TransferTypeEnum.DEPOSIT and self.fiat_ramp:
+        if self.fiat_ramp and self.transfer_type in [TransferTypeEnum.DEPOSIT, TransferTypeEnum.WITHDRAWAL]:
             self.fiat_ramp.resolve_as_completed()
```
enjeyw: I just realised: the exchange has the order of control the other way round - resolving an exchange as complete resolves the transfers as complete. (both are equivalently valid based on the context AFAIK). What's the process that triggers this in the fiat ramp flow?

app/server/models/credit_transfer.py
```diff
@@ -172,19 +174,67 @@ def resolve_as_completed(self, existing_blockchain_txn=None):
...
+        if self.fiat_ramp and self.transfer_type in [TransferTypeEnum.DEPOSIT, TransferTypeEnum.WITHDRAWAL]:
+            self.fiat_ramp.resolve_as_rejected()
+
         self.resolved_date = datetime.datetime.utcnow()
         self.transfer_status = TransferStatusEnum.REJECTED
 
         if message:
             self.resolution_message = message
 
+    def check_sender_txns_limits(self):
```
enjeyw: can we rename txn/transaction to transfer for consistency? transaction refers to a txn on the blockchaaaaaain

app/server/models/credit_transfer.py
```diff
@@ -172,19 +174,67 @@ def resolve_as_completed(self, existing_blockchain_txn=None):
...
+        if self.sender_user is None:
+            # skip if an exchange
+            return
+
+        relevant_txn_limits = [txn for txn in self.sender_user.get_txn_limits(credit_transfer=self) if str(self.transfer_type) in txn.get('txn_type')]
```
enjeyw: `limit for limit in...`?

app/server/models/user.py
```diff
@@ -573,6 +574,16 @@ def refine_with_default_transfer_usages(self, most_common, nr_of_wanted):
                 break
         return most_common
 
+    def get_txn_limits(self, credit_transfer=None):
+        relevant_limits = []
+        for limit in limits:
```
enjeyw: might be helpful to uppercase `limits` to `LIMITS` to help signify its an imported const...

app/server/models/credit_transfer.py
enjeyw: there's naming ambiguity between `limits` and `limit_rules`, and that's my fault lol

app/server/models/credit_transfer.py
```diff
@@ -172,19 +174,67 @@ def resolve_as_completed(self, existing_blockchain_txn=None):
...
+            if limit.get('total_amount', None) is None:
+                # GE Limits
+                transaction_count = db.session.query(func.count(CreditTransfer.id).label('count'))\
+                    .filter(CreditTransfer.created >= filter_before)\
+                    .filter(CreditTransfer.transfer_type == self.transfer_type) \
+                    .filter(CreditTransfer.sender_user == self.sender_user) \
+                    .execution_options(show_all=True).first().count
+
+                if (transaction_count or 0) > limit.get('txn_count', 0):
+                    message = 'Account Limit "{}" reached. Allowed {} transaction per {} days'.format(limit.get('name', None), limit.get('txn_count', None), limit.get('time_period_days', None))
```
enjeyw: line length crazy out of control!

app/server/models/credit_transfer.py
enjeyw: I'd consider requiring that `txn_type` (or `transfer_type`) be a list such as `applied_to_transfer_types` so we don't get caught be cases such as `'FUND' in 'REFUND'` (returns true).

enjeyw: Also `if str(self.transfer_type) in txn.get('txn_type')` looks more like an application filter than part of the limit rule itself. perhaps this should be refactored into the `get_txn_limits` method? Equally, in hindsight `get_txn_limits` probably makes more sense as a method on the transfer itself, so that we wouldn't have to redefine it for orgs or other transfer acount types in the future

app/server/utils/transaction_limits.py
```diff
@@ -0,0 +1,140 @@
...
+    if t is not None and t.token_type == token.TokenType.LIQUID:
+        return True
+
+    return False
+
+
+def check_group_user_liquid_token_type(user, credit_transfer):
+    t = credit_transfer.token
+    if t is not None and t.token_type == token.TokenType.LIQUID:
+        # TODO: add group account check
```
enjeyw: maybe throw a `NotImplementedError` here for now?

test/unit/app/models/test_models.py
```diff
@@ -249,6 +249,81 @@ def test_new_credit_transfer_rejected(create_credit_transfer):
...
+    GIVEN a CreditTransfer model
+    WHEN a new credit transfer is created
+    THEN check the correct check_sender_txns_limits apply
+    """
+    from server.utils.transaction_limits import limits
+    from server.models.kyc_application import KycApplication
+    from server.models import token
+
+    # Sempo Level 0 limits (payment only)
+    assert create_credit_transfer.check_sender_txns_limits() == [limit['rule'] for limit in limits[:2]]
```
enjeyw: definitely some ambiguity around `limit`/`rule`

app/server/models/credit_transfer.py
```diff
@@ -172,19 +174,67 @@ def resolve_as_completed(self, existing_blockchain_txn=None):
...
+        if self.sender_user is None:
+            # skip if an exchange
+            return
+
+        relevant_txn_limits = [txn for txn in self.sender_user.get_txn_limits(credit_transfer=self) if str(self.transfer_type) in txn.get('txn_type')]
+
+        for limit in relevant_txn_limits:
+            filter_before = datetime.datetime.today() - datetime.timedelta(days=limit.get('time_period_days', 0))
+
+            if limit.get('total_amount', None) is None:
```
enjeyw: might be nicer as `if limit.get('transfer_count') is not None`, that way we can apply it more precisely

app/server/models/credit_transfer.py
```diff
@@ -172,19 +174,67 @@ def resolve_as_completed(self, existing_blockchain_txn=None):
...
+                    .filter(CreditTransfer.transfer_type == self.transfer_type) \
+                    .filter(CreditTransfer.sender_user == self.sender_user) \
+                    .execution_options(show_all=True).first().count
+
+                if (transaction_count or 0) > limit.get('txn_count', 0):
+                    message = 'Account Limit "{}" reached. Allowed {} transaction per {} days'.format(limit.get('name', None), limit.get('txn_count', None), limit.get('time_period_days', None))
+                    self.resolve_as_rejected(message=message)
+                    raise AccountLimitError(message)
+
+                amount_avail = limit.get('txn_balance_percentage', 0)*self.sender_transfer_account.balance
```
enjeyw: could also selectively apply this if `transfer_balance_percentage` is defined.

test/unit/app/models/test_models.py
```diff
@@ -249,6 +249,81 @@ def test_new_credit_transfer_rejected(create_credit_transfer):
...
+
+    # Check Sempo Level 1 limits (payment only)
+    create_credit_transfer.sender_user.is_phone_verified = True
+    assert create_credit_transfer.check_sender_txns_limits() == [limit['rule'] for limit in limits[3:5]]
```
enjeyw: the issue with taking a sub-list is that if we added say a new rule at the front of the list, it breaks all of these in a kinda irritating way.

It'd require a bit of shuffling to fix though (perhaps by converting to a dict with unique keys for each limit), so perhaps we just ignore for now lol

app/server/models/user.py
tristanhcole: Done

app/server/models/credit_transfer.py
tristanhcole: Done

app/server/utils/transaction_limits.py
tristanhcole: good idea

test/unit/app/models/test_models.py
tristanhcole: The other option is to filter by the rule name and relevant transfer type?

i.e. 
`[limit['rule'] for limit in LIMITS[-1:] if rule_name == limit['rule']['name'] ]`

app/server/models/credit_transfer.py
```diff
@@ -172,19 +175,77 @@ def resolve_as_completed(self, existing_blockchain_txn=None):
...
+        for limit in LIMITS:
+            applied = limit.application_filter(self)
+            if applied and str(self.transfer_type) in limit.applied_to_transfer_types:
+                relevant_limits.append(limit)
+
+        return relevant_limits
+
+    def check_sender_transfer_limits(self):
+        if self.sender_user is None:
+            # skip if an exchange
```
wangfowen: is exchange the only case where `sender_user` would be None? is there any other way to check that it's an exchange? and would we not want to limit exchanges potentially?

app/server/models/credit_transfer.py
```diff
@@ -172,19 +175,77 @@ def resolve_as_completed(self, existing_blockchain_txn=None):
...
+                    .execution_options(show_all=True).first().count
+
+                if (transaction_count or 0) > limit.transfer_count:
+                    message = 'Account Limit "{}" reached. Allowed {} transaction per {} days'\
+                        .format(limit.name, limit.transfer_count, limit.time_period_days)
+                    self.resolve_as_rejected(message=message)
+                    raise AccountLimitError(message)
+
+                amount_avail = limit.transfer_balance_percentage*self.sender_transfer_account.balance
+                if self.transfer_amount > amount_avail:
```
wangfowen: should this be moved outside of the `transfer_count` if condition? we could have just a `transfer_amount` condition without a count one

app/server/models/credit_transfer.py
```diff
@@ -172,19 +175,77 @@ def resolve_as_completed(self, existing_blockchain_txn=None):
...
+                    message = 'Account Limit "{}" reached. {} available'.format(limit.name, amount_avail)
+                    self.resolve_as_rejected(message=message)
+                    raise AccountLimitError(message)
+
+            else:
```
wangfowen: instead of an else that distinguishes GE vs sempo, we should keep them generic. if there's a `total_amount` limit, then run this condition. then we could have any combination of different limits

app/server/utils/transaction_limits.py
```diff
@@ -0,0 +1,119 @@
...
+    TransferLimit('Sempo Level 2', 100000, 30, None, None, ['PAYMENT'], check_user_kyc_verified),
+    TransferLimit('Sempo Level 2', 50000, 7, None, None, ['WITHDRAWAL', 'DEPOSIT'], check_user_kyc_verified),
+    TransferLimit('Sempo Level 2', 100000, 30, None, None, ['WITHDRAWAL', 'DEPOSIT'], check_user_kyc_verified),
+
+    TransferLimit('Sempo Level 3', 500000, 7, None, None, ['PAYMENT'], check_user_kyc_two_id_verified),
+    TransferLimit('Sempo Level 3', 1000000, 30, None, None, ['PAYMENT'], check_user_kyc_two_id_verified),
+    TransferLimit('Sempo Level 3', 500000, 7, None, None, ['WITHDRAWAL', 'DEPOSIT'], check_user_kyc_two_id_verified),
+    TransferLimit('Sempo Level 3', 1000000, 30, None, None, ['WITHDRAWAL', 'DEPOSIT'], check_user_kyc_two_id_verified),
+
+    TransferLimit('GE Liquid Token - Standard User', None, 7, 0.1, 1, ['WITHDRAWAL'], check_user_liquid_token_type)
```
wangfowen: so when we create a GE token we label it as liquid? why this distinction for GE token instead of labeling a GE token as GE?

test/unit/app/models/test_models.py
```diff
@@ -249,6 +249,105 @@ def test_new_credit_transfer_rejected(create_credit_transfer):
...
+        db.session.flush()
+
+    # Sempo Level 0 LIMITS (payment only) on check LIMITS
+    with pytest.raises(AccountLimitError):
+        create_credit_transfer.check_sender_transfer_limits()
+
+    # Check GE LIMITS for Liquid Token (withdrawal only) on check LIMITS
+    create_credit_transfer.token.token_type = token.TokenType.LIQUID
+    create_credit_transfer.transfer_type = "WITHDRAWAL"
+    create_credit_transfer.sender_transfer_account.balance = 1000
```
wangfowen: would be good to test the % right on the boundary, and also to test transfer count. what if they did a transfer earlier this week vs two weeks ago

app/server/utils/transaction_limits.py
tristanhcole: @enjeyw

app/server/models/credit_transfer.py
enjeyw: Maybe just change the comment to 'skip if there is no sender, which implies system send'

> would we not want to limit exchanges potentially?
We might, but you'd put the limit  on the inbound of the two credit transfers

>is there any other way to check that it's an exchange
I think there's an exchange TranferType right @tristanhcole ?

app/server/models/credit_transfer.py
enjeyw: +1 i think this is a seperate test that just gets applied sequentially. sorry this is my fault - meant to put it in the  orginal review.

app/server/utils/transaction_limits.py
enjeyw: because that's what it's technically called. i think it makes more sense to name it after the thing it is, rather than the client

app/server/models/credit_transfer.py
tristanhcole: There is a TransferType I could use to check instead, but yeah, some sends (such as system or disbursement) only have a transfer account, no user?

app/server/models/credit_transfer.py
tristanhcole: Ok will ship update today...

app/server/models/credit_transfer.py
tristanhcole: Hm ok. will take a look and ship update

test/unit/app/models/test_models.py
tristanhcole: ok

app/server/models/credit_transfer.py
tristanhcole: In the PoLI payments API you call `deposit. resolve_as_completed()` to resolve the `make_deposit_transfer()` utils fn...

app/server/models/credit_transfer.py
tristanhcole: Comment updated

app/server/models/credit_transfer.py
tristanhcole: ok updated

tristanhcole:
https://github.com/teamsempo/SempoBlockchain/pull/47#pullrequestreview-320652573
Did you see the new pattern?

app/server/models/credit_transfer.py
```diff
@@ -226,7 +225,15 @@ def check_sender_transfer_limits(self):
...
-                amount_avail = limit.transfer_balance_percentage * self.sender_transfer_account.balance
+                allowed_percent_transfer = limit.transfer_balance_percentage * self.sender_transfer_account.balance
+
+                if self.transfer_amount > allowed_percent_transfer:
+                    message = 'Account % Limit "{}" reached. {} available'.format(
+                        limit.name,
+                        max(allowed_percent_transfer, 0)
+                    )
+                    self.resolve_as_rejected(message=message)
+                    raise AccountLimitError(message)
```
tristanhcole: @enjeyw good now?

PR #48, opened 2019-11-10
sops config sharing

enjeyw:
Under the current system, dev config files need to be shared manually via slack or similar, which makes it hard to keep everything up to date and in one place.

This update allows us to securely share config files in an encrypted form over Github, using AWS KMS, (or GPG Keys) to manage access controls.

Once you've been authorised, you can access config settings by running `_SOPS_load.py` inside `.sempoconfig`.
This will extract all keys into the config_files folder. To encrypt them again for sharing on git, run
`_SOPS_save.py`.

Not that SOPS doesn't handle merge conflicts currently - if you try and merge an encrypted file, it will break in a bad way! Instead, if you need to merge in two config files, you need to save the old config, load the new one and merge them by hand.

Hopefully in a future update this will be improved to enable seemless merging.


.sempoconfig/_SOPS_load.py
```diff
@@ -0,0 +1,26 @@
...
+import os
+
+try:
+    os.chdir('./.sempoconfig')
+except FileNotFoundError:
+    pass
+
+os.environ['AWS_PROFILE'] = "sempo"
+
+i = input('You are about to replace your LOCAL config files with those from git - are you sure? (y/n)')
```
tristanhcole: good idea to confirm :)

PR #49, opened 2019-11-11
Dynamic USSD menu transfer reasons and business categories

RubenGeo:


app/server/locale/ussd.en.yml
```diff
@@ -26,15 +26,7 @@ en:
...
-      2. Water
-      3. Energy
-      4. Education
-      5. Health
-      6. General shop
-      7. Environment
-      8. Transport
-      9. Labour
+      %options%
       10. Other
```
RubenGeo: I removed the categories from the tranlations and instead added the dynamic %options% variable. However send_token_reason and directory_listing differed in that send_token_reason has: "10. Other". 

Would a user select this option to other available tranfer reasons or is it used to set a tranfer as the 'other category?

RubenGeo: If other meas like show 'other' option it should remain if it's a separate category it should be removed

RubenGeo: This also affects the other option in the seed script

enjeyw: I think Other is just another option in its own right, rather than a deeper breakdown

app/server/utils/ussd/kenya_ussd_processor.py
```diff
@@ -76,5 +76,19 @@ def replace_vars(menu: UssdMenu, ussd_session: UssdSession, display_text: str, u
...
+                business_usage_string = None
+                if usage.translations is not None and user.preferred_language is not None:
+                    business_usage_string = usage.translations.get(
+                        user.preferred_language)
+                    print(business_usage_string)
+                if business_usage_string is None:
+                    business_usage_string = usage.name
+                message_option = '%d. %s ' % (i+1, business_usage_string)
+                menu_options += message_option
+            replacements.append(['%options%', menu_options])
```
enjeyw: Is there a test that hits this code?
I was trying work out how the USSD menu option number for a custom increments with respect to the hard-coded options, but i couldn't find any way to this code...

app/server/utils/ussd/kenya_ussd_processor.py
```diff
@@ -76,5 +76,19 @@ def replace_vars(menu: UssdMenu, ussd_session: UssdSession, display_text: str, u
...
+            menu_options = ''
+            for i, usage in enumerate(usages):
+                business_usage_string = None
+                if usage.translations is not None and user.preferred_language is not None:
+                    business_usage_string = usage.translations.get(
+                        user.preferred_language)
+                    print(business_usage_string)
+                if business_usage_string is None:
+                    business_usage_string = usage.name
+                message_option = '\n%d. %s' % (i+1, business_usage_string)
```
RubenGeo: How do new lines and tabs get translated to an sms? As a response in postman I just got everything back with spaces and no new lines even also for the other menu's

test/unit/app/utils/test_kenya_ussd_processor.py
```diff
@@ -0,0 +1,45 @@
...
+    ("directory_listing", "sw", '\n1. Chakula\n2. Elimu'),
+    ("directory_listing", None, '\n1. Food\n2. Education'),
+])
+def test_replace_vars(mocker, test_client, init_database, menu_name, language, expected):
+    start_state = partial(UssdSessionFactory)
+    user = standard_user()
+    user.preferred_language = language
+
+    type(user).get_most_relevant_transfer_usage = mocker.PropertyMock(
+        return_value=mock_get_most_relevant_transfer_usage)
```
RubenGeo: I think this is a bit of a hack? Is there another way how I can mock the `get_most_relevant_transfer_usage` function

app/migrations/seed.py
```diff
@@ -265,25 +265,36 @@ def update_or_create_menu(name, description, parent_id=None):
...
-        {'name': 'Water', 'icon': 'message'},
-        {'name': 'Energy', 'icon': 'message'},
-        {'name': 'Education', 'icon': 'message'},
-        {'name': 'Health', 'icon': 'message'},
-        {'name': 'General shop', 'icon': 'message'},
-        {'name': 'Environment', 'icon': 'message'},
-        {'name': 'Transport', 'icon': 'message'},
-        {'name': 'Labour', 'icon': 'message'},
-        {'name': 'Other', 'icon': 'message'},
+        {'name': 'Food', 'icon': 'message', 'translations': {
```
wangfowen: should we do translations this way? or should we have a `translations_key` to do i18n? if nothing passed default to name

app/server/utils/ussd/kenya_ussd_processor.py
wangfowen: ah... i actually haven't tested this. that is good to know. we should test it with the africas talking sandbox. definitely need to figure out how to fix it if it's not actually new lines

app/server/utils/ussd/kenya_ussd_processor.py
wangfowen: +1 to this. also, how does the dynamic options get hooked up to actual usage? for example, if a user responds with 2 for the directory listing, how does the state machine know what "2" corresponds to in terms of transfer usages? i didn't see that bit. an edge case to consider would be if their `get_most_relevant_transfer_usage` changes in between displaying it to customer and them responding

app/migrations/seed.py
RubenGeo: I'm wondering if that does not make is harder for someone using the transfer usage api to add a new transfer usage as they would then also have to edit the yml files?

app/server/utils/ussd/kenya_ussd_processor.py
RubenGeo: If I want to do test it with africas talking sandbox where should I get started?

app/server/utils/ussd/kenya_ussd_processor.py
RubenGeo: Ah thanks for the explanation I did not think that through yet.

app/server/utils/ussd/kenya_ussd_processor.py
enjeyw: sign up here:
https://account.africastalking.com/

go to sandbox, and then ussd, create an endpoint to an ngrok (https://ngrok.com/) server running on ur local machine, and then launch the africa's talking simulator :)

app/server/utils/ussd/kenya_ussd_processor.py
```diff
@@ -76,5 +76,19 @@ def replace_vars(menu: UssdMenu, ussd_session: UssdSession, display_text: str, u
                     replacements.append(['%remaining_attempts%', "You have {} attempts remaining.".format(3 - user.pin_failed_attempts)])
             else:
                 replacements.append(['%remaining_attempts%', ''])
+        elif 'directory_listing' == menu.name or 'send_token_reason' == menu.name:
```
enjeyw: I know this isn't ur doing, but we check for equalities 3 different ways over the course of this function lol!

`menu.name == 'exchange_token_confirmation'`
` elif 'pin_authorization' in menu.name ` (is the 'in' required?)
`'directory_listing' == menu.name `

test/unit/app/utils/test_kenya_ussd_processor.py
enjeyw: I've never used this particular mock interface, so don't take my word for it lol, but I think this is actually a pretty nice usage...

app/server/utils/ussd/kenya_ussd_processor.py
RubenGeo: What should directory listening return? For example if you select option 1. Food where should you go? Whould you get back list of all user that fall into that business category? Or the transfer accounts related to that user?

RubenGeo: Would it happen that the `get_most_relevant_transfer_usage` changes in between? Since this is based on the transfers that a user has sent. So they themselves would have to send transfers while the menu is showing to change that ranking. Or are the multiple people accessing one user account? Or am i missing something else?

enjeyw:
Couple of things a just thought of:
- what does the user see if they’ve never made any transfers (or 1 transfer)? Is it an empty list (or a list of length 1 for the case of 1 transfer) or does it have some concept how to populate the list with defaults?

2. How does a user select a transfer usage category that isn’ in their top 10, be it a less popular existing option or a new one entirely?

app/server/utils/ussd/kenya_ussd_processor.py
wangfowen: > What should directory listening return? For example if you select option 1. Food where should you go?
per comment on slack, don't worry about that for this ticket, if you can just have it translate to the transfer usage, we'll address that in the other ticket
> Or are the multiple people accessing one user account? 
multiple users may share one group account so this is a potential edge case that could happen

app/migrations/seed.py
wangfowen: the nice thing about i18n is the fallbacks, but yeah that could be annoying to use

wangfowen:
@enjeyw we have defaults that are displayed if there's not enough to fill the list. #2 is a great question, since our "other" behavior is now different than what's on GE. for transfer reason, "other" will always be the last option and choosing that takes them to the option to type in a new transfer reason, which would have the same behavior as our admin interface of `find_or_create` transfer usage. for directory listing, we should talk to will. other could just show every other kind not listed in the top 10...?

enjeyw:
did we get a conclusion on number 2?

app/migrations/seed.py
RubenGeo: So should I leave it like this? @enjeyw @wangfowen

app/server/utils/ussd/kenya_ussd_processor.py
RubenGeo: I think the in is needed as there  are multiple menu's with pin_authorization:
image


I p

app/server/utils/ussd/kenya_ussd_processor.py
RubenGeo: Next lines do show up in the africa is talking sandbox:
I think it looks better with \n:
image vs
image
Should I:

1. Remove the \n from my code to make the menu's uniform
2. Or should we add \n to the other menu's (create a trello task for it)?

enjeyw: Definitely agree that \n. looks better. My vote is number 2!

app/migrations/seed.py
enjeyw: I vote leave as is - feels like a much better UX to be able to update transfer usages without having to change YML (we change usages for the oxfam instance pretty regularly).

app/server/utils/ussd/kenya_ussd_processor.py
enjeyw: touché.

enjeyw:
@RubenGeo just pinging you this last one incase the comment slipped through?

app/server/utils/ussd/kenya_ussd_processor.py
wangfowen: see my PR for getting it working with sandbox. it fixes this issue

enjeyw:
Ohh ok - just getting to that pr now lol!

app/server/utils/ussd/kenya_ussd_processor.py
RubenGeo: > @RubenGeo just pinging you this last one incase the comment slipped through?

Which one?

> @enjeyw we have defaults that are displayed if there's not enough to fill the list. #2 is a great question, since our "other" behavior is now different than what's on GE. for transfer reason, "other" will always be the last option and choosing that takes them to the option to type in a new transfer reason (@RubenGeo don't know if you did it this way already, i think i left that out in the specs), which would have the same behavior as our admin interface of `find_or_create` transfer usage. for directory listing, we should talk to will. "other" could just show every other kind not listed in the top 10...?

So should I revert the "other" back to show other categories instead of a seperate option? And then create a menu that show all options that were not in the top 10? Does it not make more sense to have like 10. 'Show other options' and translate that to sw?

RubenGeo: Wait so when a user sends tokens to another user and they can't find their intended transfer usage in orginal list of transfer usage. They're allowed to add a new tranfer usage. 

I thought creating new transer usage was for admin? And the USSD interface is not really handy for adding new transfer usage as you would also need to set translations?

RubenGeo:
> @enjeyw we have defaults that are displayed if there's not enough to fill the list. #2 is a great question, since our "other" behavior is now different than what's on GE. for transfer reason, "other" will always be the last option and choosing that takes them to the option to type in a new transfer reason (@RubenGeo don't know if you did it this way already, i think i left that out in the specs), which would have the same behavior as our admin interface of `find_or_create` transfer usage. for directory listing, we should talk to will. "other" could just show every other kind not listed in the top 10...?

So should I revert the "other" back to show other categories instead of a seperate option? And then create a menu that show all options that were not in the top 10? Does it not make more sense to have like 10. 'Show other options' and translate that to sw?

RubenGeo:
I created `directory_listing_other`menu for now that lists all other tranfser usages. And depending on the discussion I can also add a send_token_reason_other menu. Or a menu that adds another transer usage

wangfowen:
moving this discussion to the trello, since we should get will's input / it's harder to discuss on here

app/migrations/seed.py
wangfowen: and then for translations, we just manually query db to add them in?

enjeyw:
looks like there's a merge error in there @RubenGeo 
`
app_1                   |   File "/src/server/utils/ussd/kenya_ussd_state_machine.py", line 422
app_1                   |     <<<<<<< HEAD
app_1                   |      ^
app_1                   | SyntaxError: invalid syntax

`

RubenGeo:
I think we can finally merge this. Let me know if it still needs changes! @enjeyw

app/server/utils/ussd/kenya_ussd_state_machine.py
```diff
@@ -171,14 +172,30 @@ def save_transaction_amount(self, user_input):
...
 
     def save_transaction_reason(self, user_input):
-        #TODO: use ruben's dynamic code to convert number to reason
-        self.session.set_data('transaction_reason_translated', user_input)
-        self.session.set_data('transaction_reason_id', "1")
+        chosen_transfer_usage = self.get_select_transfer_usage(user_input)
+        self.session.set_data('transaction_reason', chosen_transfer_usage.name)
+
+        if self.user.preferred_language is not None and chosen_transfer_usage.translations is not None:
+            reason_translated = chosen_transfer_usage.translations[self.user.preferred_language]
```
wangfowen: `translations[self.user.preferred_language]` could still be none. should fallback on `chosen_transfer_usage.name` if it is

app/server/utils/ussd/kenya_ussd_state_machine.py
```diff
@@ -171,14 +172,30 @@ def save_transaction_amount(self, user_input):
         self.session.set_data('transaction_amount', user_input)
 
     def save_transaction_reason(self, user_input):
-        #TODO: use ruben's dynamic code to convert number to reason
-        self.session.set_data('transaction_reason_translated', user_input)
-        self.session.set_data('transaction_reason_id', "1")
+        chosen_transfer_usage = self.get_select_transfer_usage(user_input)
+        self.session.set_data('transaction_reason', chosen_transfer_usage.name)
```
wangfowen: we don't use transaction_reason in the data anymore, just translated and id. we can rename translated to name if it's confusing, or transaction_reason_i18n? no strong opinion here

app/server/locale/ussd.en.yml
```diff
@@ -53,17 +59,24 @@ en:
...
-      9. Labour
+      CON Choose Market Category 
+      %{options}
+      9. Show other options
     complete: END Your request has been sent. You will receive an SMS shortly.
+    directory_listing_other:
+      first: |
+        CON Choose Market Category 
+        %{other_options}
+        9. Show other options
```
wangfowen: swahili lists 0. Back as an option, but english doesn't

app/server/locale/ussd.en.yml
```diff
@@ -53,17 +59,24 @@ en:
...
+      middle: |
+        CON Choose Market Category 
+        %{other_options}
+        9. Show other options
+        10. Show previous options
+      last: |
+        CON Choose Market Category 
+        %{other_options}
+        10. Show previous options
```
wangfowen: would this show 1-8 and then skip to 10? what would happen if they pressed 9?

app/server/utils/ussd/kenya_ussd_state_machine.py
```diff
@@ -232,6 +247,14 @@ def process_exchange_token_request(self, user_input):
         amount = float(self.session.get_data('exchange_amount'))
         ussd_tasker.exchange_token(self.user, agent, amount)
 
+    def store_transfer_usage(self, user_input):
+        pass
```
wangfowen: this still gets triggered in the after of the state machine, but it'd do nothing? how come the storing happens at the processor level and not here?

app/migrations/seed.py
```diff
@@ -265,24 +270,34 @@ def update_or_create_menu(name, description, parent_id=None):
...
+        {'name': 'Water', 'icon': 'message',
+            'translations': {'en': 'Water', 'sw': 'Maji'}},
+        {'name': 'Energy', 'icon': 'message', 'translations': {
+            'en': 'Energy', 'sw': 'Kuni/makaa/mafuta/stima'}},
+        {'name': 'Education', 'icon': 'message',
+            'translations': {'en': 'Education', 'sw': 'Elimu'}},
+        {'name': 'Health', 'icon': 'message',
+            'translations': {'en': 'Health', 'sw': 'Afya'}},
+        {'name': 'General shop', 'icon': 'message', 'translations': {
+            'en': 'General shop', 'sw': 'Uchuuzi'}},
```
wangfowen: will changed the translation for General Shop to "Duka la jumla"

wangfowen:
tried running this in sandbox, got an error at:
```
text = KenyaUssdProcessor.custom_display_text(current_menu, ussd_session, user)
  File "/Users/owen/workspace/sempo/SempoBlockchain/app/server/utils/ussd/kenya_ussd_processor.py", line 101, in custom_display_text
    usages = user.get_most_relevant_transfer_usage()
  File "/Users/owen/workspace/sempo/SempoBlockchain/app/server/models/user.py", line 587, in get_most_relevant_transfer_usage
    next(s for s in relevant_transer_usages if s.id == id) for id in most_common_ids]
  File "/Users/owen/workspace/sempo/SempoBlockchain/app/server/models/user.py", line 587, in <listcomp>
    next(s for s in relevant_transer_usages if s.id == id) for id in most_common_ids]
```

app/server/locale/ussd.en.yml
RubenGeo: I dunno tbh. It was like that before I started working on it. But I think it does make sense to add it

app/server/models/user.py
```diff
@@ -570,34 +570,28 @@ def get_most_relevant_transfer_usage(self):
...
+        most_common_ids = [row[0] for row in result]
         # Get most common business_usage_id
-        number_of_wanted_business_id = 9
 
-        most_common_ids = transfer_usage_ids[:number_of_wanted_business_id]
-        # If less than the wanted nr of usage is found fill up the list with default
-        if len(most_common_ids) < number_of_wanted_business_id:
-            most_relevant_ids = self.refine_with_default_transfer_usages(
-                most_common_ids, number_of_wanted_business_id)
+        most_relevant_ids = self.refine_with_default_transfer_usages(
```
wangfowen: this doesn't appear to be used anymore?

app/server/models/user.py
```diff
@@ -570,34 +570,28 @@ def get_most_relevant_transfer_usage(self):
...
+            most_common_ids)
         relevant_transer_usages = TransferUsage.query.filter(
```
wangfowen: is this still necessary? now all TransferUsages are in this list right?

app/server/locale/ussd.en.yml
RubenGeo: Yes. I thought it would make more sense to leave that one out so 9 and 10 have the same meaning in both menus.

If you enter 9 you stay in the same menu. Due to this:
            current_menu_nr = current_menu_nr + 1 if current_menu_nr < max_menu else max_menu
https://github.com/teamsempo/SempoBlockchain/blob/b3ba6d0203898313730f0f94aab2a622d8dc28c6/app/server/utils/ussd/kenya_ussd_state_machine.py#L193

app/server/utils/ussd/kenya_ussd_processor.py
```diff
@@ -97,8 +97,69 @@ def custom_display_text(menu: UssdMenu, ussd_session: UssdSession, user: User) -
             else:
                 return i18n_for(user, "{}.first".format(menu.display_key))
 
-        return None
-
+        if menu.name == 'directory_listing' or menu.name == 'send_token_reason':
+            items_per_menu = 8
 
+            usages = user.get_most_relevant_transfer_usage()
```
wangfowen: i think this makes sense to happen in the after in the state machine still. all setting of data would happen there, here it's just using stuff from the session data to get the display text

wangfowen: mocking of the test would also be easier then, you just pass in data of your liking

app/server/utils/ussd/kenya_ussd_processor.py
```diff
@@ -97,8 +97,69 @@ def custom_display_text(menu: UssdMenu, ussd_session: UssdSession, user: User) -
...
+            KenyaUssdProcessor.store_transfer_usage(ussd_session, usages)
+            ussd_session.set_data('usage_menu', 1)
+            ussd_session.set_data('usage_menu_max', math.floor(len(usages)/items_per_menu))
+            menu_options = KenyaUssdProcessor.create_usages_list(usages[:items_per_menu], user)
+            return i18n_for(
+                user, menu.display_key,
+                options=menu_options
+            )
 
+        if menu.name == 'directory_listing_other' or menu.name == 'send_token_reason_other':
```
wangfowen: if you make the above change to pass in arbitrary data, you can test this better to see that all the different options work as you expect

app/server/locale/ussd.en.yml
wangfowen: agreed on the UX, and sounds like mistakes won't cause issues which is great!

app/server/locale/ussd.en.yml
RubenGeo: I'm adding the back button

app/server/models/user.py
RubenGeo: No, first the most common Transferusages are listed, so the ones that a user has sent a tranfer too, and than the list is supplemented with the default usages. 

We could however extend that list even further with basically the rest of the tranferusages. Then the list would be in this order:
- The transferusages that the user has used before (can either be default or not default)
- The default transferusage
- All other tranfer 

What do you think? 

For now I'm leaving it as is. Unless you guys think the above mentioned change makes sense (or another change)

wangfowen:
You removed the limit on the queries. Aren't we having the user page
through all possible usages now so they'll see all the options ordered by
most used? So default or not are all going to show up in the list

app/server/models/user.py
wangfowen: @RubenGeo You removed the limit on the queries. Aren't we having the user page
through all possible usages now so they'll see all the options ordered by
most used? So default or not are all going to show up in the list

app/server/models/user.py
```diff
@@ -561,7 +562,7 @@ def user_details(self):
         # should drop the country code from phone number?
         return "{} {} {}".format(self.first_name, self.last_name, self.phone)
 
-    def get_most_relevant_transfer_usage(self):
+    def get_most_relevant_transfer_usages(self):
```
wangfowen: now this gets 20 most recent transfers for the user, then creates a map of those usages to the count in this method

app/server/utils/user.py
```diff
@@ -662,3 +663,32 @@ def get_user_by_phone(phone: str, region: str, should_raise=False) -> Optional[U
             raise Exception('User not found.')
         else:
             return None
+
+
+def transfer_usages_for_user(user: User) -> List[TransferUsage]:
```
wangfowen: this is now the method used by the state machine. it outputs *all* transfer usages, sorted by user's most used (based on that map of usage to count), then the defaults, then everything else

test/functional/app/api/test_ussd_api.py
```diff
@@ -116,12 +81,21 @@ def get_session():
 
     resp = req("12.5")
     assert "CON Select Transfer" in resp
+    assert "1. Education" in resp
+    assert "9." in resp
+
+    resp = req("9")
+    assert "CON Please specify" in resp
+    assert "10. Show previous options" in resp
```
wangfowen: also tests dynamic menu / page flip in this end to end test now

app/server/utils/user.py
wangfowen: if we don't output all transfer usages, a user could only ever see the defaults since those are the only ones that show up until a user sends a different kind... but they don't have a way to send a different kind aside from default in that case

app/server/models/user.py
```diff
@@ -570,40 +571,20 @@ def get_most_relevant_transfer_usage(self):
         '''
 
         sql = text('''
-            SELECT *, COUNT(*)  FROM
-                (SELECT u.business_usage_id FROM credit_transfer c
-                LEFT JOIN "user" u ON c.recipient_user_id = u.id
-                WHERE sender_user_id = {} AND c.transfer_status = 'COMPLETE'
+            SELECT *, COUNT(*) FROM
+                (SELECT c.transfer_use::text FROM credit_transfer c
```
wangfowen: we want to group by transfer use instead of business_usage_id since business_usage_id is optional on users, but transfer use will always be set

app/server/models/user.py
```diff
@@ -570,40 +571,20 @@ def get_most_relevant_transfer_usage(self):
...
-            TransferUsage.id).all()
-        for transer_usage_id, in default_transer_usages:
-            if transer_usage_id not in most_common:
-                most_common.append(transer_usage_id)
-            if len(most_common) == nr_of_wanted:
-                break
-        return most_common
+        most_common_uses = {}
+        for row in result:
+            for use in json.loads(row[0]):
```
wangfowen: it's a bit strange that transfer use is a json field that's an array of transfer usage names. harder to look up + perform sql queries on what's the reason for not making a join table of TransferUsage ids to CreditTransfers?

app/server/utils/ussd/kenya_ussd_state_machine.py
```diff
@@ -241,23 +243,23 @@ def process_exchange_token_request(self, user_input):
...
-        items_per_menu = 8
-        usages = self.user.get_most_relevant_transfer_usage()
-        transfer_usage_id_order = []
-        for usage in usages:
-            transfer_usage_id_order.append(usage.id)
-        if self.session.session_data is None:
-            self.session.session_data = {'transfer_usage_mapping': transfer_usage_id_order}
-        elif type(self.session.session_data) is dict:
-            self.session.session_data['transfer_usage_mapping'] = transfer_usage_id_order
-        self.session.set_data('usage_menu', 1)
```
wangfowen: @RubenGeo off by 1 error if you navigated from `send_transfer_reason` to `send_transfer_reason_other` hehe

tristanhcole:
If this is good to go @enjeyw @RubenGeo @wangfowen can it be merged in, please? I believe it fixes a number of USSD bugs?

PR #50, opened 2019-11-12
Docker migration fix

enjeyw:
If there were problems with the database migration files (often a missing merge-heads), sometimes the docker tests would still pass, because:
- The system doesn't explicitly test to see if all migrations worked correctly
- A new migration may not be necessary for the functional/unit tests to pass

This update causes docker tests to explicitly check whether migration builds work, and throws an error if they dont


tristanhcole:
@enjeyw tests aren't passing.

tristanhcole:
[Fix CI/CD migrations error](https://trello.com/c/wHyoSDT6/112-fix-ci-cd-migrations-error)

enjeyw:
@tristanhcole they're failing because of a bug in the data-update process that we didn't know we had, but is now being identified due to this fix... basically the PR is working 😆

PR #53, opened 2019-11-14
hook in business category to admin user creation

wangfowen:
same code that was approved before, but now merging into master since ruben's changes are in. re-tested that everything works as expected

### DIFF
--- a/app/server/utils/user.py
+++ b/app/server/utils/user.py
@@ -7,6 +7,7 @@ from eth_utils import to_checksum_address
 
 from server import db
 from server.models.device_info import DeviceInfo
+from server.models.transfer_usage import TransferUsage
 from server.models.upload import UploadedResource
 from server.models.user import User
 from server.models.custom_attribute_user_storage import CustomAttributeUserStorage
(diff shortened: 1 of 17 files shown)


app/server/utils/user.py
```diff
@@ -389,7 +390,20 @@ def proccess_create_or_modify_user_request(
     transfer_account_name = attribute_dict.get('transfer_account_name')
     first_name = attribute_dict.get('first_name')
     last_name = attribute_dict.get('last_name')
-    business_usage_id = attribute_dict.get('business_usage_id')
+
```
enjeyw: Unless i'm misunderstanding how this works, the way in which an "other" usage gets added feels a bit odd. Like rather than having this deep concept of "other" that threads through the backend that requires the creation of an "other" category that isn't actually ever used, wouldn't it make more sense for the api have only one business usage field, and then if you provide a string that isn't an existing usage it creates a new one. Then on the front end, the concept of other would be constrained just to the UI of the drop down.

wangfowen: yeah agreed, we should not have an other category anymore in itself. changed behavior to reflect this, but haven't tested it since my dev environment is broken

wangfowen: k was able to test that it works now. change is ready for merging

PR #54, opened 2019-11-14
pin auth and related ussd features

wangfowen:
squashed philip's commits, fixed merge conflicts, and made changes per nick's feedback


app/server/models/user.py
```diff
@@ -431,21 +445,32 @@ def decode_single_use_JWS(cls, token, required_type):
...
         return is_valid
 
     def delete_password_reset_tokens(self):
         self.password_reset_tokens = []
 
-    def clear_expired_tokens(self):
+    def delete_pin_reset_tokens(self):
+        self.pin_reset_tokens = []
+
+    def clear_expired_password_reset_tokens(self):
```
enjeyw: feel like we could DRY these functions, to something like:

```
def clear_expired_reset_tokens(self, token_list):

        if is None:
            token_list = []

        valid_tokens = []
        for token in token_list:
            validity_check = self.decode_single_use_JWS(token, 'R')
            if validity_check['success']:
                valid_tokens.append(token)
        token_list = valid_tokens

def clear_expired_pin_reset_tokens(self):
        self.clear_expired_reset_tokens(self.pin_reset_tokens)

def clear_expired_password_reset_tokens(self):
        self.clear_expired_password_reset_tokens(self.pin_reset_tokens)

```

app/server/models/user.py
```diff
@@ -513,14 +550,23 @@ def set_pin(self, supplied_pin=None, is_activated=False):
...
+        is_resetting = len(self.pin_reset_tokens) > 0
+        return is_resetting
+
+    @property
+    def pin(self):
+        return self.pin_hash
+
+    @pin.setter
+    def pin(self, pin):
+        self.pin_hash = self.salt_hash_secret(pin)
```
enjeyw: I think these make more sense to follow the same pattern as the password; a hashed password is set explicitly using `def hash_password(self, password):...` and the hash itself isn't really accessed directly at all,

app/server/utils/ussd/kenya_ussd_state_machine.py
```diff
@@ -114,6 +117,26 @@ def new_pins_match(self, user_input):
             pins_match = True
         return pins_match
 
+    def authorize_pin(self, pin):
+        authorized = False
+        if self.user.failed_pin_attempts is None:
+            self.user.failed_pin_attempts = 0
+        if self.user.failed_pin_attempts == 3:
```
enjeyw: this is probably a ridiculous comment, but I always feel safer when these sorts of checks use `>=` inequality check, so that if somehow the failed attempt count gets to 4, an attacker doesn't then find themsleves with unlimited attempts available.

app/server/utils/ussd/kenya_ussd_state_machine.py
```diff
@@ -125,7 +148,11 @@ def is_token_agent(self, user_input):
 
     # TODO: [Philip] Add community token when available
     def upsell_unregistered_recipient(self, user_input):
-        upsell_message = f"{self.user.first_name} {self.user.last_name} amejaribu kukutumia {self.user} lakini hujasajili. Tuma information yako: jina, nambari ya simu, kitambulisho, eneo, na aina ya biashara yako kwa 0757628885."
+        upsell_message = i18n_for(self.user,
+            'upsell_message',
```
enjeyw: "would you like fries with that?"

app/server/utils/ussd/kenya_ussd_state_machine.py
```diff
@@ -138,21 +165,21 @@ def save_transaction_reason_other(self, user_input):
...
-        pass
+        return self.authorize_pin(user_input)
 
     def is_blocked_pin(self, user_input):
-        pass
+        return self.user.failed_pin_attempts == 3
 
     def is_valid_new_pin(self, user_input):
-        pass
+        return self.is_valid_pin(user_input) and not self.user.check_salt_hashed_password(user_input, self.user.pin)
```
enjeyw: just use `self.user.verify_pin(user_input)` instead and make `check_salt_hashed_password` a private method?

app/server/models/user.py
```diff
@@ -460,6 +485,18 @@ def clear_expired_tokens(self):
                 valid_tokens.append(token)
         self.password_reset_tokens = valid_tokens
 
+    def clear_expired_pin_reset_tokens(self):
```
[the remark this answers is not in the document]
wangfowen: arguments are passed by value in python right? so you'd have to do

```
def clear_expired_pin_reset_tokens(self):
    token_list = self.clear_expired_reset_tokens(self.pin_reset_tokens)
    self.pin_reset_tokens = token_list
```

it doesn't modify the `self.pin_reset_tokens` in the shared method?

wangfowen:
i've made changes per feedback for this PR and the making USSD work one and put them both into this PR now @enjeyw

test/unit/app/utils/test_kenya_state_machine.py
```diff
@@ -2,22 +2,30 @@
...
+from server.utils import user as user_utils
 
 fake = Faker()
 fake.add_provider(phone_number)
 # why do i get dupes if i put it directly on standard_user...?
 phone = partial(fake.msisdn)
 
-standard_user = partial(UserFactory)
+unactivated_user = partial(UserFactory, is_activated=False)
+standard_user = partial(UserFactory, pin_hash=User.salt_hash_secret('0000'), failed_pin_attempts=0)
```
wangfowen: since we don't have `pin=` anymore, this is how we can preset pins in test. does this seem correct @enjeyw?

test/unit/app/utils/test_kenya_state_machine.py
```diff
@@ -154,15 +163,72 @@ def test_change_language(mocker, test_client, init_database, session_factory, us
...
     state_machine = KenyaUssdStateMachine(session, user)
     state_machine.send_sms = mocker.MagicMock()
 
-    #state_machine.feed_char("0000")
-    #assert state_machine.state == "complete"
-    #assert user.custom_attributes.market_enabled == False
-    #state_machine.send_sms.assert_called_with("opt_out_of_market_place_sms")
+    state_machine.feed_char("0000")
+    assert state_machine.state == "complete"
+    market_enabled = list(filter(lambda x: x.name == 'market_enabled', user.custom_attributes.all()))
```
wangfowen: @RubenGeo @enjeyw is this the correct way to get a specific custom_attribute...? seems very nonideal

wangfowen: also the value ends up being str 'false' instead of bool False

app/server/api/ussd_api.py
```diff
@@ -23,6 +25,7 @@ class ProcessKenyaUssd(MethodView):
         Method returns a text response to be displayed to the user's screen
         All responses mid-session begin with CON. All responses that terminate the session begin with the word END
     """
+    @show_all
```
wangfowen: oops i should have left all my new changes as a commit so you can see them. instead i'll just comment here what changed. made this `show_all`, made what's stored on session full `user_input`

app/server/models/user.py
```diff
@@ -422,34 +439,49 @@ def decode_single_use_JWS(cls, token, required_type):
...
-    def clear_expired_tokens(self):
-
-        # For some reason the existing user get an None instead of a []
-        # during migration. This is to ensure no TypeError occurs
+    def delete_pin_reset_tokens(self):
+        self.pin_reset_tokens = []
 
-        if self.password_reset_tokens is None:
-            self.password_reset_tokens = []
+    def clear_expired_reset_tokens(self, token_list):
```
wangfowen: DRYed this method, removed pin setter and getter

test/unit/app/utils/test_kenya_state_machine.py
```diff
@@ -154,15 +163,72 @@ def test_change_language(mocker, test_client, init_database, session_factory, us
...
+    assert market_enabled[0].value == 'false'
+    state_machine.send_sms.assert_called_with("opt_out_of_market_place_sms")
+
+
+# TODO: when add the other things requiring pin auth, also test the failure states here
+@pytest.mark.parametrize("session_factory, user_factory, user_input, expected, before_failed_pin_attempts, after_failed_pin_attempts",
```
wangfowen: finished opt_out_of_marketplace test, removed redundant authorize_pin test, removed separate helpers for various numbers of failed_pin_attempts, instead just instantiated them in the tests here

app/server/utils/ussd/kenya_ussd_processor.py
```diff
@@ -67,14 +65,19 @@ def replace_vars(menu: UssdMenu, ussd_session: UssdSession, display_text: str, u
             replacements.append(['%token_name%', token.name])
             replacements.append(['%exchange_amount%', ussd_session.session_data['exchange_amount']])
         elif 'pin_authorization' in menu.name or 'current_pin' in menu.name:
-            if user.pin_failed_attempts() > 0:
+            if user.failed_pin_attempts is not None and user.failed_pin_attempts > 0:
```
wangfowen: per nick's comment, tested for None case but explicitly instead of the or way

app/server/models/user.py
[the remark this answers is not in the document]
enjeyw: oh yess good catch

test/unit/app/utils/test_kenya_state_machine.py
enjeyw: yeah totally legit methinks.

test/unit/app/utils/test_kenya_state_machine.py
enjeyw: `user.custom_attributes` isn't actually a list of attributes, but rather an SQLAlchemy query, a better way of getting custom attributes is probably:
`user.custom_attributes.filter_by(name='market_enabled').all()`

enjeyw: ^ which is super efficient because it's converted into a raw postgres query.

enjeyw: just tried setting the value type for the custom attributes to JSON on my local machine and fixes the typing issue nicely:
`

from server import db
from sqlalchemy.dialects.postgresql import JSON
from server.models.utils import ModelBase
from server.models.user import User


class CustomAttributeUserStorage(ModelBase):
    __tablename__ = 'custom_attribute_user_storage'

    name = db.Column(db.String)
    value = db.Column(JSON)

    user_id = db.Column(db.Integer, db.ForeignKey('user.id'))
    uploaded_image_id = db.Column(db.Integer)
`

enjeyw:
Looks good to merge @wangfowen, though did comment on possible updates to that test and solving the 'false' string issue

PR #55, opened 2019-11-16
get ussd pin working with sandbox

wangfowen:


app/server/api/ussd_api.py
```diff
@@ -31,19 +32,24 @@ def post(self):
         user_input = post_data.get('text')
         service_code = post_data.get('serviceCode')
 
+
         if phone_number:
             msisdn = proccess_phone_number(phone_number, 'KE')
-            user = User.query.filter_by(phone=msisdn).first()
+            user = User.query.execution_options(show_all=True).filter_by(phone=msisdn).first()
```
wangfowen: need this since the API doesn't have an org...? user wasn't being found. is this the correct way to do this or how to do?

app/server/api/ussd_api.py
```diff
@@ -31,19 +32,24 @@ def post(self):
...
             else:
-                current_menu = KenyaUssdProcessor.process_request(session_id, user_input, user)
-                ussd_session = create_or_update_session(session_id, user, current_menu, user_input, service_code)
+                current_menu = KenyaUssdProcessor.process_request(session_id, latest_input, user)
+                ussd_session = create_or_update_session(session_id, user, current_menu, latest_input, service_code)
                 text = menu_display_text_in_lang(current_menu, user)
                 #TODO(ussd): change this out for i18n placeholders
                 text = KenyaUssdProcessor.replace_vars(current_menu, ussd_session, text, user)
+
+                db.session.commit()
```
wangfowen: i removed all commits and just do one at the very end instead. this seems a bit scary to not have commits anywhere else. but otherwise we get that user is deleted error.... thoughts?

app/server/locale/ussd.en.yml
```diff
@@ -1,30 +1,30 @@
 en:
   kenya:
-    initial_language_selection: >
+    initial_language_selection: |
```
wangfowen: this preserves the newlines

app/server/utils/ussd/kenya_ussd_state_machine.py
```diff
@@ -112,7 +112,7 @@ def is_valid_pin(self, user_input):
     def new_pins_match(self, user_input):
         pins_match = False
         # get previous pin input
-        initial_pin = self.session.user_input.split('*')[-2]
```
wangfowen: user_input to session lags one behind, so it was actually "1234" in the input, not "1234*1234". but saving it to session_data seems safer than depending on the previous input

app/server/api/ussd_api.py
```diff
@@ -31,19 +32,24 @@ def post(self):
...
+            latest_input = user_input.split('*')[-1]
             # TODO(ussd): 'exit_not_registered' if no user
             if None in [user, msisdn, session_id]:
                 current_menu = UssdMenu.find_by_name('exit_invalid_request')
                 text = menu_display_text_in_lang(current_menu, user)
             else:
-                current_menu = KenyaUssdProcessor.process_request(session_id, user_input, user)
-                ussd_session = create_or_update_session(session_id, user, current_menu, user_input, service_code)
+                current_menu = KenyaUssdProcessor.process_request(session_id, latest_input, user)
+                ussd_session = create_or_update_session(session_id, user, current_menu, latest_input, service_code)
```
wangfowen: or should i save the entire chain of `user_input` to the session instead of just the latest?

test/unit/app/utils/test_kenya_state_machine.py
```diff
@@ -56,6 +59,7 @@
      (initial_pin_entry_state, standard_user, "0000", "initial_pin_confirmation"),
      (initial_pin_entry_state, standard_user, "AAAA", "exit_invalid_pin"),
      # initial_pin_confirmation transitions tests
+     #TODO: remove these tests and the other testing of pin, they're redundant
```
PhilipWafula: These tests don't test the pin alone, they also test that a specific transition works, it just happens to use the pin feature. In a sense they are still unit-tests  in that they test that that transition will behave in a specific way provided certain values. What do you think ? I think we should retain the so that every transition is tested, so that should the transition break for a reason other than the pin we can still catch trace the breakage.

app/server/utils/ussd/kenya_ussd_state_machine.py
```diff
@@ -100,9 +98,11 @@ def change_opted_in_market_status(self, user_input):
             }
         }
         set_custom_attributes(attrs, self.user)
-        db.session.commit()
         self.send_sms("opt_out_of_market_place_sms")
 
+    def set_pin_data(self, user_input):
```
PhilipWafula: This is a really cool way to do it

test/unit/app/utils/test_kenya_state_machine.py
```diff
@@ -183,6 +186,7 @@ def test_opt_out_of_marketplace(mocker, test_client, init_database):
     state_machine.send_sms.assert_called_with("opt_out_of_market_place_sms")
 
 
+#TODO: also test that users pin becomes the correct thing after
```
PhilipWafula: Does this TODO refer to a test for the pin changing post `complete_initial_pin_change` and `complete_pin_change` ?

test/unit/app/utils/test_kenya_state_machine.py
```diff
@@ -199,6 +203,7 @@ def test_authorize_pin(test_client, init_database, session_factory, user_input,
     assert user.failed_pin_attempts == failed_pin_attempts
 
 
+#TODO: also test that users pin does not become the new thing after
```
PhilipWafula: In the event that this TODO and the one above https://github.com/teamsempo/SempoBlockchain/pull/55/files#r347084848 refer to those tests, then I think the appropriate place for those tests to be is in the utils.user file where those methods are. So I'll put those tests there.

test/unit/app/utils/test_kenya_state_machine.py
```diff
@@ -220,3 +215,39 @@ def test_exit_pin_blocked(test_client, init_database, session_factory, user_fact
     state_machine.feed_char(user_input)
     assert state_machine.state == expected
     assert user.failed_pin_attempts == failed_pin_attempts
+
+
+@pytest.mark.parametrize("session_factory, user_factory, user_input, expected_new_pin, expected_activation_status",
+                         [(initial_pin_confirmation_state, unactivated_user, '0000', '0000', True)])
```
wangfowen: if there's only one, we don't need to parameterize

test/unit/app/utils/test_user.py
```diff
@@ -53,3 +59,40 @@ def test_save_device_info(save_device_info, new_sempo_admin_user, serial_number,
...
+                            expected_activation_status):
+    ua_user = user_factory()
+    ua_user.phone = phone()
+    assert ua_user.pin is None
+    assert ua_user.is_activated is False
+
+    user.send_sms = mocker.MagicMock()
+    change_initial_pin(ua_user, user_input)
+
+    assert ua_user.verify_pin(expected_new_pin) is True
```
wangfowen: we don't need this test if the state machine test does essentially the same thing

app/server/api/ussd_api.py
enjeyw: this is a totally fine way of doing it, though you can also use the `show_all` decorator if you're absolutely confident that the information being queried only relates to the user. We use this on the me api:
```
from server.utils.auth import requires_auth, show_all


class MeAPI(MethodView):
    @requires_auth
    @show_all
    def get(self):

```

app/server/api/ussd_api.py
enjeyw: this seems fine to me lol. Why u find scary?

enjeyw: also, a while back I implemented the auto commit that happens at the end of a request if it's a 200 response, so technically u don't need a commit here at all, but I'm kinda feelling like i now dislike the auto-commit for being far too magic..

tristanhcole: I disagree. I think it makes perfect sense to auto-commit on successful responses.

enjeyw: the reason i dislike it is that it's not uncommon for us to want to commit on a failed response as well to log the failure reason in the db, and this needs to be done manually in the api function  itself. This means the way we do commits becomes inconsistent between successes and failures, which feels bad compared to "commit your session before you return a response if you want keep the data, and don't if you don't".

app/server/utils/ussd/kenya_ussd_state_machine.py
enjeyw: yeah 100 agreed.

app/server/api/ussd_api.py
enjeyw: what's the downside to saving the entire chain, but just using the latest? Added complexity without a real  justification

test/unit/app/utils/test_kenya_state_machine.py
```diff
@@ -167,26 +164,23 @@ def test_change_language(mocker, test_client, init_database, session_factory, us
...
 
 
-# TODO: need pin authorisation implemented to do this
 def test_opt_out_of_marketplace(mocker, test_client, init_database):
     session = opt_out_of_market_place_pin_authorization_state()
     user = standard_user()
     user.phone = phone()
-    #TODO: how fetch from custom attributes?
-    #assert user.custom_attributes['market_enabled'] == None
+    # TODO: how fetch from custom attributes?
```
RubenGeo: This should work: `user.custom_attributes.all()`

app/server/locale/ussd.en.yml
RubenGeo: I was just wondering. When I used the africa is talking sandbox it really looked like it simulated a smartphone. Would it also work on a feature phone?

enjeyw:
suggest change line 68 of kenya_ussd_processor.py to
```
if user.failed_pin_attempts or 0 > 0:
```

The none value was breaking for me

app/server/api/ussd_api.py
enjeyw: Having now played around with it, I definitely think it'd be good to save the chain to help with diagnostics

enjeyw:
I can imagine the session table ending up at a gazillion lines pretty quickly, but i guess that's something we can worry about later lol!

app/server/locale/ussd.en.yml
wangfowen: yep, feature phone interface should be the same

app/server/api/ussd_api.py
wangfowen: good idea, will do

PR #57, opened 2019-11-21
Setup directory listing async task.

PhilipWafula:
Added worker class to process the sending of directory listings.

Added directory listing method to return user phone number and bio.

Added directory listing transition.

Refactor how directory listing async task is called.

Added test setup for directory listing transition.

Added directory listing celery task.

Refactored directory listing to take phone number.
Did this as opposed to passing user object to the Directory Listing worker as an argument to circumvent serialization issues that arise when a user object is passed.


.sempoconfig/common_config.ini
```diff
@@ -1,78 +1,82 @@
 [APP]
```
wangfowen: @enjeyw are these supposed to be committed? i don't see a gitignore for them

app/server/locale/directory_listing.en.yml
```diff
@@ -0,0 +1,12 @@
+en:
+  business_categories: ['Food',
```
wangfowen: these don't need to be in a separate yml file, they can be moved inside of the ussd.yml

app/server/models/user.py
```diff
@@ -146,6 +146,10 @@ def phone(self):
     def phone(self, phone):
         self._phone = proccess_phone_number(phone)
 
+    # TODO [Philip]: Revisit when user bio attribute is implemented
+    def directory_listing(self):
+        return f"{self.phone} {self.bio}"
```
wangfowen: bio is implemented. it's a custom_attribute

app/server/utils/ussd/kenya_ussd_state_machine.py
```diff
@@ -155,6 +155,17 @@ def process_send_token_request(self):
     def fetch_user_exchange_rate(self):
         pass
 
+    def send_directory_listing(self, user_input):
+        try:
+            send_directory_listing_task = celery_app.signature('worker.celery_task.send_directory_listing',
```
wangfowen: @enjeyw is this the kind of thing where we'd throw it in misc tasker?

@PhilipWafula see `utils/misc_tasks.py`

worker/celery_tasks.py
```diff
@@ -87,6 +88,12 @@ def log_error(request, exc, traceback, credit_transfer_id):
         blockchain_processor.send_blockchain_result_to_app(result)
 
 
+@celery_app.task()
+def send_directory_listing(recipient_phone, user_input):
```
wangfowen: @enjeyw is this one giant file that has all the celery tasks ever? why not put these directly in blockchain_tasker and misc_tasker?

app/server/models/user.py
PhilipWafula: Will add this in a few, currently working on the react bits finally got the hang of it, I think

app/server/utils/ussd/kenya_ussd_state_machine.py
wangfowen: or it might make sense to make a `ussd_tasks.py` since we'll have other ones coming up

.sempoconfig/common_config.ini
PhilipWafula: Came in with Nick's seed data merge to master

app/server/utils/ussd/kenya_ussd_state_machine.py
PhilipWafula: Yeah makes sense. Should this be its own task or can I just lump it in here?

wangfowen: just leave it here for now. when i work on send_tokens i'll refactor it

.sempoconfig/common_config.ini
wangfowen: egh, this is why i like rebase better :/ having a hard time figuring out where philip's changes are in the commit tree to see how this happened, whereas with rebase, they'd just be at the top

app/server/utils/user.py
```diff
@@ -224,6 +224,9 @@ def create_transfer_account_user(first_name=None, last_name=None, preferred_lang
         if current_app.config['AUTO_APPROVE_TRANSFER_ACCOUNTS'] and not is_self_sign_up:
             transfer_account.approve_and_disburse()
 
+        db.session.add(transfer_account)
```
wangfowen: had to add this or else `transfer_account.id` is `None` and so `default_transfer_account_id` isn't set correctly... but we're not supposed to commit early...? @enjeyw

enjeyw: this _should_ work:
```

db.session.add(transfer_account)
user.default_transfer_account = transfer_account

```

app/server/utils/ussd/directory_listing.py
```diff
@@ -0,0 +1,86 @@
...
+            - users matching recipient provided business category
+            - users with the recipient's token
+            - users who are opted in to market [custom_attribute 'market_enabled' has value true]
+            - order query by transaction count and limit to top 5 transacting users
+        """
+        users = User.query.join(CustomAttributeUserStorage).join(TransferAccount).filter(
+            User.business_usage_id == self.selected_business_category.id,
+            TransferAccount.token_id == token_id,
+            CustomAttributeUserStorage.name == 'market_enabled',
+            CustomAttributeUserStorage.value == 'true').order_by(
```
enjeyw: we defo should update CustomAttributeUserStorage.value to a JSON to avoid str comparisons

app/server/utils/user.py
enjeyw: this should work:
```

db.session.add(transfer_account)
user.default_transfer_account = transfer_account

```

wangfowen: doesn't seem to work, still getting None on line `transfer_account = TransferAccount.query.get(user.default_transfer_account_id)` in `test_directory_listing`

PR #58, opened 2019-11-21
Set os path

RubenGeo:


app/migrations/clear_seed_dev.py
```diff
@@ -1,9 +1,16 @@
 import subprocess
 from sqlalchemy import create_engine
+import sys
+import os
+
+sys.path.append(os.path.abspath(os.path.join(os.getcwd(), "..", "..")))
+sys.path.append(os.path.abspath(os.path.join(os.getcwd(), "..")))
+sys.path.append(os.path.abspath(os.path.join(os.getcwd(), "..")))
```
wangfowen: this one's repeated? also, is it bad to have these path additions that are relative to current directory? for i18n, doing that broke for me on docker. there i used a path relative to the file itself instead:

```
dirname = os.path.dirname(__file__)
i18n.load_path.append(os.path.abspath(os.path.join(dirname, 'locale')))
```

enjeyw: agree that the suggestion by @wangfowen seems to make more sense

PR #60, opened 2019-11-24
state machine infra for everything remaining

wangfowen:
this is a big one, encompasses a lot of the remaining USSD stuff

- fixed `replace_vars` to use i18n var replacement
- split up state machine tests
- state machine code and tests for remaining states
- starting functionality for actual processes

TODOs:
- tests for new replace_vars behavior
- finish processes functionality
- tests for the processes


app/server/constants.py
```diff
@@ -9,11 +9,13 @@
 RANKED_ADMIN_TIERS = ['sempoadmin', 'superadmin', 'admin', 'subadmin', 'view']
 RANKED_VENDOR_TIERS = ['supervendor', 'vendor', 'cashier']
 RANKED_BENEFICIARY_TIERS = ['beneficiary']
+POSSIBLE_TOKEN_AGENTS = ['grassroots_token_agent']
```
wangfowen: we can sunset this if we do end up launching with fiat ramp

app/server/utils/user.py
```diff
@@ -624,3 +625,19 @@ def change_initial_pin(user: User, new_pin):
 
 def change_current_pin(user: User, new_pin):
     change_pin(user, new_pin)
+
+
+def default_token(user: User) -> Token:
```
wangfowen: i didn't follow the discussion on default org. do all users now have a default organisation? what's that organisation's token? trying to see if this method would always return something correct and/or if i can just go with one or the other of transfer account vs organisation

test/unit/app/utils/ussd/kenya/test_exchange.py
```diff
@@ -0,0 +1,58 @@
+import pytest
```
wangfowen: i've split out the state machine tests into different groupings, the one file was getting too ridiculously massive and they're not all that related. on the upside, it's more obvious now if we're missing any tests related to a particular grouping of states. on the downside, pin overlaps a bit with the other groupings, we could end up in a place where we make a mistake thinking a test is in one file and not doing it in another...? not that it's that much easier to spot it in one massive file

app/server/utils/user.py
enjeyw: I think u comment this before we chatted lol, but users won't necessarily have a default org, but there will necessarily be a system level organisation that can be found via 
`Organisation.query.filter_by(is_master=true)...`

app/server/utils/ussd/kenya_ussd_state_machine.py
```diff
@@ -139,33 +145,50 @@ def complete_initial_pin_change(self, user_input):
     def complete_pin_change(self, user_input):
         change_current_pin(user=self.user, new_pin=user_input)
 
-    def recipient_exists(self, phone_number):
-        pass
+    # recipient exists, is not initiator, matches active and agent requirements
+    def is_expected_recipient(self, user, should_be_active, should_be_agent):
```
enjeyw: very minor but the original `is_valid_recipient` feels like a better naming to me - took me a couple of seconds to grok that we weren't _expecting_ a particular recipient per se.

test/unit/app/utils/ussd/kenya/test_exchange.py
enjeyw: I think this is much better! saves scrolling for a billion years trying to find things.

enjeyw:
@wangfowen type declaration causes an import error :(

Tried swapping over to the other import form and it still fails :((((((

In the end the only thing i could get to work was removing the type declaration. bloody python
```
  File "/Users/Nick/PycharmProjects/SempoBlockchain/app/server/utils/transaction_limits.py", line 4, in <module>
    from server.models.credit_transfer import CreditTransfer
ImportError: cannot import name 'CreditTransfer'
```

enjeyw:
> @wangfowen type declaration causes an import error :(
> 
> Tried swapping over to the other import form and it still fails :((((((
> 
> In the end the only thing i could get to work was removing the type declaration. bloody python
> 
> ```
>   File "/Users/Nick/PycharmProjects/SempoBlockchain/app/server/utils/transaction_limits.py", line 4, in <module>
>     from server.models.credit_transfer import CreditTransfer
> ImportError: cannot import name 'CreditTransfer'
> ```

oh that's actually what is causing CI to fail.

enjeyw:
Not from this PR, but 
```
    def has_valid_pin(self):
        # not in the process of resetting pin and has a pin
        self.clear_expired_pin_reset_tokens()
        not_resetting = len(self.pin_reset_tokens) == 0

        return self.pin is not None and not_resetting
```

in models/user.py is invalid. 

Should be:

```
    def has_valid_pin(self):
        # not in the process of resetting pin and has a pin
        self.clear_expired_pin_reset_tokens()
        not_resetting = len(self.pin_reset_tokens) == 0

        return self.pin_hash is not None and not_resetting
```

app/server/utils/transaction_limits.py
```diff
@@ -11,11 +12,11 @@ def check_user_liquid_token_type(credit_transfer):
...
 
 
-def check_group_user_liquid_token_type(credit_transfer):
+def check_group_user_liquid_token_type(credit_transfer: CreditTransfer):
     t = credit_transfer.token
-    if t is not None and t.token_type == token.TokenType.LIQUID:
-        # TODO: add group account check
-        raise NotImplementedError
+    if t is not None and t.token_type == token.TokenType.LIQUID and \
+            credit_transfer.sender_user.has_group_account_role():
```
enjeyw: should be 
`credit_transfer.sender_user.has_group_account_role` as `has_group_account_role` is a hybrid property rather than a pure method

app/server/utils/ussd/kenya_ussd_processor.py
```diff
@@ -42,41 +44,60 @@ def next_state(session: UssdSession, user_input: str, user: User) -> UssdMenu:
...
+                # TODO: replace this with a no bio message?
+                return i18n_for(user, menu.display_key, user_bio=bio)
+            else:
+                return i18n_for(user, menu.display_key, user_bio=bio)
+
+        if menu.name == 'send_token_confirmation':
+            recipient = get_user_by_phone(ussd_session.get_data('recipient_phone'), 'KE', True)
+            recipient_phone = recipient.user_details()
+            token = default_token(user)
+            transaction_amount = ussd_session.get_data('transaction_amount')
```
enjeyw: I'm getting:

```
  File "/Users/Nick/PycharmProjects/SempoBlockchain/app/server/utils/ussd/kenya_ussd_processor.py", line 61, in custom_display_text
    transaction_amount = ussd_session.get_data('transaction_amount')
  File "/Users/Nick/PycharmProjects/SempoBlockchain/app/server/models/ussd.py", line 54, in get_data
    return self.session_data[key]
KeyError: 'transaction_amount'

```

app/server/utils/ussd/kenya_ussd_state_machine.py
wangfowen: i changed it because it's not necessarily a valid recipient. in the case of upsell, we're looking to match specifically on an `is_disabled` user. but i'll change the name back, expected is confusing you're right

app/server/utils/ussd/kenya_ussd_processor.py
wangfowen: is this when you run it for real? that is still completely untested, will get to it

enjeyw: ohh yeah it was me trying to dry run with ussd lol

app/server/utils/ussd/token_processor.py
```diff
@@ -19,16 +23,16 @@ def send_sms(user, message_key, **kwargs):
...
+        TokenProcessor.send_sms(user, message_key, amount=amount, token_name=default_token(user),
                                 other_user=other_user.user_details(), date=tx_time.strftime('%d/%m/%Y'), reason=reason,
                                 time=tx_time.strftime('%I:%M %p'), balance=balance)
 
     @staticmethod
     def exchange_success_sms(message_key: str, user: User, other_user: User, own_amount: float, other_amount: float,
                              tx_time: datetime, balance: float):
-        # TODO: convert amounts?
-        TokenProcessor.send_sms(message_key, own_amount=own_amount, other_amount=other_amount,
+        # TODO: convert amounts - truncate? need to turn from cents to dollar?
```
enjeyw: Yeah - might be a 'not now' thing, but different currencies truncate in different ways (some don't show decimal places, some do). This used to be a setting a system level, but probably makes more sense at a token level now.

app/server/utils/ussd/token_processor.py
```diff
@@ -56,44 +75,73 @@ def transfer_token(sender: User, recipient: User, amount: float, reason_id: Opti
...
     @staticmethod
     def send_balance_sms(user: User):
-        balance = TokenProcessor.get_balance(user)
-        # TODO: send sms
+        def get_token_info(transfer_account: TransferAccount):
+            token = transfer_account.token
+            limit = TokenProcessor.get_limit(user, token)
+            return {
+                # TODO: is it name or symbol?
+                "name": token.name,
```
enjeyw: Symbol probably makes more sense... maybbe??

app/server/utils/ussd/token_processor.py
```diff
@@ -56,44 +75,73 @@ def transfer_token(sender: User, recipient: User, amount: float, reason_id: Opti
...
+                "balance": transfer_account.balance,
+                "limit": limit.transfer_balance_percentage,
+                "exchange_rate": TokenProcessor.get_exchange_rate(user, token)
+            }
+        # TODO: should we merge transfer accounts with same tokens..? any currencies we should filter out?
+        token_info = map(get_token_info, user.transfer_accounts)
+        token_balances = "\n".join(map(lambda x: f"{x['name']} {x['balance']}", token_info))
+        # this one is also not well generalized currently - how get real world currency name of reserve token?
```
enjeyw: on the system, we won't call the reserve token `DAI` or whatever - we'll just call it eg USD, so can be called by the symbol just as the other ones are

app/server/utils/ussd/token_processor.py
```diff
@@ -56,44 +75,73 @@ def transfer_token(sender: User, recipient: User, amount: float, reason_id: Opti
...
+            token = transfer_account.token
+            limit = TokenProcessor.get_limit(user, token)
+            return {
+                # TODO: is it name or symbol?
+                "name": token.name,
+                "balance": transfer_account.balance,
+                "limit": limit.transfer_balance_percentage,
+                "exchange_rate": TokenProcessor.get_exchange_rate(user, token)
+            }
+        # TODO: should we merge transfer accounts with same tokens..? any currencies we should filter out?
```
enjeyw: I wanna say no to the first one, but then we'd need a way of distinguishing them to the user... we could probably pull the transfer account's associated organisation name...

no filtering as far as i know

app/server/utils/ussd/token_processor.py
```diff
@@ -0,0 +1,147 @@
...
+            TokenProcessor.send_success_sms("send_token_sender_sms", amount, sender, recipient, reason_str, tx_time,
+                                            sender_balance)
+            TokenProcessor.send_success_sms("send_token_recipient_sms", amount, recipient, sender, reason_str, tx_time,
+                                            recipient_balance)
+        except Exception:
+            TokenProcessor.send_sms(sender, "send_token_error_sms", amount=amount,
+                                    token_name=default_token(sender).name, recipient=recipient.user_details())
+
+    @staticmethod
+    def exchange_token(sender: User, agent: User, amount: float):
```
enjeyw: i'm a wee bit confused by this method @wangfowen - it creates a credit transfer rather than an exchange and then doesn't seem to do anything with it?

PR #61, opened 2019-11-29
tests for token processor + fixes to code

wangfowen:
- tests for token processor
- fixes to code, all functionality confirmed working
- added is_ghost to TransferAccount to make send_balance_sms work correctly - if we create transfer accounts as temporary measures to exchange currencies to a user's default currency, those transfer accounts don't show up in the sms. as soon as a non-exchange transaction happens on that transfer account, then it is no longer a ghost account


app/server/utils/ussd/token_processor.py
```diff
@@ -43,8 +43,9 @@ def get_balance(user: User):
 
     @staticmethod
     def get_limit(user: User, token: Token):
-        example_transfer = CreditTransfer(transfer_type='EXCHANGE', sender_user=user, token=token, amount=0)
+        example_transfer = CreditTransfer(transfer_type='EXCHANGE', sender_user=user, recipient_user=user, token=token, amount=0)
         limits = example_transfer.get_transfer_limits()
+        db.session.delete(example_transfer)
```
wangfowen: this feels super jank... we create the CreditTransfer to get the limits. if we don't delete it though, it actually inits the CreditTransfer in the db for whatever reason, and then the next time we run this again, the transfer limit check fails because we've already had 1 transaction

app/server/models/ussd.py
```diff
@@ -49,6 +50,10 @@ def set_data(self, key, value):
             self.session_data = {}
         self.session_data[key] = value
 
+        # https://stackoverflow.com/questions/42559434/updates-to-json-field-dont-persist-to-db
```
wangfowen: this took a while to figure out and seems really easy to make the same mistake elsewhere that uses json

app/server/utils/ussd/token_processor.py
```diff
@@ -23,28 +23,31 @@ def send_sms(user, message_key, **kwargs):
     @staticmethod
     def send_success_sms(message_key: str, amount: float, user: User, other_user: User, reason: str, tx_time: datetime,
                          balance: float):
-        # TODO: convert amounts - truncate? need to turn from cents to dollar?
-        TokenProcessor.send_sms(user, message_key, amount=amount, token_name=default_token(user),
+        rounded_balance = str(round(balance, 2))
```
wangfowen: decided to round everything, otherwise it looks terrible with a bunch of decimal places

app/server/utils/ussd/token_processor.py
```diff
@@ -61,75 +64,92 @@ def transfer_token(sender: User, recipient: User, amount: float, reason_id: Opti
...
-        token_balances = "\n".join(map(lambda x: f"{x['name']} {x['balance']}", token_info))
-        # this one is also not well generalized currently - how get real world currency name of reserve token?
-        token_exchanges = "\n".join(
-            map(lambda x: f"{x['limit'] * x['balance']} {x['name']} (1 {x['name']} = {x['exchange_rate']} KSH)",
-                token_info))
+
+        reserve_token = user.get_reserve_token()
+        # transfer accounts could be created for other currencies exchanged with, but we don't want to list those
+        # could there be multiple reserve? should we filter out all reserve tokens?
+        transfer_accounts = filter(lambda x: x.is_ghost is not True and x.token_id != reserve_token.id,
```
wangfowen: currently if a user has any non-liquid tokens aside from the master reserve, `get_limit` will error out. might be a bit aggressive to error out, but we need to ensure we don't try to do the limit stuff on non-grassroots tokens

test/functional/app/api/test_ussd_api.py
```diff
@@ -0,0 +1,137 @@
...
+from server.models.ussd import UssdSession, UssdMenu
+from server.utils.user import default_transfer_account
+
+fake = Faker()
+fake.add_provider(phone_number)
+phone = partial(fake.msisdn)
+
+
+@pytest.fixture(scope='module')
+def init_menus(test_client, init_database):
```
wangfowen: is it possible to run seed.py in test env? then we can drop this

test/unit/app/utils/ussd/test_token_processor.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+def test_exchange_token(mocker, test_client, init_database, initialised_blockchain_network):
+    sender = UserFactory(preferred_language="en", phone=phone(), first_name="Bob", last_name="Foo")
+    token1 = Token.query.filter_by(symbol="SM1").first()
+    create_transfer_account_for_user(sender, token1, 200)
+
+    agent = UserFactory(phone=phone(), first_name="Joe", last_name="Bar")
+    agent.set_held_role('TOKEN_AGENT', 'grassroots_token_agent')
+    # this is under the assumption that token agent would have default token being the reserve token. is this the case?
```
wangfowen: do we know if this is a valid assumption?

app/server/models/ussd.py
enjeyw: yeah I've been caught out by this. annoying that it's required.

app/server/utils/ussd/token_processor.py
enjeyw: So we need a 'GE' flag or something to ensure the limit doesn't get applied to non-ge tokens?

test/functional/app/api/test_ussd_api.py
enjeyw: ummmmm we would import it, and then call the methods?

test/unit/app/utils/ussd/test_token_processor.py
enjeyw: I don't think this is a valid assumption right now, but it _should be_. I'll create a todo for 'review token agent setup'

app/server/utils/ussd/token_processor.py
wangfowen: or i could drop the raise exception and make it so we get the first limit that has a percentage on it. but then if we add other percentage based limits in the future and they accidentally get applied to GE users, it could produce the wrong behavior

enjeyw: > but then if we add other percentage based limits in the future and they accidentally get applied to GE users, it could produce the wrong behavior

That feels bad. My instinct is that filtering should just be smart enough to filter as required. 'Explicit is better than implicit' as well...

PR #62, opened 2019-12-02
Dev deploy

enjeyw:


eth_worker/strip_contracts_to_abi_and_bytecode.py
```diff
@@ -2,7 +2,7 @@
 import json
 import os
 
-for path in glob.glob('./complete_compiled_contracts/*.json'):
+for path in glob.glob('/Users/Nick/WebstormProjects/BancorContracts/solidity/build/contracts/*.json'):
```
tristanhcole: @enjeyw check this isn't an accidental commit

app/migrations/seed.py
```diff
@@ -287,7 +293,99 @@ def update_or_create_menu(name, description, parent_id=None):
...
+
+    master_organisation = Organisation.master_organisation()
+    if master_organisation is None:
+        print('Creating master organisation')
+        if reserve_token:
+            print('Binding to reserve token')
+        master_organisation = Organisation(is_master=True, token=reserve_token)
+        db.session.add(master_organisation)
+
+    db.session.commit()
```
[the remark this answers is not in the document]
enjeyw: nah not supposed to be a flush because it can be used as an individual complete item

eth_worker/strip_contracts_to_abi_and_bytecode.py
enjeyw: oops yep

PR #63, opened 2019-12-05
Multi orgs

tristanhcole:
bunch of todo's

- [x] Fix undefined transfer accounts
- [ ] Fix g.active_organisation.org_level_transfer_account error
- [ ] Fix infinite load spinner on user page


app/server/utils/user.py
```diff
@@ -202,8 +202,7 @@ def create_transfer_account_user(first_name=None, last_name=None, preferred_lang
         user.set_held_role('BENEFICIARY', 'beneficiary')
 
     if organisation:
-        user.organisations.append(organisation)
```
tristanhcole: @enjeyw A user (non-admin) is tied to an org but they won't have the transfer account associated to them?

app/client/app.jsx
```diff
@@ -22,7 +22,7 @@ const composeSetup = process.env.NODE_ENV !== 'prod' && typeof window === 'objec
   window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ ?
   window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ : compose
 
-const store = createStore(
+export const store = createStore(
```
enjeyw: @tristanhcole this seems to be exported but not used anywhere?

tristanhcole: oh yeah I ended up storing the client org ID in the browser local storage ...

tristanhcole: so doesn't need to be exported anymore...

enjeyw: ok will remove 👍

app/server/api/auth_api.py
```diff
@@ -27,14 +27,19 @@
 
 def get_user_organisations(user):
```
tristanhcole: @enjeyw

app/server/models/user.py
```diff
@@ -502,9 +503,16 @@ def create_admin_auth(self, email, password, tier='view', organisation=None):
...
-            self.organisations.append(organisation)
-            self.default_organisation = organisation
+            self.add_user_to_organisation(organisation, is_admin=True)
 
+    def add_user_to_organisation(self, organisation: Organisation, is_admin=False):
+        self.organisations.append(organisation)
+        self.default_organisation = organisation
+
+        if organisation.org_level_transfer_account is None:
+            organisation.org_level_transfer_account = db.session.query(server.models.transfer_account.TransferAccount).execution_options(show_all=True).get(organisation.org_level_transfer_account_id)
```
enjeyw: line length @tristanhcole!

PR #64, opened 2019-12-05
implement directory listing

wangfowen:
directory listing fully functional and tested. code isn't great and it's no longer on a background thread...

screenshot of there not being any users matching and then there being users matching:
Screen Shot 2019-12-05 at 3 41 16 PM

### DIFF
--- a/app/server/api/user_api.py
+++ b/app/server/api/user_api.py
@@ -131,11 +131,11 @@ class UserAPI(MethodView):
         )
 
         if response_code == 200:
-            db.session.commit()
-
             user = response_object['data']['user']
             if organisation is not None and user is not None:
                 organisation.send_welcome_sms(user)
+
+            db.session.commit()
 
         return make_response(jsonify(response_object)), response_code
 
--- a/app/server/models/organisation.py
+++ b/app/server/models/organisation.py
@@ -58,12 +58,12 @@ class Organisation(ModelBase):
     def master_organisation() -> "Organisation":
         return Organisation.query.filter_by(is_master=True).first()
 
-    def send_welcome_sms(self, to_user):
+    def send_welcome_sms(self, to_user: dict):
         if self.custom_welcome_message_key:
             message = i18n_for(to_user, "organisation.{}".format(self.custom_welcome_message_key))
         else:
             message = i18n_for(to_user, "organisation.generic_welcome_message")
-        message_processor.send_message(to_user.phone, message)
+        message_processor.send_message(to_user.get('phone'), message)
 
 
     def __init__(self, is_master=False, **kwargs):
--- a/app/server/models/user.py
+++ b/app/server/models/user.py
@@ -99,8 +99,12 @@ class User(ManyOrgBase, ModelBase):
         secondary=user_transfer_account_association_table,
         back_populates="users")
 
-    # TODO: work out if this should be deprecated
     default_transfer_account_id = db.Column(db.Integer, db.ForeignKey('transfer_account.id'))
+
+    default_transfer_account = db.relationship('TransferAccount',
+                                           primaryjoin='TransferAccount.id == User.default_transfer_account_id',
+                                           lazy=True,
+                                           uselist=False)
 
     default_organisation_id = db.Column(
         db.Integer, db.ForeignKey('organisation.id'))
--- a/app/server/utils/i18n.py
+++ b/app/server/utils/i18n.py
@@ -3,7 +3,8 @@ import i18n
 
 # kwargs are for placeholders
 def i18n_for(user, key: str, **kwargs) -> str:
-    preferred_language = user.get('preferred_language') if isinstance(user, dict) else user.preferred_language
-    if user is not None and preferred_language is not None:
-        i18n.set('locale', preferred_language)
+    if user is not None:
+        preferred_language = user.get('preferred_language') if isinstance(user, dict) else user.preferred_language
+        if preferred_language is not None:
+            i18n.set('locale', preferred_language)
     return i18n.t(key, **kwargs)
--- a/app/server/utils/i18n.py
+++ b/app/server/utils/i18n.py
@@ -3,6 +3,7 @@ import i18n
 
 # kwargs are for placeholders
 def i18n_for(user, key: str, **kwargs) -> str:
-    if user is not None and user.preferred_language is not None:
-        i18n.set('locale', user.preferred_language)
+    preferred_language = user.get('preferred_language') if isinstance(user, dict) else user.preferred_language
+    if user is not None and preferred_language is not None:
+        i18n.set('locale', preferred_language)
     return i18n.t(key, **kwargs)
--- a/app/server/utils/user.py
+++ b/app/server/utils/user.py
@@ -231,10 +231,8 @@ def create_transfer_account_user(first_name=None, last_name=None, preferred_lang
             transfer_account.approve_and_disburse()
 
         db.session.add(transfer_account)
-        #TODO: if i don't have this, default_transfer_account_id is null, if i do, API crashes...
-        #db.session.commit()
 
-    user.default_transfer_account_id = transfer_account.id
+    user.default_transfer_account = transfer_account
 
     return user
 
--- a/app/server/utils/user.py
+++ b/app/server/utils/user.py
@@ -231,7 +231,8 @@ def create_transfer_account_user(first_name=None, last_name=None, preferred_lang
             transfer_account.approve_and_disburse()
 
         db.session.add(transfer_account)
-        db.session.commit()
+        #TODO: if i don't have this, default_transfer_account_id is null, if i do, API crashes...
+        #db.session.commit()
 
     user.default_transfer_account_id = transfer_account.id
 
--- a/app/server/utils/ussd/directory_listing_processor.py
+++ b/app/server/utils/ussd/directory_listing_processor.py
@@ -9,7 +9,7 @@ from server.models.transfer_account import TransferAccount
 from server.utils.user import default_token
 
 
-def user_directory_listing(user):
+def user_directory_listing(user: User) -> str:
     bio = user.custom_attributes.filter_by(name='bio').first()
     if bio is None:
         return user.phone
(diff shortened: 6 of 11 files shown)


app/server/utils/user.py
```diff
@@ -231,7 +231,8 @@ def create_transfer_account_user(first_name=None, last_name=None, preferred_lang
             transfer_account.approve_and_disburse()
 
         db.session.add(transfer_account)
-        db.session.commit()
+        #TODO: if i don't have this, default_transfer_account_id is null, if i do, API crashes...
```
wangfowen: i have no idea how to solve this

app/server/utils/ussd/directory_listing_processor.py
```diff
@@ -1,39 +1,35 @@
...
+        return f"{user.phone} {bio.value}"
 
 
-class DirectoryListingProcessor(threading.Thread):
+class DirectoryListingProcessor(object):
     def __init__(self, user: User, transfer_usage: TransferUsage):
         self.recipient = user
         self.selected_business_category = transfer_usage
 
-        threading.Thread.__init__(self)
```
wangfowen: got rid of it being on a background thread, otherwise i get these fun exceptions:

```
Traceback (most recent call last):
  File "/usr/local/Cellar/python/3.6.5_1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/Users/owen/workspace/sempo/SempoBlockchain/app/server/utils/ussd/directory_listing_processor.py", line 30, in run
    self.runnable()
  File "/Users/owen/workspace/sempo/SempoBlockchain/app/server/utils/ussd/directory_listing_processor.py", line 82, in send_directory_listing
    transfer_account = TransferAccount.query.get(self.recipient.default_transfer_account_id)
  File "/Users/owen/workspace/sempo/SempoBlockchain/venv/lib/python3.6/site-packages/flask_sqlalchemy/__init__.py", line 519, in __get__
    return type.query_class(mapper, session=self.sa.session())
  File "/Users/owen/workspace/sempo/SempoBlockchain/venv/lib/python3.6/site-packages/sqlalchemy/orm/scoping.py", line 78, in __call__
    return self.registry()
  File "/Users/owen/workspace/sempo/SempoBlockchain/venv/lib/python3.6/site-packages/sqlalchemy/util/_collections.py", line 1012, in __call__
    return self.registry.setdefault(key, self.createfunc())
  File "/Users/owen/workspace/sempo/SempoBlockchain/venv/lib/python3.6/site-packages/sqlalchemy/orm/session.py", line 3214, in __call__
    return self.class_(**local_kw)
  File "/Users/owen/workspace/sempo/SempoBlockchain/venv/lib/python3.6/site-packages/flask_sqlalchemy/__init__.py", line 136, in __init__
    self.app = app = db.get_app()
  File "/Users/owen/workspace/sempo/SempoBlockchain/venv/lib/python3.6/site-packages/flask_sqlalchemy/__init__.py", line 982, in get_app
    'No application found. Either work inside a view function or push'
RuntimeError: No application found. Either work inside a view function or push an application context. See http://flask-sqlalchemy.pocoo.org/contexts/.
```

what should the proper fix be?

enjeyw: Might be fixed by the `copy_current_request_context` decorator:
https://flask.palletsprojects.com/en/1.1.x/api/#flask.copy_current_request_context

app/server/utils/ussd/directory_listing_processor.py
```diff
@@ -42,20 +38,38 @@ def get_directory_listing_users(self):
...
-            User.business_usage_id == self.selected_business_category.id,
-            TransferAccount.token_id == token_id,
-            CustomAttributeUserStorage.name == 'market_enabled',
-            CustomAttributeUserStorage.value == True).order_by(
-            User.transaction_count
-        ).limit(5).all()
-
-        return users
+        # users with no credit transfers still have count of 1 since left join... not great but it works
+        sql = text('''
```
enjeyw: oouch full sql...

app/server/utils/ussd/directory_listing_processor.py
```diff
@@ -1,39 +1,35 @@
...
 from server.models.transfer_account import TransferAccount
+from server.utils.user import default_token
 
 
 def user_directory_listing(user):
-    # TODO: add custom attribute bio
-    return f"{user.phone}"
+    bio = user.custom_attributes.filter_by(name='bio').first()
+    if bio is None:
+        return user.phone
```
enjeyw: does this need to be stringified to work like in the above code?

app/server/api/user_api.py
```diff
@@ -131,12 +131,12 @@ def post(self, user_id):
         )
 
         if response_code == 200:
-            db.session.commit()
-
             user = response_object['data']['user']
             if organisation is not None and user is not None:
                 organisation.send_welcome_sms(user)
 
+            db.session.commit()
```
enjeyw: if the commit happens not last, then if `organisation.send_welcome_sms(user)` fails, the api will return a `500` or whatever, so the front end will say the user isn't created when they actually have been.

app/server/models/organisation.py
```diff
@@ -58,12 +58,12 @@ class Organisation(ModelBase):
...
         return Organisation.query.filter_by(is_master=True).first()
 
-    def send_welcome_sms(self, to_user):
+    def send_welcome_sms(self, to_user: dict):
         if self.custom_welcome_message_key:
             message = i18n_for(to_user, "organisation.{}".format(self.custom_welcome_message_key))
         else:
             message = i18n_for(to_user, "organisation.generic_welcome_message")
-        message_processor.send_message(to_user.phone, message)
+        message_processor.send_message(to_user.get('phone'), message)
```
enjeyw: `to_user` is a dict, so needs to use `.get(`

app/server/utils/i18n.py
```diff
@@ -3,6 +3,7 @@
 
 # kwargs are for placeholders
 def i18n_for(user, key: str, **kwargs) -> str:
-    if user is not None and user.preferred_language is not None:
-        i18n.set('locale', user.preferred_language)
+    preferred_language = user.get('preferred_language') if isinstance(user, dict) else user.preferred_language
```
enjeyw: same issue as above

app/server/utils/user.py
enjeyw: see review comment!

app/server/utils/user.py
```diff
@@ -231,7 +231,8 @@ def create_transfer_account_user(first_name=None, last_name=None, preferred_lang
...
+        #db.session.commit()
 
     user.default_transfer_account_id = transfer_account.id
```
enjeyw: was this the bit where i suggested on a different PR doing 
 `user.default_transfer_account = transfer_account`
and u were like `nah doesn't work?`

enjeyw: not sure why i turned ur quote into a code block, but you're welcome :P

app/server/utils/ussd/directory_listing_processor.py
wangfowen: do we want this task to run in a different thread since it's doing a massive query?

app/server/utils/ussd/directory_listing_processor.py
wangfowen: nope because we just stick the returned content in a formatted string, but you're right i should make it a string for consistency

app/server/utils/user.py
wangfowen: yeah, tried it and still was null

wangfowen: might be because we don't have a line similar to `default_organisation`?

```
    default_transfer_account_id = db.Column(db.Integer, db.ForeignKey('transfer_account.id'))

    default_organisation_id = db.Column(
        db.Integer, db.ForeignKey('organisation.id'))

    default_organisation = db.relationship('Organisation',
                                           primaryjoin=Organisation.id == default_organisation_id,
                                           lazy=True,
                                           uselist=False)
```

because if i add it, then c i r c u l a r d e p e n d e n c y

app/server/utils/i18n.py
wangfowen: huh, how has this worked thus far then? we do have tests that test both translations

app/server/models/organisation.py
wangfowen: wait are you sure? tests are now failing:
```
AttributeError: 'User' object has no attribute 'get'
```

wangfowen: oh it's because my test makes a bad assumption

app/server/utils/i18n.py
wangfowen: this seems wrong, user is normally actually a user passed in

wangfowen: oh wait i see, tests if is dict

app/server/utils/ussd/directory_listing_processor.py
wangfowen: oh wait, user.phone is already a string lol. but i've added type checking now to ensure it

app/server/utils/ussd/directory_listing_processor.py
wangfowen: i'll leave it as sync for now since everything else is also sync. if we find it to be too slow we can throw it in a different thread again

app/server/utils/ussd/directory_listing_processor.py
wangfowen: yeah.. brutal query

app/server/models/user.py
```diff
@@ -99,9 +99,13 @@ class User(ManyOrgBase, ModelBase):
         secondary=user_transfer_account_association_table,
         back_populates="users")
 
-    # TODO: work out if this should be deprecated
     default_transfer_account_id = db.Column(db.Integer, db.ForeignKey('transfer_account.id'))
 
+    default_transfer_account = db.relationship('TransferAccount',
```
wangfowen: @enjeyw looks like we can make the primaryjoin a string as well, so that avoids the dependency issue. and then we're able to do the line you suggested yayy

PR #65, opened 2019-12-07
Transfer Type Improvments

enjeyw:
Based on slack discussion regarding transfer types not being set for exchanges. Will also probably include a deeper investigation into transfer subtypes


tristanhcole:
Looks good

PR #66, opened 2019-12-07
Migration fix

tristanhcole:
is_ghost migration fix


tristanhcole:
@wangfowen think you forgot to commit your migrations file! :)

tristanhcole:
@wangfowen my bad. for some reason, the migration wasn't working!

PR #67, opened 2019-12-08
Multi orgs bug fixes

tristanhcole:
- bug fixes
- auto login invited users once password (and if required TFA token) are set

### DIFF
--- a/app/client/api/authApi.js
+++ b/app/client/api/authApi.js
@@ -54,7 +54,7 @@ export const registerAPI = ({body}) => {
     body: JSON.stringify(body)
     })
     .then(response => {
-      return handleResponse(response)
+      return response.json();
     })
     .catch(error => {
       throw error;
--- a/app/client/api/authApi.js
+++ b/app/client/api/authApi.js
@@ -1,4 +1,4 @@
-import { handleResponse, storeSessionToken, removeSessionToken, getToken, getTFAToken, generateQueryString, getOrgId } from '../utils'
+import {handleResponse, getToken, getTFAToken, generateQueryString, getOrgId, generateFormattedURL} from '../utils'
 import { startConfiguration } from 'pusher-redux';
 
 //Auth API Call
@@ -111,7 +111,7 @@ export const GetTFAAPI = () => {
   }).catch(error => {
       throw error;
   })
-}
+};
 
 
 export const ValidateTFAAPI = (payload) => {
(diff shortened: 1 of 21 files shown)


tristanhcole:
[Admin Sign Up Flow Fix](https://trello.com/c/GiE6eXKX/127-admin-sign-up-flow-fix)

app/client/api/authApi.js
```diff
@@ -54,7 +54,7 @@ export const registerAPI = ({body}) => {
     body: JSON.stringify(body)
     })
     .then(response => {
-      return handleResponse(response)
+      return response.json();
```
enjeyw: What was the issue that handle response was causing that requires non standard handling?

app/client/api/creditTransferAPI.js
```diff
@@ -1,13 +1,13 @@
...
+    var query_string = generateQueryString(query);
     if (query) {
-        const query_string = generateQueryString(query);
         var URL = `/api/credit_transfer/${query_string}`;
     } else if (path) {
-        URL = `/api/credit_transfer/${path}/`;
+        URL = `/api/credit_transfer/${path}/${query_string}`;
     } else {
-        URL = '/api/credit_transfer/';
+        URL = `/api/credit_transfer/${query_string}`;
```
enjeyw: I used to be put off by string literals, now I lovvvve them 😂

app/client/api/authApi.js
tristanhcole: TFA returns a 401 that causes `handleResponse()` to raise an error. This is the same for the login API (where TFA is also active). Just a special case :)

https://github.com/teamsempo/SempoBlockchain/blob/713de7fffc712b9a06052002bc70434f5fe7b542/app/client/utils.js#L54-L59

app/client/api/creditTransferAPI.js
tristanhcole: I feel like we could have a `generateFormattedURL(`/api/credit_transfer/`, query, path)` util that generates a correct URL automatically ...

enjeyw:
I was literally going to suggest the same thing!

app/client/api/authApi.js
enjeyw: Ohhh yeah that makes sense

app/client/api/creditTransferAPI.js
enjeyw: I was literally gonna suggest the same thing, but I couldn’t find any other location where we need to parse the path.

tristanhcole: cool refactored

PR #71, opened 2019-12-10
bunch of admin fix things

wangfowen:
- reset pin flow - on admin user page, if user has more than 0 failed pin attempts, a button appears to reset their pin. then user will go through reset pin flow next time they use ussd
- fix see my business ussd menu - bad menu key
- error out if bad menu so that we get logs - wasn't catching bad menu key error
- show correct currencies on accounts and transfers - based on token of credit transfer/transfer account. still not doing it for create new user or on dashboard
- fix initial disbursement - wasn't working before even when config was set. we should do a full audit of the is_beneficiary situation. had several times now of catching new places where behavior was incorrect


app/client/api/userAPI.js
```diff
@@ -54,4 +54,22 @@ export const editUserAPI = ({body, path}) => {
     .catch(error => {
         throw error;
     })
-};
\ No newline at end of file
+};
+
+export const resetPinAPI = (userId) => {
```
wangfowen: we gotta refactor some of this. tons of code repeat across the APIs

app/client/components/user/singleUserManagement.jsx
```diff
@@ -225,6 +225,25 @@ class SingleUserManagement extends React.Component {
                           </ManagerText>
                         </SubRow>
                       </Row>
+                    <Row style={{margin: '0em 1em'}}>
```
wangfowen: this whole component could use a lot of refactoring

app/client/reducers/userReducer.js
```diff
@@ -71,6 +75,27 @@ const editStatus = (state = initialEditStatusState, action) => {
   }
 };
 
+const initialPinStatus = {
+  isRequesting: false,
+  error: null,
+  success: false
+};
+const pinStatus = (state = initialPinStatus, action) => {
```
wangfowen: all of these request/success flows also could use refactoring haha...

app/client/sagas/creditTransferSagas.js
```diff
@@ -27,10 +27,11 @@ import {handleError} from "../utils";
 
 function* updateStateFromCreditTransfer(result) {
   //Schema expects a list of credit transfer objects
+  let credit_transfer_list = [];
   if (result.data.credit_transfers) {
-    var credit_transfer_list = result.data.credit_transfers
```
wangfowen: this is dangerous code, could end up with unexpected global variables. always declare a `var` (ES6 `const` or `let`) top of scope before using it in lower scopes

app/client/sagas/userSagas.js
```diff
@@ -87,6 +88,30 @@ function* watchEditUser() {
   yield takeEvery(EDIT_USER_REQUEST, editUser);
 }
 
+function* resetPin({userId}) {
```
wangfowen: did i mention we should do some refactoring?

app/server/models/user.py
```diff
@@ -567,7 +567,7 @@ def has_valid_pin(self):
         self.clear_expired_pin_reset_tokens()
         not_resetting = len(self.pin_reset_tokens) == 0
 
-        return self.pin_hash is not None and not_resetting
+        return self.pin_hash is not None and not_resetting and self.failed_pin_attempts < 3
```
wangfowen: now if a user is blocked, they can't use USSD at all until we reset their pin. otherwise it was ending up in some weird places like trying to reset pin even though they had 0 attempts left. didn't let them know they were blocked til the very end. 
would we want this? downside i can think of is if we have ussd menus that don't need pin, they can't access those anymore either

app/client/components/creditTransfer/creditTransferList.jsx
```diff
@@ -324,22 +323,33 @@ class CreditTransferList extends React.Component {
                 Header: "Amount",
                 accessor: "transfer_amount",
                 headerClassName: 'react-table-header',
-                Cell: cellInfo => (<p style={{margin: 0}}>{formatMoney(cellInfo.value / 100)}</p>),
+                Cell: cellInfo => {
+                  let currency;
+                  const transferAccountId = cellInfo.original.sender_transfer_account_id;
+                  if (transferAccountId) {
+                    // this is not ideal... would be better if credit transfer just had the associated transfer account
```
wangfowen: @tristanhcole @enjeyw

tristanhcole:
[Refactor web API](https://trello.com/c/AJBoE83s/136-refactor-web-api)

app/client/api/userAPI.js
tristanhcole: added todo: https://trello.com/c/AJBoE83s/136-refactor-web-api

app/client/sagas/creditTransferSagas.js
tristanhcole: thanks, nice catch

app/client/reducers/userReducer.js
tristanhcole: added todo: https://trello.com/c/cGlzTuiu/137-refactor-web-reducers-sagas

app/client/sagas/userSagas.js
tristanhcole: added todo: https://trello.com/c/cGlzTuiu/137-refactor-web-reducers-sagas

app/server/models/user.py
tristanhcole: This sounds reasonable to me. What do you think @enjeyw ?

app/client/components/creditTransfer/creditTransferList.jsx
tristanhcole: hmm need to think about this

PR #73, opened 2019-12-12
Dashboard Updates & Fixes

tristanhcole:
* adding exchange metric
* correctly handling exchange object on dashboard

[UPDATE]
* adding fixes to login bug
* adding fixes to web api versioning (fallback and /activate API)


app/client/components/dashboard/beneficiaryLiveFeed.jsx
```diff
@@ -69,13 +71,45 @@ class BeneficiaryLiveFeed extends React.Component {
...
+                            let currency, fromExchangeTransfer, recipientCurrency;
+                            const transferAccountId = transfer.sender_transfer_account_id;
+                            if (transferAccountId) {
+                              const transferAccount = this.props.transferAccounts.byId[transferAccountId];
+                              currency = transferAccount && transferAccount.token && transferAccount.token.symbol;
+                            }
+                            const money = formatMoney(transfer.transfer_amount / 100, undefined, undefined, undefined, currency);
+
+
+                            if (transfer.from_exchange_to_transfer_id !== null && typeof(transfer.from_exchange_to_transfer_id) !== "undefined") {
```
wangfowen: `transfer.from_exchange_to_transfer_id !== undefined` is okay also

app/client/components/dashboard/beneficiaryLiveFeed.jsx
```diff
@@ -69,13 +71,45 @@ class BeneficiaryLiveFeed extends React.Component {
                               </UserTime>
                             );
 
-                            if (transfer.transfer_type === 'PAYMENT') {
+                            let currency, fromExchangeTransfer, recipientCurrency;
```
wangfowen: it's a bit more tedious, but best practice is to make each variable a new line:

```
let currency;
let fromExchangeTransfer;
let recipientCurrency;
```

this way if anything ever changes, you'll quickly see if one should be a `const` instead, and more importantly, if one of them is ever moved to a new line, you won't accidentally end up with a global. #justjavascriptthings

app/client/components/dashboard/beneficiaryLiveFeed.jsx
```diff
@@ -69,13 +71,45 @@ class BeneficiaryLiveFeed extends React.Component {
...
+                            const transferAccountId = transfer.sender_transfer_account_id;
+                            if (transferAccountId) {
+                              const transferAccount = this.props.transferAccounts.byId[transferAccountId];
+                              currency = transferAccount && transferAccount.token && transferAccount.token.symbol;
+                            }
+                            const money = formatMoney(transfer.transfer_amount / 100, undefined, undefined, undefined, currency);
```
wangfowen: `transferFromMoney` would be more clear

app/server/schemas.py
```diff
@@ -147,6 +147,8 @@ def filter_rejected(self, data, many):
     sender_transfer_account    = fields.Nested("server.schemas.TransferAccountSchema", only=("id", "balance", "token"))
     recipient_transfer_account = fields.Nested("server.schemas.TransferAccountSchema", only=("id", "balance", "token"))
 
+    from_exchange_to_transfer_id = fields.Function(lambda obj: obj.from_exchange.to_transfer.id)
```
wangfowen: this naming's a bit confusing, at first i though it was a conversion "from_exchange" "to_transfer" which didn't make sense. then i thought it was "from_exchange_to"'s "transfer", especially because the resulting variable was called `fromExchangeTransfer`, so i thought `fromExchangeTransfer.transfer_amount` was the wrong amount, that it'd be the amount exchanged from. only when i came and saw this line did i understand it

app/client/components/dashboard/beneficiaryLiveFeed.jsx
```diff
@@ -69,13 +71,45 @@ class BeneficiaryLiveFeed extends React.Component {
...
+
+                            if (transfer.from_exchange_to_transfer_id !== null && typeof(transfer.from_exchange_to_transfer_id) !== "undefined") {
+
+                              fromExchangeTransfer = this.props.creditTransfers.byId[transfer.from_exchange_to_transfer_id];
+                              const transferAccountId = fromExchangeTransfer.sender_transfer_account_id;
+                              if (transferAccountId) {
+                                const transferAccount = this.props.transferAccounts.byId[transferAccountId];
+                                recipientCurrency = transferAccount && transferAccount.token && transferAccount.token.symbol;
+                              }
+                              const transferToMoney = formatMoney(fromExchangeTransfer.transfer_amount / 100, undefined, undefined, undefined, recipientCurrency);
```
wangfowen: per comment on schema, `fromExchangeTransfer. transfer_amount` is a bit confusing. thought it was the exchanged from amount. and then shouldn't it be `exchangedToMoney`?

app/client/components/dashboard/beneficiaryLiveFeed.jsx
```diff
@@ -69,13 +71,45 @@ class BeneficiaryLiveFeed extends React.Component {
...
+                                )
+                              }
+                            } else if (transfer.transfer_type === 'PAYMENT') {
                                 return (
                                     <UserWrapper key={transfer.id}>
                                       <UserSVG src="/static/media/transfer.svg"/>
                                       <UserGroup>
-                                        <TopText>{sender_user_name} transfer</TopText>
-                                        <BottomText><Highlight>{formatMoney(transfer.transfer_amount / 100)}</Highlight> to
+                                        <TopText>{sender_user_name} transfered</TopText>
```
wangfowen: transferred

app/client/components/dashboard/metricsBar.jsx
```diff
@@ -21,41 +21,43 @@ const mapDispatchToProps = (dispatch) => {
...
-
-    let beneficiaryTermPlural = window.BENEFICIARY_TERM_PLURAL;
-
-    if (Object.keys(creditTransferStats).length == 0) {
-      return (
-        <p>No Transfer Data</p>
-      );
-    }
-
+  if (Object.keys(creditTransferStats).length == 0) {
```
wangfowen: `===` javascript is the worrrsttt

app/server/utils/credit_transfer.py
```diff
@@ -46,6 +46,11 @@ def calculate_transfer_stats(total_time_series=False):
             .filter(CreditTransfer.transfer_type == TransferTypeEnum.PAYMENT).first().total
     )
 
+    total_exchanged = (
+        db.session.query(func.sum(CreditTransfer.transfer_amount).label('total'))
```
wangfowen: so this would automatically do it just for the org right? do we have to worry about currency conversions? or would all exchanges be from/to the same currencies?

app/client/components/dashboard/beneficiaryLiveFeed.jsx
```diff
@@ -69,13 +71,45 @@ class BeneficiaryLiveFeed extends React.Component {
                               </UserTime>
                             );
 
-                            if (transfer.transfer_type === 'PAYMENT') {
+                            let currency, fromExchangeTransfer, recipientCurrency;
+                            const transferAccountId = transfer.sender_transfer_account_id;
+                            if (transferAccountId) {
+                              const transferAccount = this.props.transferAccounts.byId[transferAccountId];
+                              currency = transferAccount && transferAccount.token && transferAccount.token.symbol;
```
enjeyw: we should use this syntax more. it's sooo clleaaan

tristanhcole: yeah I really like `currency = transferAccount && transferAccount.token && transferAccount.token.symbol;`... feels like python.

@wangfowen's code :)

app/client/components/dashboard/beneficiaryLiveFeed.jsx
```diff
@@ -69,13 +71,45 @@ class BeneficiaryLiveFeed extends React.Component {
...
+                            }
+                            const money = formatMoney(transfer.transfer_amount / 100, undefined, undefined, undefined, currency);
+
+
+                            if (transfer.from_exchange_to_transfer_id !== null && typeof(transfer.from_exchange_to_transfer_id) !== "undefined") {
+
+                              fromExchangeTransfer = this.props.creditTransfers.byId[transfer.from_exchange_to_transfer_id];
```
enjeyw: this feels backwards?
`transfer` as on line 83 is the `from_transfer` from the perspective of the whole exchange
 so the transfer on line 85 should be the `to_transfer`?

app/client/components/dashboard/beneficiaryLiveFeed.jsx
enjeyw: i think it's actually already backwards from the line above... so `transferToMoney` is correct, but `fromExchangeTransfer` should be `toExchangeTransfer`

app/client/components/dashboard/metricsBar.jsx
enjeyw: I've started writing `!==` everywhere in my python code and wondering why it doesn't work.

app/client/components/pages/transferAccountListPage.jsx
```diff
@@ -16,7 +16,7 @@ const mapStateToProps = (state) => {
     transferAccounts: state.transferAccounts,
     mergedTransferAccountUserList: Object.keys(state.transferAccounts.byId)
       .map((id) => {return {...{id, ...state.users.byId[state.transferAccounts.byId[id].primary_user_id]}, ...state.transferAccounts.byId[id]}})
-      .filter(mergedObj => Object.keys(mergedObj).length > 2),  // added to remove {id: 1, balance: 12} objects
+      .filter(mergedObj => Object.keys(mergedObj).length > 3),  // added to remove {id: 1, balance: 12, token: {...}} objects
```
enjeyw: feels like there's lot of opportunities for inferring appropriateness just based on length to go wrong - is it possible to explicitly require some attributes exist instead? more fundamentally, this feels like bad data coming from the backend that i should look into...

app/server/schemas.py
enjeyw: yeah i had the same brain-ouchie. Maybe just even exchange_to_transfer_id - doesn't solve the full problem, but the word 'from' isn't really necessary.

app/server/utils/credit_transfer.py
enjeyw: i can definitely see opportunities for this to go wrong. Might be better to explicity filter on the token type being the organisation's so we don't have weirdness there. Even better still would be to list total exchanges in and total exchanges out as two different numbers, because that's where u get meaning from the data.

app/client/components/dashboard/beneficiaryLiveFeed.jsx
```diff
@@ -69,13 +71,45 @@ class BeneficiaryLiveFeed extends React.Component {
...
+                              const transferAccountId = fromExchangeTransfer.sender_transfer_account_id;
+                              if (transferAccountId) {
+                                const transferAccount = this.props.transferAccounts.byId[transferAccountId];
+                                recipientCurrency = transferAccount && transferAccount.token && transferAccount.token.symbol;
+                              }
+                              const transferToMoney = formatMoney(fromExchangeTransfer.transfer_amount / 100, undefined, undefined, undefined, recipientCurrency);
+
+                              if (transfer.transfer_type === 'EXCHANGE') {
```
enjeyw: i know it's a small change, but it feels like we should put `(transfer.transfer_type === 'EXCHANGE')` on the outside at the same conditional level as `(transfer.transfer_type === 'EXCHANGE')` and put the above code inside that? just makes it easier to reason with if the same check types are all at the same level...

app/client/components/dashboard/beneficiaryLiveFeed.jsx
tristanhcole: Ok, thanks for the tip!

app/client/components/dashboard/beneficiaryLiveFeed.jsx
tristanhcole: What do you mean by `transferFromMoney`?

tristanhcole: Ahh I get it

app/server/schemas.py
tristanhcole: The reason I used `from_exchange_to_transfer` is that it actually follows the existing model relationships...

`from_exchange` = https://github.com/teamsempo/SempoBlockchain/blob/76228540b8fad295dee3938f0c8902642fe699e0/app/server/models/credit_transfer.py#L64
`to_transfer_id` = https://github.com/teamsempo/SempoBlockchain/blob/76228540b8fad295dee3938f0c8902642fe699e0/app/server/models/exchange.py#L128

app/client/components/dashboard/beneficiaryLiveFeed.jsx
tristanhcole: I'm confused now 😕

app/server/utils/credit_transfer.py
tristanhcole: Yes, this automatically filters based on the `?org=1` query string. 

@enjeyw Should we calculate _valid_ exchanges as ones with the `organisation.token_id` being equal to `exchange.from_token_id` OR `exchange.to_token_id` ?

app/client/components/pages/transferAccountListPage.jsx
tristanhcole: You could add a filter to the transfer account list to check for `transfer_account.users` not being undefined or null... i.e. the transfer account has at least one user associated with it?

app/server/schemas.py
enjeyw: oh yeah i get that lol - it's just confusing when strung into a single variable name. The only reason it's called `from_exchange` and not just `exchange` on the model is because it would collide names with the other exhcnage relationship

app/server/utils/credit_transfer.py
enjeyw: ^ so based on that suggestion above, i'd split them out into two seperate ones -> exchanges into the currency and exchanges out - exchanges in and out combined isn't really that useful operational info IMO

app/client/components/pages/transferAccountListPage.jsx
enjeyw: yeah that sounds good

app/client/components/dashboard/beneficiaryLiveFeed.jsx
enjeyw: @tristanhcole here

app/client/components/dashboard/beneficiaryLiveFeed.jsx
enjeyw: comment i just tagged u in should explain it.

app/server/schemas.py
tristanhcole: open to improved naming suggestions! 😄

app/client/components/dashboard/beneficiaryLiveFeed.jsx
tristanhcole: Updated

tristanhcole:
updating:

* adding fixes to login bug
* adding fixes to web api versioning (fallback and /activate API)

app/client/components/pages/transferAccountListPage.jsx
tristanhcole: https://github.com/teamsempo/SempoBlockchain/blob/b21953f9b95525b5cacddcfdc8319e030094592c/app/client/components/pages/transferAccountListPage.jsx#L19

PR #75, opened 2019-12-13
load organisation to redux for token in create user

wangfowen:


app/client/api/organisationApi.js
```diff
@@ -0,0 +1,16 @@
+import {generateFormattedURL, getToken, handleResponse} from "../utils";
```
wangfowen: more code repeat to refactor out i'm sorryyyy

app/client/components/user/CreateUserForm.tsx
```diff
@@ -132,6 +135,8 @@ export default connect(state => {
   const selector = formValueSelector('createUser');
   return {
     businessUsageValue: selector(state, 'businessUsage'),
+    // @ts-ignore
+    organisation: state.organisation.data
```
wangfowen: i thiiink we need to convert all the reducers to typescript to get the typing working correctly for this, would be really nice if we do

app/client/sagas/organisationSagas.js
```diff
@@ -1 +1,48 @@
-//TODO: use the new organisation reducers
```
wangfowen: again more repeat :(

app/server/me_api/organisation.py
```diff
@@ -0,0 +1,23 @@
...
+
+from server.schemas import me_organisation_schema
+from server.utils.auth import requires_auth
+
+
+class MeOrganisationAPI(MethodView):
+    @requires_auth
+    def get(self):
+
+        organisation = g.user.default_organisation
```
wangfowen: do we want to use default organisation?

app/client/api/organisationApi.js
enjeyw: lolllll my fault for building it this way in the first place

app/client/components/user/CreateUserForm.tsx
enjeyw: yeah, it's annoying that it can't be incremental

enjeyw: would be good to do at the same time as eliminating that YUGGGEEE amount of copypasta boilerplate

app/server/me_api/organisation.py
enjeyw: shouldn't this take the active organisation defined by the query string?

wangfowen: so every request now passes in org in the query string? always as 'org'?

PR #77, opened 2019-12-17
Query string auth

enjeyw:
Adding query string auth capacity so that we can properly secure our USSD api endpoint.
Takes auth as `/api/v1/ussd/kenya?username='foo'&password='bar'`, where username and password are from the admin setting page.

Also added a few more tests to auth


app/server/api/auth_api.py
```diff
@@ -2,134 +2,21 @@
...
-from server.utils.feedback import request_feedback_questions
 from server.utils.amazon_ses import send_reset_email, send_activation_email, send_invite_email
-from server.utils.blockchain_transaction import get_usd_to_satoshi_rate
 
-import time, random
+import random
 
 auth_blueprint = Blueprint('auth', __name__)
 
-
```
enjeyw: was kinda gross having this at the top of the api file, so shifted it into utils to be more consistent with other endpoints

app/server/utils/auth.py
```diff
@@ -36,16 +46,54 @@ def requires_auth(f=None,
...
                        allowed_basic_auth_types=allowed_basic_auth_types,
-                       ignore_tfa_requirement = ignore_tfa_requirement)
+                       ignore_tfa_requirement=ignore_tfa_requirement,
+                       allow_query_string_auth=allow_query_string_auth)
 
     @wraps(f)
     def wrapper(*args, **kwargs):
 
-        auth = request.authorization
+        # ----- FIRST GET AUTH VALUES -----
```
enjeyw: This stuff is the actual guts of the PR

app/server/utils/auth.py
```diff
@@ -95,28 +128,6 @@ def wrapper(*args, **kwargs):
                         }
                         return make_response(jsonify(response_object)), 401
 
-                    g.user = user
```
enjeyw: shifted this stuff to post auth, because setting things pre-auth makes me uneasy.

PR #78, opened 2019-12-17
frontend testing framework

tristanhcole:
* nuking old test setup
* updating package.json to correctly setup Jest and Enzyme
* adding frontend tests runner to `config.yml`


tristanhcole:
[Frontend Web testing](https://trello.com/c/mlBIryaj/161-frontend-web-testing)

.circleci/config.yml
```diff
@@ -141,6 +142,28 @@ jobs:
...
+          keys:
+            # when lock file changes, use increasingly general patterns to restore cache
+            - node-v1-{{ .Branch }}-{{ checksum "~/project/app/package-lock.json" }}
+            - node-v1-{{ .Branch }}-
+            - node-v1-
+      - run:
+          name: Run Frontend Tests
+          command: |
+            cd app
+            npm run test
```
tristanhcole: Split out front end tests from back end so CI runs faster
https://circleci.com/docs/2.0/parallelism-faster-jobs/

EDsCODE: parallelizing looks good..also i think the status check for "ci/circleci: test" needs to be removed

tristanhcole: good point - will leave active until this PR is ready to merge in - so that the other branches keep working!

EDsCODE: sounds good! will give it a full look this weekend

app/jest.config.js
```diff
@@ -0,0 +1,28 @@
+// jest.config.js
+module.exports = {
+  verbose: true,
+  roots: [
+    "./client"
+  ],
+  transform: {
+    "^.+\\.[t|j]sx?$": "babel-jest"
```
EDsCODE: we should use "ts-jest" here to transform tsx elements. we might need more coverage on the regex for ts/js elements too but either way ts-jest will allow us to test on tsx files

we can either do it here or i can add it in after the fact when i start writing tests in the next PR!

app/package.json
```diff
@@ -62,20 +52,19 @@
...
-    "enzyme-adapter-react-16": "^1.1.1",
-    "enzyme-to-json": "^3.3.4",
+    "enzyme": "^3.10.0",
+    "enzyme-adapter-react-16": "^1.15.1",
+    "enzyme-to-json": "^3.4.3",
     "husky": "^4.2.3",
-    "identity-obj-proxy": "^3.0.0",
-    "jest": "^23.4.1",
-    "jest-cli": "^23.4.1",
+    "jest": "^24.0.0",
```
EDsCODE: could update these versions to latest/later versions unless there's a specific dependency intended

app/jest.config.js
tristanhcole: good point - updated https://kulshekhar.github.io/ts-jest/user/config/#jest-config-with-helper

> `ts-jest/presets/js-with-babel` | TypeScript files will be handled by ts-jest, and JavaScript files will be handled by babel-jest.

app/package.json
tristanhcole: updated all jest related packages to latest

PR #81, opened 2019-12-18
get typescript root reducer working

wangfowen:
- converted auth reducers to typescript
- turns out upgrading redux to 4.0.0 is sufficient for getting it working, don't need to convert all reducers to typescript. see CreateUser/CreateUserForm, it now recognizes the type of state passed in

don't know about correctness of all the types in the payloads for actions... i looked at their usages and guessed at what they should be. won't cause breakages til we turn usages to typescript as well

still some type errors on root reducer. doesn't break app, but breaks build :(

pretttty sure nothing i deleted breaks things, but hard to tell for sure


app/client/components/adminUserList.jsx
```diff
@@ -199,7 +199,7 @@ class AdminUserList extends React.Component {
       );
 	  }
 
-	  if (this.props.login.adminTier === 'view' || 'subadmin') {
+	  if (this.props.login.adminTier === 'view' || this.props.login.adminTier === 'subadmin') {
```
tristanhcole: @enjeyw double-checking subadmin/enroller type _shouldn't_ have access to the admin list?

This reflects the current auth on backend - `@requires_auth(allowed_roles={'ADMIN': 'admin'})`

app/client/reducers/auth/loginReducer.ts
```diff
@@ -0,0 +1,99 @@
...
+        ...state,
+        isLoggingIn: false,
+        token: null,
+        userId: null,
+        intercomHash: null,
+        webApiVersion: null,
+        organisationName: null,
+        organisationId: null,
+        organisations: null,
+        requireTransferCardExists: null,
```
tristanhcole: @enjeyw On master, this was `requireTransferCardExists: action.requireTransferCardExists` which is undefined and causing one of the login bugs...

PR #84, opened 2019-12-19
Exchange bugs

enjeyw:
Fixing various bugs that were causing exchanges to miss-behave. Summary of changes:

- Increasing available memory for workers in docker containers
- Using Enum TransferType/SubType for logic checks, instead of strings
- Adding a 'Standard'  TransferSubtype
- Transfer limits now:
  - Have improved class signatures to avoid 'Nones' 
  - Have a field for custom filtering of the data that is aggregated 
  - Have tests that are more independent of the implementation logic.


app/server/utils/transfer_limits.py
```diff
@@ -0,0 +1,251 @@
...
+        self.application_filter = application_filter
+        self.time_period_days = time_period_days
+        self.total_amount = total_amount
+        self.transfer_count = transfer_count
+        self.transfer_filter = transfer_filter
+        self.transfer_balance_fraction = transfer_balance_fraction
+
+
+LIMITS = [
+    TransferLimit('Sempo Level 0: P7', [STANDARD_PAYMENT], is_liquid_and_user_is_not_phone_and_not_kyc_verified, 7,
```
enjeyw: I extended the names so that we could filter on them during tests - will this have any negative side effect @tristanhcole ?

test/unit/app/models/test_models.py
```diff
@@ -296,24 +297,21 @@ def test_new_credit_transfer_check_sender_transfer_limits(create_credit_transfer
...
+    from server.utils.transfer_limits import LIMITS
     from server.models.kyc_application import KycApplication
     from server.models import token
-    from server.utils.transfer_enums import TransferTypeEnum
 
     # Sempo Level 0 LIMITS (payment only)
     assert create_credit_transfer.check_sender_transfer_limits() == [
         limit for limit in LIMITS
-        if limit.name == 'Sempo Level 0'
-        and str(create_credit_transfer.transfer_type) in limit.applied_to_transfer_types
```
enjeyw: reliance on this for building the comparision list was meant that if the check on the transfer type broke somewhere in the codebase, this list broke as well. In one case, this resulted in an empty list on both sides of the equality ie
`create_credit_transfer.check_sender_transfer_limits() -> assety [] == [] ` 
This means we were passing tests even though none of the limits were being applied 😱

app/server/api/credit_transfer_api.py
```diff
@@ -298,7 +298,7 @@ def post(self, credit_transfer_id):
                         send_user=g.user,
                         receive_user=recipient_user,
                         uuid=uuid,
-                        transfer_subtype='DISBURSEMENT')
+                        transfer_subtype=TransferSubTypeEnum.DISBURSEMENT)
```
tristanhcole: Are you sure that will work @enjeyw ? might need to be `TransferSubTypeEnum.DISBURSEMENT.value`?

app/server/models/credit_transfer.py
```diff
@@ -287,11 +280,11 @@ def __init__(self,
                  recipient_user=None,
                  sender_transfer_account=None,
                  recipient_transfer_account=None,
-                 transfer_type=None,
+                 transfer_type: TransferTypeEnum=None,
```
tristanhcole: Nice, strings were getting passed in making a temperamental difficult sometimes

app/server/utils/transfer_limits.py
```diff
@@ -0,0 +1,251 @@
...
+    #     for transfer_type in limit.applied_to_transfer_types:
+    #         if isinstance(transfer_type, (tuple, list)):
+    #             if str(self.transfer_type) == transfer_type[0]\
+    #                     and str(self.transfer_subtype) == transfer_type[1]:
+    #                 relevant_limits.append(limit)
+    #                 continue
+    #         else:
+    #             if str(self.transfer_type) == transfer_type:
+    #                 relevant_limits.append(limit)
+    #                 continue
```
tristanhcole: ?

app/server/utils/transfer_limits.py
```diff
@@ -0,0 +1,251 @@
...
+    return credit_transfer.token and credit_transfer.token.token_type is token.TokenType.LIQUID
+
+
+def transfer_is_agent_out_subtype(credit_transfer):
+    return credit_transfer.transfer_subtype is TransferSubTypeEnum.AGENT_OUT
+
+# ~~~~~~COMPOSITE CHECKS~~~~~~
+
+
+def is_liquid_and_has_foo(ct):
```
tristanhcole: :thinking_face:

app/server/utils/transfer_limits.py
```diff
@@ -0,0 +1,251 @@
...
+def is_liquid_and_user_is_kyc_business_verified(credit_transfer):
+    if is_user_and_liquid_token(credit_transfer):
+        return False
+
+    return user_business_kyc_is_verified(credit_transfer)
+
+# ~~~~~~LIMIT FILTERS~~~~~~
+
+
+@curry
```
tristanhcole: wat

test/unit/app/models/test_models.py
tristanhcole: nice catch

app/server/utils/transfer_limits.py
tristanhcole: Other than the string comparison in the tests file, no

tristanhcole:
@enjeyw just realized that you've made the limits apply to only `is_liquid_tokens`. Should be all tokens for Sempo limits.

app/server/api/credit_transfer_api.py
enjeyw: I think that's what was causing trouble in the first place - `TransferSubTypeEnum.DISBURSEMENT.value` is a string, so invites string comparison, whereas u can just check equality for most things on `TransferSubTypeEnum.DISBURSEMENT`

app/server/utils/transfer_limits.py
enjeyw: ohh just an alternate implementation that i didn't wanna throw out just yet.

app/server/utils/transfer_limits.py
enjeyw: ohhhh shit that was from when i was talking to u about SLAP during that call lol! Can u remove when u change the implementation of sempo limits?

app/server/utils/transfer_limits.py
enjeyw: https://en.wikipedia.org/wiki/Currying

PR #85, opened 2019-12-21
updating transfer limits to cover ALL tokens

tristanhcole:
* Updating transfer limits to include all tokens (as spec implies https://help.withsempo.com/en/articles/3358285-what-are-sempo-account-limits)
* adding group accounts tests


tristanhcole:
[Update Sempo Account limits to include all tokens](https://trello.com/c/eDje5kWZ/171-update-sempo-account-limits-to-include-all-tokens)

app/server/models/credit_transfer.py
```diff
@@ -340,5 +341,3 @@ def __init__(self,
 
         self.append_organisation_if_required(self.recipient_transfer_account.organisation)
         self.append_organisation_if_required(self.sender_transfer_account.organisation)
-
-        self.check_sender_transfer_limits()
```
tristanhcole: https://github.com/teamsempo/SempoBlockchain/blob/d86628d3085c628ca224824933e9db2ab9dcf617/app/server/models/credit_transfer.py#L344-L345

app/server/utils/ussd/token_processor.py
```diff
@@ -91,8 +91,9 @@ def get_limit(user: User, token: Token) -> Optional[TransferLimit]:
         if len(limits) == 0:
             return None
         else:
+            # todo: fix in token agent branch
             # might want to do something different if there's more than one limit...
-            return limits[0]
+            return limits[-1]
```
tristanhcole: This is a quick fix to always get last limits (GE), which will be fixed in Token Agent branch

test/unit/app/models/test_models.py
```diff
@@ -330,23 +330,34 @@ def test_new_credit_transfer_check_sender_transfer_limits(create_credit_transfer
...
 
-    # Check GE LIMITS for Liquid Token (withdrawal)
+    # Check additional GE LIMITS for Liquid Token (withdrawal)
     create_credit_transfer.token.token_type = token.TokenType.LIQUID
     create_credit_transfer.transfer_type = TransferTypeEnum.WITHDRAWAL
     create_credit_transfer.sender_transfer_account.balance = 10000
     assert create_credit_transfer.check_sender_transfer_limits() == [
         limit for limit in LIMITS
-        if 'GE Liquid Token - Standard User' == limit.name
+        if 'GE Liquid Token - Standard User' == limit.name or 'Sempo Level 3: WD' in limit.name
```
tristanhcole: TODOs:
- refactor test_models.py into file for each model
- refactor test_new_credit_transfer_check_sender_transfer_limits into sperate tests

app/server/utils/transfer_limits.py
```diff
@@ -82,52 +87,48 @@ def token_is_liquid_type(credit_transfer):
...
+
 def transfer_is_agent_out_subtype(credit_transfer):
     return credit_transfer.transfer_subtype is TransferSubTypeEnum.AGENT_OUT
 
 # ~~~~~~COMPOSITE CHECKS~~~~~~
 
 
-def is_liquid_and_has_foo(ct):
-    return token_is_liquid_type(ct) and ct.is_foo==True
+def is_user_and_any_token(credit_transfer):
```
enjeyw: why can't we just use 'is_user' directly? feels like "is any token" would always be true?

tristanhcole: we want to apply Sempo Limits to any reserve and liquid token transactions. Might be weird cases were no token is defined, seemed reasonable to me

PR #88, opened 2020-01-03
Celery Upgrades

enjeyw:
Celery currently gets a bit caught due to blocked queues. This PR is improving celery flows. Contains:
- using celery native uuids for task tracking instead of waiting for postgres to giveus an id. This means that we don't need to a worker to accept a task before we can log track it.

- Provisioning for retrying failed blockchain tasks from the start, including recalculating nonces etc (both automatically and manually) - this means that if a task fails due to a bad current state on the blockchain (for example insufficient balance), or bad nonce, we can easily retry it at a later point


app/server/api/blockchain_transaction_api.py
```diff
@@ -85,76 +85,17 @@ def post(self):
 
         call = post_data.get('call')
```
enjeyw: this entire file is due for a review, because we don't use most of it anymore. Will do when i look into making sure new funds added to orgs are detected, since that's the functionality that overlaps with the old code here.

app/server/utils/ge_migrations/poa_explorer.py
```diff
@@ -0,0 +1,85 @@
+# -*- coding: utf-8 -*-
+
```
enjeyw: this file is just moved. No need to review

app/server/utils/ge_migrations/rds_migrate.py
```diff
@@ -0,0 +1,385 @@
+from flask import current_app
```
enjeyw: also moved - don't review

app/server/models/organisation.py
```diff
@@ -48,7 +48,8 @@ class Organisation(ModelBase):
     @property
     def queried_org_level_transfer_account(self):
         if self.org_level_transfer_account_id:
-            return server.models.transfer_account.TransferAccount.query.get(self.org_level_transfer_account_id)
+            return server.models.transfer_account.TransferAccount\
+                .query.execution_options(show_all=True).get(self.org_level_transfer_account_id)
```
tristanhcole: `execution_options(show_all=True)` would this be needed? Wouldn't an _org_level_transfer_account_ be tied to an organisation?

app/server/utils/blockchain_tasks.py
```diff
@@ -178,7 +190,7 @@ def make_token_transfer(self, signing_address, token,
         :param token: ERC20 token being transferred
         :param from_address: address of wallet sending token
         :param to_address:
-        :param amount: the NON WEI amount being sent, eg 2.3 Dai
+        :param amount: the CENTS amount being sent, eg 2300 Cents = 2.3 Dollars
         :param dependent_on_tasks: list of task IDs that must complete before txn will attempt
         :return: task id for the transfer
```
tristanhcole: shouldn't it be `:return: task uuid for the transfer`?

see 
https://github.com/teamsempo/SempoBlockchain/pull/88/files#diff-1930434c0329c8a9ba0046d6053c39a2R269

app/server/utils/user.py
```diff
@@ -511,6 +511,15 @@ def proccess_create_or_modify_user_request(
         }
         return response_object, 400
 
+    business_usage = None
+    if business_usage_id:
```
tristanhcole: `business_usage_id` was changed to `business_usage` on line 175. Looks like an error

Also `business_usage = None` is unnecessary as default None in `create_transfer_account_user` params.

eth_worker/eth_manager/processor.py
```diff
@@ -200,7 +251,8 @@ def check_transaction_hash(self, tx_hash):
         tx_receipt = self.w3.eth.getTransactionReceipt(tx_hash)
 
         def print_and_return(return_dict):
```
tristanhcole: `print_and_return` only returns 😒

quick_setup_script.py
```diff
@@ -257,31 +257,31 @@ def __init__(self, api_host='http://0.0.0.0:9000/api/v1/', email=None, password=
...
     #     api_token=
-    #     'eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJleHAiOjE1NzcxNjc3MDgsImlhdCI6MTU3NjU2MjkwOCwiaWQiOjEsInJvbGVzIjp7IkFETUlOIjoic2VtcG9hZG1pbiJ9fQ.S1dMa668pSLkbNAYzgWOC6Td-E84xgDxEg5ilinKiGI|eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJleHAiOjE1NzYyMTEzMDIsImlhdCI6MTU3NjEyNDg3MiwiaWQiOjZ9.zkEUtwMgOSrcLy68Rtv_JeMCj9HrsOyQUfH3Dc3itYE'
+    #     'eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJleHAiOjE1NzcxODU3NzQsImlhdCI6MTU3NjU4MDk3NCwiaWQiOjEsInJvbGVzIjp7IkFETUlOIjoic2VtcG9hZG1pbiJ9fQ.CGrSHCUcpJlGq3sBgim7omwTBZAx-v0N3AAUurlkqhI|eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJleHAiOjI0Mzk2MDQxMjYsImlhdCI6MTU3NTY5MDQ5NiwiaWQiOjJ9.WaSdLvU5aGxLmNo5uZV0_PmV7LOTymeBBxOymy0Er7U'
     # )
-    # #
+
+    s = Setup(
+        api_host='http://0.0.0.0:9000/api/v1/',
+        api_token=
+        'eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJleHAiOjE1Nzg2MjE3MzIsImlhdCI6MTU3ODAxNjkzMiwiaWQiOjEsInJvbGVzIjp7IkFETUlOIjoic2VtcG9hZG1pbiJ9fQ.-Luhouipdrtw2mSFTSzlUR3CB5FWPGMLyZP2k6cv2b4|eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJleHAiOjE1NzYyMTEzMDIsImlhdCI6MTU3NjEyNDg3MiwiaWQiOjZ9.zkEUtwMgOSrcLy68Rtv_JeMCj9HrsOyQUfH3Dc3itYE'
```
tristanhcole: I feel like these should be in local config or something instead of committing to github? even though it is dev...

tristanhcole:
a few changes and conflicts need to be resolved manually

quick_setup_script.py
enjeyw: because they change all the time (every 24 hours) and they're not really great to stick in local config. We could throw them into a second .gitignored file.

app/server/utils/user.py
tristanhcole: @enjeyw updated comment to make it clearer what I was saying ehre

eth_worker/eth_manager/processor.py
enjeyw: u saw the comment below though yeah? Just commented it out for now incase we want it later

app/server/models/organisation.py
enjeyw: it is, but in some cases, to organisation is a different one, so won't appear unless u explicity ask to see all

quick_setup_script.py
tristanhcole: ok

app/server/utils/user.py
enjeyw: line 175 is a different function... this function has `business_usage_id` defined higher up in it?*
 and the None is required because otherwise u'd be passing in an undefined param into `create_transfer_account_user`

*That being said the code above where it's defined is kinda weird... i think business usage creation  needs a review at some point.

Also I have exactly ZERO recollection of writing these lines lol

PR #91, opened 2020-01-05
auth fixes

tristanhcole:


app/client/components/messageBar.jsx
```diff
@@ -26,19 +24,13 @@ class MessageBar extends React.Component {
...
-              <PageWrapper style={{transitionProperty: 'all', transitionDuration: '1.5s', transitionTimingFunction: 'cubic-bezier(0, 1, 0.5, 1)', top: (message.showMessage ? 0 : -35), position: 'fixed', zIndex: '99', textAlign: 'center'}}>
-                  <Message style={{backgroundColor: (message.error ? '#F44336' : '#2D9EA0'), opacity: (message.showMessage ? 1 : 0)}}>{message.messageList[0]}</Message>
-              </PageWrapper>
-          )
-      } else {
-          return (
-              null
-          )
-      }
+    return (
```
tristanhcole: message bar wasn't showing on auth screens

app/client/sagas/authSagas.js
```diff
@@ -67,6 +67,7 @@ function* watchSaveOrgId() {
   yield takeEvery(UPDATE_ACTIVE_ORG, saveOrgId);
 }
 export function* logout() {
+    yield call(removeTFAToken);
```
tristanhcole: tfa token wasn't removed on logout

app/client/sagas/authSagas.js
```diff
@@ -197,10 +206,27 @@ function* watchRegisterRequest() {
 function* activate({activation_token}) {
   try {
     const activated_account = yield call(activateAPI, activation_token);
-    yield put({type: ACTIVATE_SUCCESS, activated_account});
-    yield put(createLoginSuccessObject(activated_account));
-    yield call(storeSessionToken, activated_account.auth_token );
-  } catch (error) {
+
```
tristanhcole: Added TFA handling to the activate flow.

This means that a user who receives an email with login link, will automatically go to TFA screen. An alternative was to call login flow api, but that felt sub par as would be an extra api call to the flow (which could be slow on poor connections)

app/server/api/auth_api.py
```diff
@@ -76,7 +76,7 @@ def post(self):
         email_ok = False
 
         whitelisted_emails = EmailWhitelist.query\
-            .filter_by(email=email, referral_code=referral_code, used=False) \
```
tristanhcole: This was causing partial email matchers (@gmail.com) to not show in the list

app/server/api/auth_api.py
```diff
@@ -156,18 +156,14 @@ def post(self):
...
             user.update_last_seen_ts()
 
-            response_object = {
-                'status': 'success',
-                'message': 'Successfully activated.',
-                'auth_token': auth_token.decode(),
-                'user_id': user.id,
-                'email': user.email,
-            }
+            response_object = create_user_response_object(user, auth_token, 'Successfully activated.')
```
tristanhcole: Not sure why this wasn't a proper `create_user_response_object`. User was able to login, but with errors, causing an auto-logout as soon as they clicked something

app/server/api/auth_api.py
```diff
@@ -156,18 +156,14 @@ def post(self):
             if tfa_response_oject:
                 tfa_response_oject['auth_token'] = auth_token.decode()
 
+                db.session.commit()  # need this here to commit a created user to the db
```
tristanhcole: user wasn't being commited to db

app/server/api/auth_api.py
```diff
@@ -229,6 +225,23 @@ def post(self):
...
+            if tfa_response_oject:
+                tfa_response_oject['auth_token'] = auth_token.decode()
+
+                db.session.commit()  # need to commit here so that is_activated = True
+
+                return make_response(jsonify(tfa_response_oject)), 401
+
+            # Update the last_seen TS for this user
+            user.update_last_seen_ts()
+
```
tristanhcole: adding tfa flow to `/activate`

app/client/sagas/authSagas.js
enjeyw: Not removing TFA token on logout was by design - if the user has selected to remember the computer, it makes sense that they'd want this even after logging out (this is how AWS/Gmail etc behave)

tristanhcole: Hm, that makes sense. 

It was causing issues when I then tried to login with another user account though?

enjeyw: mmmm yeah good point. Maybe we should save tfa tokens as a dictionary indexed by user id's....

app/server/api/auth_api.py
enjeyw: nice catch

app/server/api/auth_api.py
enjeyw: don't change it now, but I'm wondering if steps like this should actually return 200 codes given the user is going down the correct path.

tristanhcole:
> lgtm

cool will merge once tests pass

app/server/api/auth_api.py
tristanhcole: yeah maybe, need to think about it

app/client/sagas/authSagas.js
tristanhcole: so should this be updated so that a TFA token isn't removed (and just leave the edgecase bug in for the time being)?

app/server/api/auth_api.py
tristanhcole: @enjeyw you're happy with this logic being added here?

see: https://github.com/teamsempo/SempoBlockchain/pull/91#discussion_r363072911

app/server/api/auth_api.py
tristanhcole: the auth flow, specifically tfa flow, for `/register`, `/login` and `/activate` feels a bit DRY

PR #92, opened 2020-01-06
Africastalking Sender Names

enjeyw:
Allows for a non-default sender ID when using Africastalking for SMS


app/server/utils/phone.py
```diff
@@ -83,4 +85,18 @@ def send_messagebird_message(self, to_phone, message):
...
+            # First try with custom sender Id
+            resp = self.africastalking_client.send(
+                message,
+                [to_phone],
+                sender_id=current_app.config.get('AT_SENDER_ID', None)
+            )
+
+            # If that fails, fallback to no sender ID
+            if resp['SMSMessageData']['Message'] == 'InvalidSenderId':
+                server.sentry.captureMessage("InvalidSenderId {}".format(current_app.config.get('AT_SENDER_ID', None)))
```
tristanhcole: nice, I need to use sentry more :)

config.py
```diff
@@ -212,6 +212,7 @@ def get_database_uri(name, host, censored=True):
 
 AT_USERNAME = specific_parser["AFRICASTALKING"]["username"]
 AT_API_KEY = specific_parser["AFRICASTALKING"]["api_key"]
+AT_SENDER_ID = specific_parser["AFRICASTALKING"].get("at_sender_id")
```
tristanhcole: @enjeyw No config file change?

enjeyw: config file change had already snuck in lol

tristanhcole: haha no worries then

tristanhcole: was wondering why tests didn't fail on that

app/server/utils/phone.py
enjeyw: yeah i think we need to make it a deeper part of our processs - like making sure we tick off bugs that get raised etc.

config.py
enjeyw: i mean they shouldn't fail anyways because it's a `get`

PR #97, opened 2020-01-09
admin disbursement approval

tristanhcole:
*adding default initial config
*changing user role invalid to a 403 (so user isn't logged out)
*changing PUT credit transfer role to superadmin


app/server/utils/auth.py
```diff
@@ -153,7 +153,7 @@ def wrapper(*args, **kwargs):
                             response_object = {
                                 'message': 'user does not have any of the allowed roles: ' + str(allowed_roles),
                             }
-                            return make_response(jsonify(response_object)), 401
+                            return make_response(jsonify(response_object)), 403
```
tristanhcole: User's where being logged out if they did a forbidden action.

Seem's more appropriate as a [`403 ERROR`](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#403)

app/server/utils/user.py
```diff
@@ -509,15 +509,6 @@ def proccess_create_or_modify_user_request(
...
 
-    if additional_initial_disbursement and not additional_initial_disbursement <= current_app.config[
-        'MAXIMUM_CUSTOM_INITIAL_DISBURSEMENT']:
-        response_object = {
-            'message': 'Disbursement more than maximum allowed amount ({} {})'
-                .format(current_app.config['MAXIMUM_CUSTOM_INITIAL_DISBURSEMENT'] / 100,
-                        current_app.config['CURRENCY_NAME'])
-        }
-        return response_object, 400
-
```
tristanhcole: disbursements allowed greater than `MAXIMUM_CUSTOM_INITIAL_DISBURSEMENT`, but will not be automatically resolved as complete

tristanhcole:
[Disbursement for existing user](https://trello.com/c/gZfSNWv4/182-disbursement-for-existing-user)

app/client/components/creditTransfer/creditTransferList.jsx
```diff
@@ -181,64 +181,48 @@ class CreditTransferList extends React.Component {
...
-          this.props.users.byId[item.recipient_user].last_name
-        )
-      } else if (item.transfer_type === 'RECLAMATION' && item.authorising_user_email) {
-        item.recipient = <div style={{color: '#96DADC'}}> {item.authorising_user_email} </div>
-      } else {
-        item.recipient = '-'
-      }
-      return (item)
-    })
-    // .map(id => this.props.creditTransfers.byId[id])
```
tristanhcole: recipient users use custom accessor- this code is unecessary now

tristanhcole: `_customRecipient` and `_customSender`

app/client/components/pages/creditTransferListPage.jsx
```diff
@@ -29,7 +29,7 @@ const mapDispatchToProps = (dispatch) => {
 
 class CreditTransferListPage extends React.Component {
   componentDidMount() {
-      this.props.loadCreditTransferList({per_page:50, page:1})
+      this.props.loadCreditTransferList()
```
tristanhcole: load whole credit transfer list - if this takes to long @enjeyw, we could load first 1000, then async load the rest (showing a loading bar top left or something).

tristanhcole: redux is built to update pretty seamlessly

app/server/api/credit_transfer_api.py
```diff
@@ -118,7 +118,7 @@ def get(self, credit_transfer_id):
 
             return make_response(jsonify(response_object)), 201
 
-    @requires_auth(allowed_roles={'ADMIN': 'admin'})
+    @requires_auth(allowed_roles={'ADMIN': 'superadmin'})
```
tristanhcole: only superadmin's can complete/reject credit transfer's now

app/server/utils/auth.py
enjeyw: ohh nice! definitely agree

app/client/components/pages/creditTransferListPage.jsx
enjeyw: what's wrong with paginating this right now?

app/client/components/user/CreateUserForm.tsx
```diff
@@ -65,7 +65,11 @@ const validate = (values: ICreateUser) => {
 
 class CreateUserForm extends React.Component<InjectedFormProps<ICreateUser, Props> & Props> {
   componentDidMount() {
-    this.props.initialize({ accountType: TransferAccountTypes.USER.toLowerCase(), gender: 'female' });
+    this.props.initialize({
+      accountType: TransferAccountTypes.USER.toLowerCase(),
+      gender: 'female',
+      additionalInitialDisbursement: window.DEFAULT_INITIAL_DISBURSEMENT / 100 || undefined
```
enjeyw: I think it would makes sense to deprecate the `starting_balance` config option if we're going down this path. (I do think we should go down this path, it makes sense, though the word "additional" should probably be stripped for the naming convention of the varaibles)

app/client/sagas/creditTransferSagas.js
```diff
@@ -46,8 +46,7 @@ function* updateStateFromCreditTransfer(result) {
     yield put({type: UPDATE_USER_LIST, users});
   }
 
-  if (result.message === 'Transfer Successful') {
-    //TODO: More string comparisons!??!?!
+  if (result.is_create === true) {
```
enjeyw: YAAAAYYY DEATH TO STRINGS

app/server/utils/user.py
```diff
@@ -589,7 +580,8 @@ def proccess_create_or_modify_user_request(
             additional_initial_disbursement,
             organisation.token,
             receive_user=user,
-            transfer_subtype=TransferSubTypeEnum.DISBURSEMENT)
+            transfer_subtype=TransferSubTypeEnum.DISBURSEMENT,
+            automatically_resolve_complete=(additional_initial_disbursement <= current_app.config['MAXIMUM_CUSTOM_INITIAL_DISBURSEMENT']))
```
enjeyw: It's possible that
```
additional_initial_disbursement != current_app.config['MAXIMUM_CUSTOM_INITIAL_DISBURSEMENT']
```
makes more sense. AKA you need special permission to do anything other than the default.

Actually I think what u have is fine, but throwing this out there anyways lol

app/client/components/pages/creditTransferListPage.jsx
tristanhcole: because then search doesn't work

app/client/components/user/CreateUserForm.tsx
tristanhcole: oh, didn't even see that as a function.

tristanhcole: which one should we deprecate?

tristanhcole:
Suggested Disbursement Process;
- one config INITIAL_DISBURSEMENT_AMOUNT

On user creation (show default amount), anything above that amount needs approval thus is PENDING.

If admin role is subadmin, user defaults to unapproved, thus CreditTransfer is PENDING. Admins must approve users (resolving default transfers to RESOLVED, non-default stay PENDING).
If admin role is admin, user defaults to approved, thus CreditTransfer is RESOLVED, unless above default amount. Superadmins must approve non-default amounts.
If admin role is superadmin, user defaults to approved, thus CreditTransfer is RESOLVED.

This should all be handled in `make_initial_disbursement` and `approve_and_disburse` functions on the TransferAccount object. `make_initial_disbursement` can be called from user util `proccess_create_or_modify_user_request`. 

@enjeyw thoughts?

enjeyw:
> Suggested Disbursement Process;
> 
> Configs: maximum_custom_initial_disbursement and default_initial_disbursement. Deprecate starting_balance?
> 
> On user creation (show default amount), anything above that amount needs approval thus is PENDING.
> 
> * If admin role is subadmin, user defaults to unapproved, thus CreditTransfer is PENDING. Admins must approve users (resolving default transfers to RESOLVED, non-default stay PENDING).
> * If admin role is admin, user defaults to approved, thus CreditTransfer is RESOLVED, unless above default amount. Superadmins must approve non-default amounts.
> * If admin role is superadmin, user defaults to approved, thus CreditTransfer is RESOLVED.
> 
> This should all be handled in `make_initial_disbursement` and `approve_and_disburse` functions on the TransferAccount object. `make_initial_disbursement` can be called from user util `proccess_create_or_modify_user_request`.
> 
> @enjeyw thoughts?

Sounds right to me!

tristanhcole:
> > Suggested Disbursement Process;
> > Configs: maximum_custom_initial_disbursement and default_initial_disbursement. Deprecate starting_balance?
> > On user creation (show default amount), anything above that amount needs approval thus is PENDING.
> > 
> > * If admin role is subadmin, user defaults to unapproved, thus CreditTransfer is PENDING. Admins must approve users (resolving default transfers to RESOLVED, non-default stay PENDING).
> > * If admin role is admin, user defaults to approved, thus CreditTransfer is RESOLVED, unless above default amount. Superadmins must approve non-default amounts.
> > * If admin role is superadmin, user defaults to approved, thus CreditTransfer is RESOLVED.
> > 
> > This should all be handled in `make_initial_disbursement` and `approve_and_disburse` functions on the TransferAccount object. `make_initial_disbursement` can be called from user util `proccess_create_or_modify_user_request`.
> > @enjeyw thoughts?
> 
> Sounds right to me!

Cool I'll try get an update in today

app/server/models/transfer_account.py
```diff
@@ -194,19 +194,38 @@ def get_approval(self, receiving_address):
...
 
-            if self.is_beneficiary:
-                disbursement = self.make_initial_disbursement()
+        auto_approve = current_app.config['AUTO_APPROVE_TRANSFER_ACCOUNTS']
+        admin = getattr(g, 'user', None)
+
+        if not self.is_approved and admin:
+            if AccessControl.has_sufficient_tier(admin.roles, 'ADMIN', 'superadmin'):
+                auto_resolve = True
+                self.is_approved = True
```
enjeyw: is there a reason this isn't set to `self.is_approved = auto_approve`? Feels weird to have 3 different behaviours?

app/server/models/transfer_account.py
```diff
@@ -194,19 +194,38 @@ def get_approval(self, receiving_address):
...
+
+            elif AccessControl.has_sufficient_tier(admin.roles, 'ADMIN', 'admin'):
+                auto_resolve = initial_disbursement \
+                               <= current_app.config['MAXIMUM_CUSTOM_INITIAL_DISBURSEMENT']
```
enjeyw: spec in comment thread has this as:
> Superadmins must approve non-default amounts.

rather than amounts more than `MAXIMUM_CUSTOM_INITIAL_DISBURSEMENT`, which I think makes sense (or equally, amounts more than default amount, which is also implied from spec)... wondering if we should just deprecate `MAXIMUM_CUSTOM_INITIAL_DISBURSEMENT`

app/server/models/transfer_account.py
```diff
@@ -194,19 +194,38 @@ def get_approval(self, receiving_address):
...
+                self.is_approved = auto_approve
+
+            elif AccessControl.has_sufficient_tier(admin.roles, 'ADMIN', 'subadmin'):
+                auto_resolve = False
+                self.is_approved = False
+
+        if self.is_beneficiary:
+            if len(self.credit_receives) < 1:
+                # approve_and_disburse might be called for a second time to disburse
+                # so first check that no credit transfer have already been received
```
enjeyw: when does this occur?

app/server/models/transfer_account.py
```diff
@@ -194,19 +194,38 @@ def get_approval(self, receiving_address):
...
 
-    def approve_and_disburse(self):
-        if not self.is_approved:
-            self.is_approved = True
+    def approve_and_disburse(self, initial_disbursement=None, auto_resolve=False):
+        from server.utils.access_control import AccessControl
 
-            if self.is_beneficiary:
-                disbursement = self.make_initial_disbursement()
+        auto_approve = current_app.config['AUTO_APPROVE_TRANSFER_ACCOUNTS']
```
enjeyw: isn't this now deprecated in favour of the tiering system below? though i guess doesn't hurt to leave in, and just have it set to true

app/server/models/transfer_account.py
```diff
@@ -194,19 +194,38 @@ def get_approval(self, receiving_address):
...
+
+        if self.is_beneficiary:
+            if len(self.credit_receives) < 1:
+                # approve_and_disburse might be called for a second time to disburse
+                # so first check that no credit transfer have already been received
+                disbursement = self._make_initial_disbursement(initial_disbursement, auto_resolve)
                 return disbursement
 
-    def make_initial_disbursement(self, initial_balance=None):
+    def _make_initial_disbursement(self, initial_disbursement, auto_resolve=False):
```
enjeyw: If it's possible for admins to have `current_app.config['AUTO_APPROVE_TRANSFER_ACCOUNTS'] = False`, 
we might consider that initial disbursements only resolve once the user is approved. Otherwise what could happen is:
1. admin creates a user in unapproved state
2. disbursement goes through automatically because it's pending
3. superadmin elects not to approve user
4. funds are now locked with unapproved user

not a super big deal though

app/server/models/transfer_account.py
enjeyw: ohh when admin calls it.

I think it makes more sense to seperate out these steps more, because right now:
- a superadmin can't approve a user but reject a non-default amount.
- a subadmin can call this function from the 'approve' section and it'll return success, when really it has rejected them
- if a user who's yet to be activated receives credits from another party, this will incorrectly assume they've received a disbursement,

Something like the following probably makes more sense IMO:
1. Have a function that contains most of this code that fires when a user is first created -
2. Have a second approval function that admins and superadmins can call, that approves (and probably also resolves the initial disbursement)
3. leave non default amounts to be manually approved

app/server/models/transfer_account.py
tristanhcole: Sure, we can deprecate it

app/server/models/transfer_account.py
tristanhcole: > a subadmin can call this function from the 'approve' section and it'll return success, when really it has rejected them
Subadmins can't approve users... ? `@requires_auth(allowed_roles={'ADMIN': 'admin'})`

> if a user who's yet to be activated receives credits from another party, this will incorrectly assume they've received a disbursement,
That is true

> Have a second approval function that admins and superadmins can call, that approves (and probably also resolves the initial disbursement)
I don't think we need a second approval function? I agree there could be better handling in how we resolve initial disbursement

app/server/models/transfer_account.py
tristanhcole: maybe we should have `AUTO_APPROVE_TRANSFER_ACCOUNTS` apply to both enrollers and admins, none at all, or everyone. What do you think?

tristanhcole: I can't see enrollers wanting to be auto approved though...

app/server/models/transfer_account.py
tristanhcole: True -- what do you think we should do with the tiering system
https://github.com/teamsempo/SempoBlockchain/pull/97#discussion_r366329496

app/server/models/transfer_account.py
enjeyw: I think it's probably simplest to go for none at all. The logic you explained in the original documentation is very easy to reason with:
- admins and superadmins can always add users without them being approved by a third party, subadmins never can
- only superadmins can disburse non-default amounts without it being approved.

app/server/models/transfer_account.py
enjeyw: > I don't think we need a second approval function? I agree there could be better handling in how we resolve initial disbursement

"second approval function" was bad wording on my behalf. more of "A second function in addition to the one mention above, that is specifically designed as an approval one". Definitely open to other implementions, but keen to obey Single Responsibility Prinicple

app/server/models/transfer_account.py
tristanhcole: ok, will remove then

app/server/models/transfer_account.py
tristanhcole: but would `approve` resolve an initial disbursement or is that done separately? If a super admin resolves an initial disbursement as COMPLETE, but the user is unapproved, is that a big deal?

tristanhcole: How would you know when to fire the second function  that only approves?

app/server/models/transfer_account.py
tristanhcole: done

app/server/models/transfer_account.py
tristanhcole: done

app/server/models/transfer_account.py
enjeyw: > If a super admin resolves an initial disbursement as COMPLETE, but the user is unapproved, is that a big deal?

I don't think so.

> but would approve resolve an initial disbursement or is that done separately? 

I can see this going one of three ways:

- always resolve initial disbursement (if user has sufficient authority). If you want to reject an initial disbursement you need to do it before approving a user. (This is pretty much what you've implemented above)
Pros: For superadmins, consistent behaviour regardless of whether the disbursement is default or not.
Cons: Easy to accidentally approve a bad disbursement. If an admin is approving a user with non-default amount, then the amount won't be resolved, so inconsistent behaviour there.

- never resolve an initial disbursement
Pros: Consistent behaviour with no surprises. Won't ever result in superadmin approving a non default amount when they didn't mean to.
Cons: Potentially doubles the required steps required to activate a user

- only resolve initial disbursement when it's default
Pros: Admins and superadmins both experience the same behaviour. Good balance between ease of activating users, and control over disbursement amounts.
Cons: Kinda hard to predict behaviour from point of view of approver if u don't remember whether amount is default or not.

enjeyw: No strong opinion over which one is best. In the  absence of any datapoint, it'd probably just choose the path that's easiest

tristanhcole: > only resolve initial disbursement when it's default
I think this makes the most sense

app/server/models/transfer_account.py
```diff
@@ -197,30 +197,28 @@ def get_approval(self, receiving_address):
...
-
-            elif AccessControl.has_sufficient_tier(admin.roles, 'ADMIN', 'admin'):
-                auto_resolve = initial_disbursement \
-                               <= current_app.config['MAXIMUM_CUSTOM_INITIAL_DISBURSEMENT']
-                self.is_approved = auto_approve
-
-            elif AccessControl.has_sufficient_tier(admin.roles, 'ADMIN', 'subadmin'):
-                auto_resolve = False
-                self.is_approved = False
+        if not not self.is_approved and admin and AccessControl.has_sufficient_tier(admin.roles, 'ADMIN', 'admin'):
```
enjeyw: shouldn't this be just `if not`? surprised a test didn't catch this given it should probably render approving impossible

app/server/models/transfer_account.py
```diff
@@ -197,30 +197,28 @@ def get_approval(self, receiving_address):
...
 
         if self.is_beneficiary:
+            # approve_and_disburse might be called for a second time to disburse
+            # so first check that no credit transfer have already been received
             if len(self.credit_receives) < 1:
-                # approve_and_disburse might be called for a second time to disburse
-                # so first check that no credit transfer have already been received
+                # make initial disbursement
                 disbursement = self._make_initial_disbursement(initial_disbursement, auto_resolve)
                 return disbursement
```
enjeyw: are you sure this works if a user has a transfer from another party? (see previous comment)

tristanhcole: how would they get a transfer from another party?

tristanhcole: approve and disburse is called on initial user creation if not self sign up

PR #102, opened 2020-01-12
Locking strat

enjeyw:
Postgresql based nonce calculation was far too prone to race condition weirdness. Hoping a Redis based lock will behave better


app/server/utils/transfer_limits.py
```diff
@@ -200,7 +207,8 @@ def __init__(self,
 
         self.application_filter = application_filter
         self.time_period_days = time_period_days
-        self.total_amount = total_amount
+        # TODO: Make LIMIT_EXCHANGE_RATE configurable per org
+        self.total_amount = int(total_amount * config.LIMIT_EXCHANGE_RATE) if total_amount else None
```
enjeyw: `config.LIMIT_EXCHANGE_RATE` is only settable at a system level RN - should probably have this configurable per org once we do org level settings

enjeyw:
[Migrate community tokens](https://trello.com/c/I4olfBnb/5-migrate-community-tokens)

tristanhcole:
@enjeyw you'll need to resolve conflicts manually 🤕

PR #104, opened 2020-01-14
Ussd bugfixes

enjeyw:
small ussd bugfixes


app/server/utils/ussd/token_processor.py
```diff
@@ -221,7 +221,7 @@ def standard_string(t):
             token_exchanges = "\n".join(map(standard_string, token_info_list))
 
         # TODO: Don't love the string comparison here, but it keeps thing short
-        if token_exchanges == "\n":
+        if token_exchanges in ["\n", '']:
```
enjeyw: wow. much fragile.

PR #105, opened 2020-01-14
Credit Transfer Fixes

tristanhcole:
* fixing reclamation
* fixing set balance
* small bug fixes to `newTransferManager`

**MERGE: admin-disbursement-approval first**


tristanhcole:
[balance and reclamation aren't working](https://trello.com/c/5mRiqqqw/215-balance-and-reclamation-arent-working)

tristanhcole:
MERGE: admin-disbursement-approval first!!

app/client/components/creditTransfer/creditTransferList.jsx
```diff
@@ -181,64 +181,48 @@ class CreditTransferList extends React.Component {
...
-        )
-      } else if (item.transfer_type === 'RECLAMATION' && item.authorising_user_email) {
-        item.recipient = <div style={{color: '#96DADC'}}> {item.authorising_user_email} </div>
-      } else {
-        item.recipient = '-'
-      }
-      return (item)
-    })
-    // .map(id => this.props.creditTransfers.byId[id])
+    .map(id => this.props.creditTransfers.byId[id])
```
enjeyw: man this section is soooooo much cleaner

app/client/components/pages/creditTransferListPage.jsx
```diff
@@ -29,7 +29,7 @@ const mapDispatchToProps = (dispatch) => {
 
 class CreditTransferListPage extends React.Component {
   componentDidMount() {
-      this.props.loadCreditTransferList({per_page:50, page:1})
+      this.props.loadCreditTransferList()
```
enjeyw: I thought we were loading in the first 200?

app/server/models/transfer_account.py
```diff
@@ -194,19 +194,38 @@ def get_approval(self, receiving_address):
...
         return None
 
-    def approve_and_disburse(self):
-        if not self.is_approved:
-            self.is_approved = True
+    def approve_and_disburse(self, initial_disbursement=None, auto_resolve=False):
+        from server.utils.access_control import AccessControl
 
-            if self.is_beneficiary:
-                disbursement = self.make_initial_disbursement()
```
enjeyw: skipping this review section cos it's actually the other pr...

app/client/components/pages/creditTransferListPage.jsx
tristanhcole: yeah 200 or 500 is good, will update

app/server/api/credit_transfer_api.py
```diff
@@ -326,7 +326,7 @@ def post(self, credit_transfer_id):
                     response_list.append({'status': 400, 'message': str(e)})
 
                 else:
-                    db.session.commit()
+                    db.session.flush()
```
enjeyw: this doesn't work because a 400 response code doesn't commit.

As I said, I think we should just explicitly commit at the end of a request when we want to, and not do it when we don't.
Otherwise we:
- have a whole bunch of get requests that commit to the db
- have weird inconsistent cases like this one, where we need to commit before make response
- do twice as many flushes as required.

app/client/components/pages/creditTransferListPage.jsx
enjeyw: this still needs to be changed

tristanhcole:
> Two changes:
> 
> * pagination limit needs to be applied
> * flush needs to be a commit

Done @enjeyw

PR #107, opened 2020-01-15
terms

tristanhcole:
updating terms to point to GE terms (which include ours) until we've fully migrated users over and updated our terms to include community currencies


tristanhcole:
[for the terms - can you link to our site which includes both GE and a link to Sempo? or visa versa?](https://trello.com/c/7vfvQ7BY/209-for-the-terms-can-you-link-to-our-site-which-includes-both-ge-and-a-link-to-sempo-or-visa-versa)

PR #109, opened 2020-01-15
Build eth_worker Simulator

michielderoos:
Add the ability to simulate responses from eth_worker. This is to make testing easier, and cuts down on dependencies when testing and developing the core application (No need for redis or ganache).  It also makes everything run super fast. 

Right now the simulator doesn't do much more than the mocks from blockchain_tasker, but my design goal here was to pipe the tasks to a set of functions so we can dynamically generate agreeable results. With every task linked to a function, we can also make this stateful in the future (if we have to).
  
Note: This is a reasonably big refactor, so I think it would need a good amount of QA (especially since I don't have full context on everything yet)  
Thanks for reviewing! :)


app/server/__init__.py
```diff
@@ -123,8 +123,9 @@ def after_request(response):
 
         for task in g.celery_tasks:
             try:
+                # TODO: Standardize this task (pipe through execute_synchronous_celery)
                 task.delay()
-            except Exception as e:
+            except Exception:
```
michielderoos: lint

app/server/__init__.py
```diff
@@ -123,8 +123,9 @@ def after_request(response):
 
         for task in g.celery_tasks:
             try:
+                # TODO: Standardize this task (pipe through execute_synchronous_celery)
```
michielderoos: Note: I'm not immediately worried about this one, since I don't see anywhere that g.celery_tasks is actually written to

app/server/utils/misc_tasks.py
```diff
@@ -3,6 +3,7 @@
 class MiscTasker(object):
     @staticmethod
     def set_ip_location(ip_address_id, ip):
+        # TODO: Standardize this task (pipe through execute_synchronous_celery)
```
michielderoos: Just limiting scope of this PR to blockchain_tasks for now. Would like to follow this up with another PR for other worker tasks if that's fine

app/server/__init__.py
enjeyw: yeah I put this in here for tasks that we only wanted to execute AFTER the server db had committed, to avoid a race condition where the task might request an object by its ID, only to find the object doesn't exist yet, but then at some point in a refactor found ourselves not actually needing it. Could come in useful in the future though, if we start returning more data from workers to the main server

app/server/utils/blockchain_tasks_simulator.py
```diff
@@ -17,7 +17,7 @@ def deploy_contract(kwargs, args):
     return FakeCeleryAsyncResult()
 
 def call_contract_function(kwargs, args):
-    return FakeCeleryAsyncResult(result=18)
+    return FakeCeleryAsyncResult(result=int(1e19))
```
enjeyw: Sorry explained this poorly in my last review.

The issue here is that `get_wallet_balance` (and similar things like `get_allowance`) needs to return a very different sized number (~1e19) to `get_token_decimals` (~18), so they can't use an identical `call_contract_function` response.

michielderoos: Oh that makes a lot more sense! Fixed and ready for re-👀

app/server/utils/blockchain_tasks_simulator.py
```diff
@@ -17,7 +17,24 @@ def deploy_contract(kwargs, args):
...
+        function_name = kwargs['function']
+        if function_name in function_results:
+            result = function_results[function_name]
+        else:
+            print('[WARN] Contract function \'${function_name}\' does not have a default simulator result. Using default value')
+            result = function_results['default']
+    else:
+        print('[WARN] Contract function called without \'function\' argument provided in kwargs.')
+        result = function_results['default']
+    return FakeCeleryAsyncResult(result=result)
```
enjeyw: boomshakalaka

PR #111, opened 2020-01-16
Adding configurable blockexplorer links

enjeyw:
The old system could only point to etherscan. Now we can point to xdai on the blockexplorer via config settings


tristanhcole:
sops?

enjeyw:
> sops?

no need for local since blockexplorer doesn't work for ganache anyways

tristanhcole:
[links for wallet ID on xDAI (not EthScan)](https://trello.com/c/guVAgcCH/214-links-for-wallet-id-on-xdai-not-ethscan)

PR #117, opened 2020-01-18
disbursement bug fixes

tristanhcole:


app/server/api/credit_transfer_api.py
```diff
@@ -458,7 +458,7 @@ def post(self):
 )
 
 credit_transfer_blueprint.add_url_rule(
-    '/credit_transfer/<int:credit_transfer_id>',
+    '/credit_transfer/<int:credit_transfer_id>/',
```
tristanhcole: I changed this in [PR 105](https://github.com/teamsempo/SempoBlockchain/pull/105) and don't know why

app/server/models/transfer_account.py
```diff
@@ -198,7 +198,7 @@ def approve_and_disburse(self, initial_disbursement=None, auto_resolve=False):
         from server.utils.access_control import AccessControl
 
         admin = getattr(g, 'user', None)
-        auto_resolve = initial_disbursement != current_app.config['DEFAULT_INITIAL_DISBURSEMENT']
+        auto_resolve = initial_disbursement == current_app.config['DEFAULT_INITIAL_DISBURSEMENT']
```
tristanhcole: not sure how this wasn't picked up... 🤷‍♂

app/server/schemas.py
```diff
@@ -407,7 +407,7 @@ class TransferUsageSchema(Schema):
 
 transfer_accounts_schema = TransferAccountSchema(
     many=True,
-    only=('balance', 'created', 'id', 'users', 'token', 'primary_user_id', 'blockchain_address')
+    only=('balance', 'created', 'id', 'users', 'token', 'primary_user_id', 'blockchain_address', 'is_approved')
```
tristanhcole: `is_approved` wasn't being returned in transfer account schema but was being accessed on that page...

app/server/api/credit_transfer_api.py
tristanhcole: i think it was to make the tests easier to write...

tristanhcole: v stupid in hindsight

enjeyw: I hate how this even matters. Feels like u should be able to just use either.

app/server/models/transfer_account.py
enjeyw: lol!

app/server/api/transfer_account_api.py
```diff
@@ -119,7 +119,7 @@ def put(self, transfer_account_id):
             if not approve == transfer_account.is_approved and transfer_account.is_approved is not True:
                 transfer_account.approve_and_disburse()
 
-            db.session.commit()
+            db.session.flush()
```
enjeyw: i mean i think there's a 90% we'll end up shifting all of these back pretty soon but ok lol

PR #118, opened 2020-01-18
Handling Org Pathing

EDsCODE:
- ensure dashboard is viewing correct organization when specific user account page is copy and pasted
- changing organization should redirect to base path or the next most specific path
- link transfer accounts on beneficiary live feed


EDsCODE:
some other formatted in my environment obliterated the original spacing on utils file

EDsCODE:
I changed the implementation back to the simpler setup where changing organization will reroute you to the dashboard. I realized that simply returning the inner route such as going from /accounts/6 to /accounts isn't consistent across all routes (/users/2 -> /users would cause an error). So I'll leave the implementation for changing organizations simpler for now. 

Direct linking into specific account should be good to go

EDsCODE:
consolidated implementation into HoC that can be applied to any component/page that will need organization information. i've applied the wrapper to /users/:id and /accounts/:id so far as they seem to be the only routes that will especially need to parse organization information upon copy/paste

app/client/components/organizationWrapper.jsx
```diff
@@ -0,0 +1,46 @@
...
+
+    componentDidMount() {
+        //load organisation if param is not matching currently active org
+        let query_params = parseQuery(location.search)
+        if(query_params["org"] && this.props.login.organisationId != query_params["org"]){
+            this.props.updateActiveOrgRequest(query_params["org_name"], query_params["org"])
+        } else {
+            browserHistory.push({
+                search: generateQueryString({org_name: this.props.login.organisationName})
+            })
```
tristanhcole: This isn't working for me on the TransferAccountList Page...

Something like this works `browserHistory.push(location.pathname + generateQueryString({org_name: this.props.login.organisationName}))`

Screen Shot 2020-01-21 at 7 01 22 pm

tristanhcole: The HOC is clean though!

app/client/components/pages/singleUserPage.jsx
```diff
@@ -95,7 +96,7 @@ class SingleUserPage extends React.Component {
   }
 }
 
-export default connect(mapStateToProps, mapDispatchToProps)(SingleUserPage);
+export default connect(mapStateToProps, mapDispatchToProps)(organizationWrapper(SingleUserPage));
```
tristanhcole: It could potentially make more sense for the HOC to be used in the `nav.jsx` file, so we can easily see which pages it's applied to?

app/client/components/organizationWrapper.jsx
```diff
@@ -0,0 +1,46 @@
...
+        }
+        
+    }
+
+    render() {
+        return <Component {...this.props}></Component>
+    }
```
tristanhcole: haha I use two spaces... we definitely need to setup a linter and some style rules so there's not constant diffs

app/client/utils.js
```diff
@@ -1,168 +1,203 @@
...
-  let orgId = getOrgId();
-  var response_string = query_string;
-  if (orgId !== null && typeof orgId !== "undefined") {
-    response_string = query_string + `org=${orgId}&`;
-  }
+    let orgId = getOrgId();
+    var response_string = query_string;
+    if (orgId !== null && typeof orgId !== "undefined") {
+        response_string = query_string + `org=${orgId}&`;
+    }
```
tristanhcole: defs need a linter

EDsCODE: haha..there were moments were I had to find a way to save without formatting because my VSCode setup has a few packages that default to format on save. I made a ticket for me (or anyone else) to visit!

https://trello.com/c/rFzwJZHg/257-setup-linter-to-standardize-code-styling

app/client/components/pages/singleUserPage.jsx
EDsCODE: yeah! I originally wanted to do this and I just tried it again for a sanity check but what happens is that the nav file uses lazy loading for the components that would need to be wrapped. the wrapper expects the component immediately and so i'm not too sure how to incorporate the HoC in the nav file without removing lazy loading

app/client/components/organizationWrapper.jsx
EDsCODE: hmmm. did you apply the wrapper to the /accounts component? I just added the wrapper to more of the routes including the transfer account list and it should work now for that page.

app/client/components/pages/singleUserPage.jsx
tristanhcole: yeah, I'm not sure. I think it's because the lazy loading doesn't immediately render a component, so  the HOC throws an error

app/client/utils.js
tristanhcole: awesome!

app/client/components/organizationWrapper.jsx
tristanhcole: ok nice... just found another small issue. The HOC is reloading the URL page, so when you click back, it goes from `/accounts/126?org_name=Grassroots%20Economics&org=2` to `/accounts/126`. Not sure how to fix this off the top of my head, but might need to rethink how the HOC is structured...

EDsCODE: good catch! just fixed it. had to use history.replace instead of history.push because push was adding another path into the history that users were able to traverse through when pressing back

app/client/components/pages/singleUserPage.jsx
EDsCODE: do you think this is blocking? I can look into a workaround before merging this PR if so

more broadly, I was going to bring up a possible discussion about reformatting the routes because I think in this hierarchy organizations should be the inner most path (ex: /org/1/accounts, /org/2/transfers) because the accounts and transfers are per org

tristanhcole: no, I think what you've got so far is fine 😄 

Yes, definitely open to reformatting the routes. We originally looked at doing that, and I think it would be a better way that as a query string in the long run

app/client/components/organizationWrapper.jsx
tristanhcole: nice

app/client/components/organizationWrapper.jsx
```diff
@@ -0,0 +1,46 @@
...
+import { parseQuery, generateQueryString } from './../utils'
+import { withRouter } from "react-router";
+
+const OrganizationWrapper = Component => class extends Component {
+
+    componentDidMount() {
+        //load organisation if param is not matching currently active org
+        let query_params = parseQuery(location.search)
+        if(query_params["org"] && this.props.login.organisationId != query_params["org"]){
+            this.props.updateActiveOrgRequest(query_params["org_name"], query_params["org"])
```
tristanhcole: Why do we have `org_name` in the query string? I think it should just be pulled from `this.props.login.organisationName` so that there is a single source of truth. It would also simplify the query string url alot...

EDsCODE: changed! i initially added both params because I didn't realize the login object has an array of all the loaded organizations

app/client/components/pages/singleUserPage.jsx
EDsCODE: https://trello.com/c/15EDvhDn/258-reformat-dashboard-routes-to-better-reflect-hierarchy

PR #119, opened 2020-01-19
refactor web api

tristanhcole:
* refactoring web api into `apiClient` to follow DRY
* removing deprecated QR code flow


tristanhcole:
[Refactor web API](https://trello.com/c/AJBoE83s/136-refactor-web-api)

app/client/api/apiClient.js
```diff
@@ -0,0 +1,49 @@
...
+    method: method,
+  };
+
+  //todo: check which headers are needed for given method
+  if (isForm) {
+    formData = new FormData();
+    Object.keys(body).map(key => {
+      formData.append(key.toString(), body[key]);
+    });
+    method !== 'GET' ? request['body'] = formData : null;
```
EDsCODE: if we're anticipating a lot more differences in request parameters between GETs, POSTs, etc and adding other request types such as PATCH we might want to have the client return an object of methods and handling corresponding elements inside each function so that, for example, we don't need to process formdata if the request is going to be a GET. 

`return {
     get: () => ....,
     post: () => ....
}`

similar to your thoughts above. But, this is an incredibly minor tweak that can be handled when we start adding more request types

app/client/components/pages/dashboardPage.jsx
```diff
@@ -33,8 +33,7 @@ const mapStateToProps = (state) => {
 const mapDispatchToProps = (dispatch) => {
   return {
     logout:       () => dispatch(logout()),
-    loadCreditTransferList: (query, path) => dispatch(loadCreditTransferList({query, path})),
-    activateAccount: (activation_token) => dispatch(activateAccount(activation_token))
+    loadCreditTransferList: (query, path) => dispatch(loadCreditTransferList({query, path}))
```
EDsCODE: this api request payload can get standardized too once more typescript conversions are performed

PR #121, opened 2020-01-19
Better USSD Messages

enjeyw:
- Users receive a proper message when they try to make a transfer but don't have enough balance
(https://trello.com/c/QgoiqKEO)
- Dynamically modify the number of directory responses shown so that we don't hit the Safaricom character limit and have out messages trimmed.
https://trello.com/c/x0g82wAA


app/migrations/versions/5f4b4c9f586d_.py
```diff
@@ -0,0 +1,92 @@
+"""empty message
```
tristanhcole: this seems weird?

app/migrations/versions/5f4b4c9f586d_.py
```diff
@@ -0,0 +1,92 @@
...
+    sa.Column('user_id', sa.Integer(), nullable=True),
+    sa.ForeignKeyConstraint(['user_id'], ['user.id'], ),
+    sa.ForeignKeyConstraint(['ussd_menu_id'], ['ussd_menu.id'], ),
+    sa.PrimaryKeyConstraint('id')
+    )
+    op.create_index(op.f('ix_ussd_session_session_id'), 'ussd_session', ['session_id'], unique=True)
+    op.drop_index('ix_ussd_sessions_session_id', table_name='ussd_sessions')
+    op.drop_table('ussd_sessions')
+    op.drop_index('ix_ussd_menus_name', table_name='ussd_menus')
+    op.drop_table('ussd_menus')
```
tristanhcole: updating table naming, thus the big migration file

tristanhcole:
@michielderoos @EDsCODE you'll need to run `python manage.py update_data` once you merge in this PR. rebinding ussd session table

PR #122, opened 2020-01-19
Transfer Account List Performance Fixes

enjeyw:
The current Transfer Account List page takes WAAAAAYYYY too long to load (up to a minute when there's 1000s of accounts). Long term fix is server side search etc.
Short term fix :
- Speeding up subsequent load times after the initial by only loading modified accounts 
- Technical improvements to the react rendering process namely:
-      reducing number of re-renderings
-     don't include ALL element a dict used for working out which rows were selected, but rather just assume non-present elements are not selected.

### DIFF
--- a/app/client/components/transferAccount/transferAccountList.jsx
+++ b/app/client/components/transferAccount/transferAccountList.jsx
@@ -172,10 +172,6 @@ class TransferAccountList extends React.Component {
 
 
   render() {
-    console.log('TA Dict', this.state.idSelectedStatus);
-
-    console.log('TA IDS', this.get_selected_ids_array(this.state.idSelectedStatus));
-
     const {account_type} = this.state;
     const loadingStatus = this.props.transferAccounts.loadStatus.isRequesting;
     let accountTypes = Object.keys(TransferAccountTypes);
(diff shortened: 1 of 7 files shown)


tristanhcole:
[Accounts ... slow](https://trello.com/c/ESRY9Mx6/221-accounts-slow)

app/client/components/transferAccount/transferAccountList.jsx
```diff
@@ -55,89 +55,76 @@ class TransferAccountList extends React.Component {
...
-    }
   }
 
-  get_selected_ids_array(selected) {
-    Object.filter = (obj, predicate) => Object.keys(obj).filter( key => predicate(obj[key]) ).reduce( (res, key) => (res[key] = obj[key], res), {} );
-
-    return Object.keys(Object.filter(selected, selected => selected === true));
+  get_selected_ids_array(statusDict) {
+    return Object.keys(statusDict).filter(id => statusDict[id] === true);
   }
```
tristanhcole: wayyy simpler

app/client/components/transferAccount/transferAccountList.jsx
```diff
@@ -185,6 +172,10 @@ class TransferAccountList extends React.Component {
 
 
   render() {
+    console.log('TA Dict', this.state.idSelectedStatus);
+
+    console.log('TA IDS', this.get_selected_ids_array(this.state.idSelectedStatus));
```
tristanhcole: remove @enjeyw

PR #123, opened 2020-01-20
user properties

tristanhcole:
* user editable custom attributes
* ability to update business category
* filter fix


tristanhcole:
[changing user properties on dev.sempo](https://trello.com/c/KY6UjnQJ/244-changing-user-properties-on-devsempo)

PR #125, opened 2020-01-20
Add pepper to hashes

michielderoos:
Adds pepper to password hashes.  
My approach here is to encrypt our existing hashes with a pepper.   
This has a couple advantages over just doing `bcrypt(hashpw+password, salt)`
First, migration becomes easy. We just encrypt all our existing passwords with pepper.
Second, if our pepper ever becomes compromised, decrypting and migrating to a new pepper becomes trivial.   
  
NOTE: Before merging (or passing circleci...), the config.inis I don't think I have access to need a pepper


michielderoos:
[Add Pepper to Password hashes so that pins can't be trivially brute-forced](https://trello.com/c/X8u3WsDl/218-add-pepper-to-password-hashes-so-that-pins-cant-be-trivially-brute-forced)

PR #129, opened 2020-01-22
Make sure currency types are consistent across dashboard

EDsCODE:
https://trello.com/c/UlhbiYYo/132-bug-fix-transfer-account-page-multi-currencies

- the default currency symbol that is used will be of the currenctly active organisation in the dashboard. This currency symbol will fill in the metrics on the dashboard
- pipe through the credit transfer token type so that symbols can be directly accessed rather than depending on the sender_account


EDsCODE:
I piped through each organisation's standard currency symbol and so the default currency that will be displayed.

related note: i think the organisation objects need to be consolidated into an object for clarity's sake so that there aren't 3 separate variables (organisationid, organisationname, organisationtoken) that need to be maintained. i didn't do that here because i think it'll fit into a broader state management refactor and so i don't heavily impact how api response objects are formed atm

app/client/components/creditTransfer/creditTransferList.jsx
```diff
@@ -124,6 +108,29 @@ class CreditTransferList extends React.Component {
...
+
+    creditTransferIds.map(i => {
+
+      // retrieve any sender account that has not been retrieved
+      let sender_account_id = this.props.creditTransfers.byId[i].sender_transfer_account_id
+      if(!this.props.transferAccounts.byId[sender_account_id]){
+        if(this.props.creditTransfers.byId[i].transfer_type !== "DISBURSEMENT"){
+          this.props.loadTransferAccountList(sender_account_id)
+        }
+      }
```
EDsCODE: upon component load, this will check if any sender accounts have not been retrieved. I excluded disbursement credit transfers because it loads a blank vendor account

EDsCODE:
after further experimenting i discovered that this solution will currently break if the accounts are nulled out as described here: https://trello.com/c/A2SeFDbE/240-investigate-transfer-account-cross-contamination-bug

the sender account doesn't exist in the credit transfer

EDsCODE:
will be more cognisant of the minor spacing and line additions before committing in the future. i realized some of the testing i did has left random spacing shifts that should not be in the PR/commits.

app/client/components/dashboard/beneficiaryLiveFeed.jsx
```diff
@@ -106,7 +103,7 @@ class BeneficiaryLiveFeed extends React.Component {
                                     <UserSVG src="/static/media/exchange.svg"/>
                                     <UserGroup>
                                       <ClickableTopText onClick={() => this.navigateToAccount(transferAccountId)}>{sender_user_name} exchanged</ClickableTopText>
-                                      <BottomText><Highlight>{transferFromMoney}</Highlight> for
+                                      <BottomText><DarkHighlight>{transferFromMoney}</DarkHighlight> for
```
EDsCODE: minor style change on live feed entries where the currency amount had inconsistent highlight

app/server/schemas.py
```diff
@@ -153,6 +153,8 @@ def filter_rejected(self, data, many):
 
     transfer_metadata       = fields.Function(lambda obj: obj.transfer_metadata)
 
+    token                   = fields.Nested(TokenSchema, only=('id', 'symbol'))
```
EDsCODE: this might cause tests to fail depending on what checks are on the schemas

tristanhcole:
I think it makes sense to completely deprecate `window.CURRENCY_NAME` and `window.CURRENCY_DECIMALS` and make it just part of the organisation object.

Need to remove from `utils.js` and `index.html`
https://github.com/teamsempo/SempoBlockchain/blob/53564c4fa0f5d51fb6c5e2cb8d5247994c0f4626/app/client/utils.js#L9-L12

app/server/schemas.py
tristanhcole: hmmm @enjeyw how would this work for credit transfers from one token to another and exchanges?

tristanhcole:
> I piped through each organisation's standard currency symbol
> 
> related note: i think the organisation objects need to be consolidated into an object for clarity's sake so that there aren't 3 separate variables (organisationid, organisationname, organisationtoken) that need to be maintained. i didn't do that here because i think it'll fit into a broader state management refactor and so i don't heavily impact how api response objects are formed atm

Nice, that's exactly how I'd do it.

Yeah, there is an organisation reducer but it isn't being used currently. Adding to the login object was just a quick fix, should be refactored into its own reducer.

app/server/utils/auth.py
```diff
@@ -293,7 +293,8 @@ def get_user_organisations(user):
     organisations = dict(
         active_organisation_name=active_organisation.name,
         active_organisation_id=active_organisation.id,
-        organisations=[dict(id=org.id, name=org.name) for org in user.organisations]
+        active_organisation_token=active_organisation.token.symbol,
+        organisations=[dict(id=org.id, name=org.name, token=org.token.symbol) for org in user.organisations]
```
tristanhcole: fine until we refactor... see https://github.com/teamsempo/SempoBlockchain/pull/129#issuecomment-578474802

app/server/schemas.py
enjeyw: Exchanges are necessarily made of two credit transfers, one for each token type, so this _should_ be fine I think

EDsCODE:
[Bug: Fix Transfer Account Page Multi-currencies](https://trello.com/c/UlhbiYYo/132-bug-fix-transfer-account-page-multi-currencies)

app/server/utils/auth.py
```diff
@@ -345,6 +346,7 @@ def create_user_response_object(user, auth_token, message):
         'server_time': int(time.time() * 1000),
         'ecdsa_public': current_app.config['ECDSA_PUBLIC'],
         'pusher_key': current_app.config['PUSHER_KEY'],
+        # todo: (used on mobile) Deprecate. Currency should be based on active organization/TA account token
         'currency_decimals': current_app.config['CURRENCY_DECIMALS'],
         'currency_name': currency_name,
         'currency_conversion_rate': conversion_rate,
```
tristanhcole: @enjeyw 
1. can we add a "display" currency decimal to token object?
2. whats the best way to return exchange rate for mobile?

PR #131, opened 2020-01-25
Condense the initial ussd flow 

EDsCODE:
language selection, pin entry, pin confirmation are connected and pipe you directly start menu

- connected first 3 phases of onboarding
- removed sms for language selection and pin confirmation (as noted by the ticket to reduce messaging)

https://trello.com/c/GrfuAsho/245-registration-process-multiple-smss


EDsCODE:
the test fails because it expects the initial_language_selection state to result in complete state but if we want to make this change to condense the initial flow initial_language_selection will have to lead to initial_pin_entry

tristanhcole:
> the test fails because it expects the initial_language_selection state to result in complete state but if we want to make this change to condense the initial flow initial_language_selection will have to lead to initial_pin_entry

Do you want me to look at this?

EDsCODE:
> > the test fails because it expects the initial_language_selection state to result in complete state but if we want to make this change to condense the initial flow initial_language_selection will have to lead to initial_pin_entry
> 
> Do you want me to look at this?

not yet. going to look into tests further

test/unit/app/utils/ussd/kenya/test_misc_state.py
```diff
@@ -88,6 +88,19 @@ def test_kenya_state_machine(test_client, init_database, user_factory, session_f
 @pytest.mark.parametrize("session_factory, user_input, language", [
     (initial_language_selection_state, "1", "en"),
     (initial_language_selection_state, "2", "sw"),
+])
```
EDsCODE: i updated the test to reflect the new flow that was proposed in the trello task

EDsCODE:
> > > the test fails because it expects the initial_language_selection state to result in complete state but if we want to make this change to condense the initial flow initial_language_selection will have to lead to initial_pin_entry
> > 
> > 
> > Do you want me to look at this?
> 
> not yet. going to look into tests further

@tristanhcole  check it out now!

EDsCODE:
[Registration process - multiple sms's](https://trello.com/c/GrfuAsho/245-registration-process-multiple-smss)

PR #133, opened 2020-01-25
Directory listing fix

enjeyw:
Fixing Directory Listings - bad wording and the listing wasn't actually returning any results


app/migrations/seed.py
```diff
@@ -275,27 +277,29 @@ def create_business_categories():
 
     print_section_title('Creating Business Categories')
     business_categories = [
-        {'name': 'food_water', 'icon': 'message', 'translations': {
+        {'name': 'Food/Water', 'icon': 'message', 'translations': {
```
enjeyw: The 'name' field appears on the admin dashboard, so making sure it looks a bit nicer

PR #134, opened 2020-01-25
fixing other business category

tristanhcole:
* fixing other biz category
* removing location from transfer account (only on user)


tristanhcole:
[Using Other for biz type](https://trello.com/c/u1TIIwRq/250-using-other-for-biz-type)

PR #136, opened 2020-01-25
adding support for kobo to submit location

tristanhcole:
Kobo can now submit gps location data as name: 'geolocation' type: gps point


tristanhcole:
[Improve Map Functionality](https://trello.com/c/waznab2T/92-improve-map-functionality)

app/server/api/user_api.py
```diff
@@ -125,7 +126,10 @@ def get(self, user_id):
     def post(self, user_id):
 
         post_data = request.get_json()
-        organisation = g.active_organisation
+        organisation = g.get('active_organisation')
+        if organisation is None:
+            # TODO: Refactor.. Query string needed for kobo
+            organisation = Organisation.query.get(request.args.get('org'))
```
tristanhcole: This is a short term fix

app/server/utils/user.py
```diff
@@ -251,7 +256,6 @@ def create_transfer_account_user(first_name=None, last_name=None, preferred_lang
         )
 
         transfer_account.name = transfer_account_name
-        transfer_account.location = location
```
tristanhcole: creating new users with location has been broken for a while it seems

app/server/api/user_api.py
michielderoos: Will a user ever not have an `active_organisation?

tristanhcole: no, but for external auth credentials organisation would need to be pulled from query string (insecure), until your PR is merged in https://github.com/teamsempo/SempoBlockchain/pull/140

tristanhcole: could just merge in your PR first ...

tristanhcole:
Merge in PR #140 first

PR #140, opened 2020-01-26
Set basic external auth per organization

michielderoos:
- Sets creds for existing orgs  
- Sets creds for new orgs when built  
- Changes API to show new creds instead of ones from config file  
- Nukes old creds    
- Also encrypt new creds in the db  

Note: We're keeping the endpoints called Kobo for now, changing is [ticketed separately] . (https://trello.com/c/6XH2Sdx1/274-rename-kobo-endpoints)


michielderoos:
[Change external basic auth to be generated on a per-organization basis](https://trello.com/c/MKruXWXG/271-change-external-basic-auth-to-be-generated-on-a-per-organization-basis)

test/helpers/factories.py
```diff
@@ -65,7 +65,7 @@ class OrganisationFactory(SQLAlchemyModelFactory):
     class Meta:
         model = Organisation
         sqlalchemy_session = db.session
-
+    name = 'Francines Company'
```
enjeyw: `produces = 'Furballs'`

michielderoos: 🐈

PR #142, opened 2020-01-28
Ensure that high priority celery tasks don't get blocked by long running low priority tasks

michielderoos:
Like the sim PR, I think this one will need a close 👀 because if it fails, it'll fail pretty badly. Much simpler than that one though!


tristanhcole:
[Ensure that high priority celery tasks don't get blocked by long running low priority tasks](https://trello.com/c/WylSuvGu/256-ensure-that-high-priority-celery-tasks-dont-get-blocked-by-long-running-low-priority-tasks)

eth_worker/eth_manager/__init__.py
```diff
@@ -57,7 +57,7 @@
 celery_app.conf.beat_schedule = {
     "maintain_eth_balances": {
         "task": utils.eth_endpoint('topup_wallets'),
-        "schedule": 600.0
```
michielderoos: Good debug time, but should probably change this back before merging?

eth_worker/_docker_worker_script.sh
```diff
@@ -20,7 +21,7 @@ else
   if [ "$ret" -ne 0 ]; then
     exit $ret
   fi
-  celery -A eth_manager worker --loglevel=INFO --concurrency=10 --pool=eventlet -Q=celery,high-priority
+  celery -A eth_manager worker --loglevel=INFO --concurrency=10 --pool=eventlet -Q=low-priority,celery,high-priority
```
michielderoos: @enjeyw Was the blocking you mentioned happening with this flag set, or before?

eth_worker/eth_manager/__init__.py
enjeyw: hmmmm i'm not sure. Topup wallets is less critical than it used to be, because all tasks automatically check to see if they have enough balance before transacting. So then the main utility of this is that it keeps txn time low by ensuring a task doesn't need to wait for topup before running. In this regard, even just once a day would probably be sufficient... maybe..

eth_worker/_docker_worker_script.sh
enjeyw: oops - i think this was a mistaken change - it supposed to be default and high-priority only! (otherwise low-priority clogs the default queue)

eth_worker/eth_manager/processor.py
```diff
@@ -462,7 +463,7 @@ def new_transaction_attempt(self, task):
 
         else:
             signature(utils.eth_endpoint('_attempt_transaction'),
-                      args=(task.uuid,)).apply_async(
+                      args=(task.uuid,)).apply(
```
michielderoos: `apply_async`->`apply`   
`apply_async` enqueues the job, where `apply` runs them in the local thread. Before this, there was a possible deadlock state where many workers are all waiting for tasks at the bottom of the queue.

awsDockerrunTemplate.json
```diff
@@ -3,11 +3,11 @@
   "containerDefinitions": [
     {
       "name": "app",
-      "image": "REPOSITORY_URI:server_TAG_SUFFIX",
+      "image": "REPOSITORY_URI:serverTAG_SUFFIX",
```
michielderoos, line 6: Could not get this working with the underscore before TAG_SUFFIX (even when trying to deploy master)

eth_worker/eth_manager/__init__.py
```diff
@@ -65,7 +71,18 @@
 
 red = redis.Redis.from_url(config.REDIS_URL)
 
-persistence_interface = SQLPersistenceInterface(w3=w3, red=red)
+engine = create_engine(
+    config.ETH_DATABASE_URI,
+    poolclass=NullPool,
```
michielderoos, line 76: image

awsDockerrunTemplate.json
enjeyw: weird... they're making it into the aws registry fine. This works to tho

app/server/utils/blockchain_tasks.py
```diff
@@ -37,9 +35,7 @@ def _execute_synchronous_celery(
             async_result.forget()
         return response
 
-    def _synchronous_call(
-            self, contract_address, contract_type, func, args=None, signing_address=None, queue='high-priority'
-    ):
+    def _synchronous_call(self, contract_address, contract_type, func, args=None, signing_address=None, queue='high-priority'):
```
enjeyw, line 38: we definitely need a python style enforcer lol. Because my IDE yells at me for the new version!

build_docker.sh
```diff
@@ -10,6 +10,7 @@ cd ../
 
 docker build -t server . -f ./app/Dockerfile --build-arg GIT_HASH=$GIT_HASH
 docker build -t proxy ./proxy
+docker build -t eth_processor . -f ./eth_worker/Dockerfile --build-arg CONTAINER_TYPE=PROCESSOR
 docker build -t low_priority_eth_worker . -f ./eth_worker/Dockerfile --build-arg CONTAINER_TYPE=LOW_PRIORITY_WORKER
 docker build -t high_priority_eth_worker . -f ./eth_worker/Dockerfile --build-arg CONTAINER_TYPE=HIGH_PRIORITY_WORKER
```
enjeyw, line 15: why do these all need seperate builds? Can't we just inject arguments at runtime?

eth_worker/_docker_worker_script.sh
```diff
@@ -14,13 +13,18 @@ elif [ "$CONTAINER_TYPE" == 'FILTER' ]; then
   python ethereum_filter_test.py
 elif [ "$CONTAINER_TYPE" == 'PROCESSOR' ]; then
   echo "Starting Processor Worker"
-  celery -A eth_manager worker --loglevel=INFO --concurrency=$WORKER_CONCURRENCY --pool=eventlet -Q=processor
+  celery -A eth_manager worker --loglevel=DEBUG --concurrency=$WORKER_CONCURRENCY --pool=eventlet -Q=processor --without-gossip --without-mingle
```
enjeyw, line 16: is this due to this stuff here?
https://www.cloudamqp.com/blog/2016-01-25-identify-and-protect-against-high-cpu-and-memory-usage.html

eth_worker/eth_manager/processor.py
enjeyw: how's that possible? Doesn't `new_transaction_attempt` just throw the task on the queue and then release the worker?

eth_worker/eth_manager/utils.py
```diff
@@ -6,22 +6,22 @@
 import config
 
 
-def execute_synchronous_celery(signature, queue='celery'):
-    async_result = signature.apply_async(queue=queue)
+def execute_synchronous_celery(signature):
+    result = signature.apply()
```
enjeyw, line 10: ohhhhhhhh like here?

prepare_aws_deploy.sh
```diff
@@ -1,10 +1,12 @@
 #!/usr/bin/env bash
 
 #bash build_docker.sh
+#REPOSITORY_URI=290492953667.dkr.ecr.ap-southeast-2.amazonaws.com/blockchaindemo
+#TAG_SUFFIX=test
 
 sed "s|REPOSITORY_URI|$REPOSITORY_URI|g; s|TAG_SUFFIX||g" awsDockerrunTemplate.json > Dockerrun.aws.json
 
-eval $(aws ecr get-login --no-include-email --region $ECR_REGION --profile ECR_BUILDER);
+eval $(aws ecr get-login --no-include-email);
```
enjeyw, line 9: we gotta work these bits out lol

eth_worker/eth_manager/processor.py
michielderoos: Meant to comment this [here](https://github.com/teamsempo/SempoBlockchain/pull/142#discussion_r402813813), yeah. Think it makes sense to do internal calls like this anyway, just to make the entire transaction happen around the same time.

eth_worker/eth_manager/utils.py
michielderoos: 💯

michielderoos: Note: I did revert this to the previous. Was getting a bit of funky behaviour and want to limit the scope of this PR

PR #144, opened 2020-01-29
Ussd self service

enjeyw:


app/migrations/versions/2591b8be0eb8_.py
```diff
@@ -0,0 +1,28 @@
...
+# revision identifiers, used by Alembic.
+revision = '2591b8be0eb8'
+down_revision = '5f4b4c9f586d'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('user', sa.Column('is_ussd_self_sign_up', sa.Boolean(), nullable=True))
```
michielderoos: Would like to hear what @enjeyw or @tristanhcole think, but I think it would be good if instead of a `is_ussd_self_sign_up` boolean here, we used an `Enum` (or maybe a `String`) called something like `signup_method`. That way we can scale this solution and track things like field signups, or future mechanisms like mobile app signups, web signups, or anything else.

app/migrations/versions/0d586d339657_.py
```diff
@@ -0,0 +1,28 @@
...
+# revision identifiers, used by Alembic.
+revision = '0d586d339657'
+down_revision = '2591b8be0eb8'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('user', sa.Column('data_sharing_accepted', sa.Boolean(), nullable=True))
```
michielderoos: The database already has a `terms_accepted` boolean, so we won't need a second `data_sharing_accepted` column

app/server/utils/user.py
```diff
@@ -383,7 +435,7 @@ def send_one_time_code(phone, user):
             'Something went wrong. ERROR: {}'.format(e))
 
 
-def proccess_create_or_modify_user_request(
+def process_create_or_modify_user_request(
```
michielderoos: Good catch! 🙌

enjeyw:
getting
>eth_worker_processor_1  | The module eth_worker was not found.
on tests, which i think is caused by (https://github.com/teamsempo/SempoBlockchain/pull/144/files#diff-a9c401b85e36381dc85d81da0528764dR15)

michielderoos:
[Self Service USSD User Creation](https://trello.com/c/eAWTc1vP/21-self-service-ussd-user-creation)

test/unit/app/utils/ussd/kenya/test_pin_state.py
```diff
@@ -62,49 +62,54 @@ def test_kenya_state_machine(test_client, init_database, user_factory, session_f
...
-      "exchange_rate_pin_authorization", 1, 2),
-     (exchange_rate_pin_authorization_state, standard_user, "0000",
-      "complete", 1, 0),
-     (exchange_rate_pin_authorization_state, standard_user, "1212",
-      "exit_pin_blocked", 2, 3),
-     (exchange_rate_pin_authorization_state, standard_user, "0000",
-      "complete", 2, 0),
- ])
+@pytest.mark.parametrize("session_factory, user_factory, user_input, expected, before_failed_pin_attempts, "
+                         "after_failed_pin_attempts",
```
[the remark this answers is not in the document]
PhilipWafula: What do you mean when you say keep the old format, do you mean the indentation ?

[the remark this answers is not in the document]
michielderoos: Yes, exactly!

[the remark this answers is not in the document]
PhilipWafula: In the interest of standardization maybe we should come up with a consistent styling toolkit for the code base, so that any dev can configure their IDE to said styling guide.
Because in this particular instance this was a change PyCharm auto implemented.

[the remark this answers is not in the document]
michielderoos: I totally agree. We should definitely try to settle on some PyLint settings sooner than later.

PhilipWafula:
> getting
> 
> > eth_worker_processor_1  | The module eth_worker was not found.
> 
> on tests, which i think is caused by (https://github.com/teamsempo/SempoBlockchain/pull/144/files#diff-a9c401b85e36381dc85d81da0528764dR15)

I don't follow, did you make a change here I should revert ?

app/migrations/versions/2591b8be0eb8_.py
enjeyw: I think an enum would probably make a lot of sense here, especially since `is_self_sign_up` is already a field.

app/server/utils/user.py
```diff
@@ -283,6 +282,55 @@ def create_transfer_account_user(first_name=None, last_name=None, preferred_lang
     return user
 
 
+def create_user_without_transfer_account(phone):
+    """
+    This method creates a user without a transfer account or blockchain wallet.
+    :param phone: string with user's msisdn
+    :return: User
+    """
+    user = User(phone=phone, signup_method=SignupMethodEnum.USSD_SELF_SIGNUP.value)
```
michielderoos: When trying to signup with the AfricasTalking sandbox, I get the following error generated from this line:  
`TypeError: 'signup_method' is an invalid keyword argument for User`

PhilipWafula: Is the attribute in your user table. 
I had to manually resolve some migration error there because the ```python manage.py db merge heads```  wasn't working.

PhilipWafula: To be precise it was generating a version that didn't exactly resolve my issue.

PhilipWafula:
> This generally looks really good! It should be fixed though--
> When trying to signup with the AfricasTalking sandbox, I get a couple of errors.
> When I try to login with an existing account (and attempting the incorrect PIN twice), the backend raises this exception:
> `Exception: ('Account creation failed. Error: ', Exception('User already has a transfer account attached.',))`
> 
> There's also an error when creating a new account, as noted in another comment in this review

This feels like something  a test should have caught. I shall look at this in a couple of minutes and resolve it.

app/server/utils/user.py
michielderoos: `sign_up_method` is in my table, and the migrations ran without a problem. One thing I did notice after reading your comment and checking the database, is that this line says `signup_method`, instead of `sign_up_method`. I _can_ sign up with that changed (pushed that change as well), but the other bugs remain.

michielderoos: I think you should add a couple more tests, since they didn't detect this error at all either.

app/migrations/versions/326a61cf9fde_.py
```diff
@@ -0,0 +1,32 @@
...
+down_revision = '0d586d339657'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('user', sa.Column('sign_up_method', sa.String(), nullable=True))
+    op.drop_column('user', 'is_ussd_self_sign_up')
+    op.drop_column('user', 'data_sharing_accepted')
```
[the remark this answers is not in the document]
PhilipWafula: I can't seem to reproduce this error, what do you mean log in ?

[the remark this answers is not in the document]
michielderoos: (Commenting for context of other reviewers, Philip and I discussed this over Slack earlier)  
By "log in" I meant when you interact with the USSD from a pre-existing account generated from the admin dashboard

PhilipWafula:
The migration simplification process seems to be failing on docker, let me see if I can trace the process and fix the issue.
Also looking into that Invalid Request response.

app/server/api/organisation_api.py
```diff
@@ -62,6 +62,8 @@ def post(self, organisation_id):
         token_id = post_data.get('token_id')
         deploy_cic = post_data.get('deploy_cic', False)
 
+        initial_disbursement_amount = post_data.get('initial_disbursement_amount')
```
tristanhcole: @PhilipWafula @michielderoos, if we are updating `initial_disbursement_amount` in this PR then we should be doing it across the whole stack and deprecating `DEFAULT_INITIAL_DISBURSEMENT` in the configs.

In my opinion, we don't want to be running dual disbursement amounts. This should not be merged in until resolved.
a) deprecate old config
b) rollback on this branch and split out to a separate PR

app/migrations/versions/6fb8a4a348f0_.py
```diff
@@ -0,0 +1,28 @@
...
+# revision identifiers, used by Alembic.
+revision = '6fb8a4a348f0'
+down_revision = '326a61cf9fde'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('organisation', sa.Column('initial_disbursement_amount', sa.Float(precision=2), nullable=True))
```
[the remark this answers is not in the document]
tristanhcole: https://github.com/teamsempo/SempoBlockchain/pull/144#discussion_r383063238

app/server/api/organisation_api.py
tristanhcole: Let me know what decision y'all decide. I started working on it here also without realising it was part of this PR. https://github.com/teamsempo/SempoBlockchain/pull/175

Can turn that PR into the ability to update from admin dashboard

tristanhcole: I've closed my PR in favor of this. Can we get this merged in soon?

PhilipWafula: @tristanhcole we spoke with @michielderoos and agreed we should split the organization initial disbursement value out into its own PR so we're gonna scale back that feature from this PR and make sure it is on its own.
I think that's a good idea because Will requested that he get auto trickle bonus and auto intial bonus features out first before we get the ussd self service out.
So I have been focused on getting those done.
What I will do is split this initial bonus thing out, then deprecate the config one and get that as a stand alone feature merged.
Then the self service PR can follow once the auto trickle bonus ones are up

PR #145, opened 2020-01-31
Adding eb config

enjeyw:
Makes it easier to create instances from other machines


.gitignore
```diff
@@ -1,7 +1,7 @@
 # Elastic Beanstalk Files
-.elasticbeanstalk/*
```
michielderoos: Should we delete (instead of comment?)

PR #146, opened 2020-02-01
Small org token bug fix

EDsCODE:
there was a bug on local development environment initialization where activating the first account wouldn't work because it tried to access the symbol of a null object 

the fix ensures that org.token is not none before accessing deeper


EDsCODE:
@enjeyw not sure why this happens occasionally but circleCI test currently failed at ussd language selection even though changes are unrelated and tests run locally

enjeyw:
> @enjeyw not sure why this happens occasionally but circleCI test currently failed at ussd language selection even though changes are unrelated and tests run locally

weird - I'll run tests again and we'll see what happens lol

enjeyw:
@EDsCODE *shrugs*

PR #149, opened 2020-02-03
fixing transfer card

tristanhcole:
otherwise sometimes raises error `TypeError: unsupported operand type(s) for +=: 'float' and 'decimal.Decimal'`


app/server/models/transfer_card.py
```diff
@@ -60,6 +60,6 @@ def update_transfer_card(self):
         total_disbursed = 0
 
         for disbursement in disbursements:
-            total_disbursed += disbursement.transfer_amount
+            total_disbursed += float(disbursement.transfer_amount)
```
michielderoos: Should we make total_disbursed a Decimal instead?

enjeyw: it needs to be a somewhat small number, as it gets loaded onto the NFC card that has limited precision. If u look at the setter for `amount_loaded` it actually gets forced into an int by `round`. Probably what would be the most sensible is to use a decimal for `total_disbursed` and then force into an int at the last moment (that way the summing of all the individual disbursements will be as accurate as possible)

enjeyw: plot twist - it's because somehow there's the occasional `_transfer_amount_wei` that is being stored as a float rather than a decimal, which means that total_disbursed tried to add inconsistent types

enjeyw: double plot twist:
it's the new disbursement itself that is getting added as a float to `_transfer_amount_wei` and then once the db commits and expires, it gets re-cast as a decimal.

The best fix would be to ensure that we always pass consistent datatypes around no matter what, which would probably involve:
- at a minimum, forcing transfer_amount_wei to only hold decimals
- deciding what types we use where
- adding typing checks
- 
I could probably do the first one now, but it could have unexpected side effects. So with that said, i think it makes the most sense just to use the quick fix above for now, and do a proper review of amount types down the path

enjeyw: thoughts @michielderoos ?

michielderoos: @enjeyw I think auditing all our money types and ensuring we always cast the same way is a solid idea. In my mind, int > decimal > float for money, but for our purposes decimal is probably always fine. It may also be a good idea to start thinking about using [money](https://github.com/carlospalol/money) or [moneyed](https://github.com/limist/py-moneyed) if we're doing refactoring anyway, but I'm not strongly opinionated about that (just something to consider).

tristanhcole:
need to bump version

app/server/models/transfer_card.py
enjeyw: Hmmm [moneyed](https://github.com/limist/py-moneyed) looks pretty decent. 

for what it's worth, we use decimal in the database because the ethereum contracts tend to have 18 decimals of precision (total of 256 bit unsigned), and Bigint taps out at something like 64 bits, but apart from that they're a huge pain. Also passing around eg 1 dollar as 1000000000000000000 sucks from a dev quality-of-life point of view lol.

Thats why we've traditionally worked in cents on the backend where possible (eg 1 dollar as 100) - it tends to be the most efficient int representation of numbers that we typically encounter on a day-to-day basis. For example, if someone on the mobile app makes a payment of 13.50 USD, we can just pass that in the json payload as 1350 and call it a day.

Then for the edge cases where we need sub-cent precision (for example on cross currency exchanges), we.... ummmmmm.... I don't know actually (eeeek) I think we cast to float? (double eeeek).

Looks like moneyed would solve most of this on the backed - we'd just work in Moneyed types the entire time and then we'd have full precision and dev friendliness. What would we do about json payloads? Just accept that `cents` is the max precision that can be sent over an api call and leave it at that?

tagging @EDsCODE since u might not have seen this thread...

michielderoos: >Just accept that cents is the max precision that can be sent over an api call and leave it at that?

PR #150, opened 2020-02-04
changing pin reset to admin tier

tristanhcole:


tristanhcole:
[Admin should be able to do a pin reset](https://trello.com/c/Yxq7x16x/279-admin-should-be-able-to-do-a-pin-reset)

PR #151, opened 2020-02-04
KYC fixes

tristanhcole:
KYC is kinda a mess due to a bunch of changing business requirements. need to split API to admin and /me for mobile. need to refactor


EDsCODE:
> Unable to test because image upload isn't working. Will return to this later though.

Resolved this by making sure tmp directory exists before attempting to use path in storing document

EDsCODE:
RE Michiel's error with nonumeric postal code: " errors with non-numeric postal codes (I.e. mine, B4E 1G1 gives a serverside error. 1234 was fine though"

Should probably just update schema/model to use string instead of int for that field

tristanhcole:
[KYC Process](https://trello.com/c/Jg6OCu3V/248-kyc-process)

tristanhcole:
@michielderoos or @EDsCODE can you give this a test now? Should send through a slack notification...

app/migrations/versions/463ddd874cb8_.py
```diff
@@ -0,0 +1,34 @@
+"""empty message
+
+Revision ID: 463ddd874cb8
+Revises: fc2eeba8e8e8
+Create Date: 2020-02-17 16:34:47.369834
+
+"""
+from alembic import op
```
tristanhcole: this migration can't downgrade.. not sure how i feel about that

michielderoos: Hard to avoid with type changes like this-- alphanumeric postal codes are pretty common so I think this is worth it.

app/server/models/kyc_application.py
```diff
@@ -50,7 +50,7 @@ class KycApplication(ModelBase):
     street_address_2    = db.Column(db.String)
     city                = db.Column(db.String)
     region              = db.Column(db.String)
-    postal_code         = db.Column(db.Integer)
+    postal_code         = db.Column(db.String)
```
michielderoos: 🙌

PR #153, opened 2020-02-04
new transfer bugfix

tristanhcole:
* fixes new transfers not showing in credit transfer list
* fixes https://sentry.io/organizations/sempo/issues/1494863946/?project=1228848&query=is%3Aunresolved&sort=date&statsPeriod=14d


app/client/components/creditTransfer/creditTransferList.jsx
```diff
@@ -80,17 +80,17 @@ class CreditTransferList extends React.Component {
...
+  componentDidUpdate(prevProps) {
     let { credit_transfer_ids, item_list } = this.props;
 
     // handles credit_transfer_ids array
-    if (credit_transfer_ids !== newProps.credit_transfer_ids) {
+    if (credit_transfer_ids !== prevProps.credit_transfer_ids) {
       this.setState({credit_transfer_ids: {}});
-      // this.sortCreditTransfers(newProps.credit_transfer_ids)
+      this.sortCreditTransfers(credit_transfer_ids)
     }
```
tristanhcole: do we need to optimize credit transfer list like we did on the transfer account @enjeyw ?

enjeyw: we should, but given it only shows the last 250 it's not such an issue

PR #154, opened 2020-02-04
Limit modifications

enjeyw:
Three new modifications to limits:

- KYC applications can be flagged as having multiple id documents, in which case they have higher limits
- Sempoadmins never have limits applied to them
- Reclamations are now included in limits


test/functional/app/api/test_credit_transfer_api.py
```diff
@@ -36,7 +36,13 @@ def test_create_credit_transfer(test_client, authed_sempo_admin_user, create_tra
     if recipient_transfer_accounts_ids:
         recipient_transfer_accounts_ids = [recipient_transfer_accounts_ids]
 
+    # Hack to prevent previous tests from causing future tests to fail
+    # TODO: Change design of entire testing process to enable quick setup of blockchain etc, but not have tests so path dependent
```
enjeyw: Transfers from previous tests are persisted because the db isn't torn down, which results in this crap

test/unit/app/models/test_credit_transfer.py
```diff
@@ -136,6 +144,27 @@ def test_new_credit_transfer_check_sender_transfer_limits_exception_on_init(exte
         c.resolve_as_completed()
         db.session.flush()
 
+# TODO: These tests suck because they impact each other (try doing this one at the end)
```
enjeyw: samesame

PR #155, opened 2020-02-05
SSL fix

tristanhcole:
* swapping all domains from @sempo.ai to @withsempo.com
* this should also include a fix for forced SSL
* added withsempo.com to verified SES domains for region us-west-2
* created a gsuite group account for `admin@withsempo.com` so that we receive replies


tristanhcole:
[QR Code is not working](https://trello.com/c/e3YeE5bP/319-qr-code-is-not-working)

PR #160, opened 2020-02-08
Dashboard chart fix

enjeyw:
Hopefully fixes the bad numbers in the dashboard

### DIFF
--- a/app/client/components/dashboard/metricsBar.jsx
+++ b/app/client/components/dashboard/metricsBar.jsx
@@ -33,28 +33,28 @@ const MetricsBar = ({creditTransferStats, login}) => {
         <ModuleHeader>
           MASTER WALLET BALANCE
         </ModuleHeader>
-        <Metric>{formatMoney(creditTransferStats.master_wallet_balance/100, undefined, undefined, undefined, login.organisationToken)}</Metric>
+        <Metric>{formatMoney(creditTransferStats.master_wallet_balance/100, 0, undefined, undefined, login.organisationToken)}</Metric>
       </MetricWrap>
 
       <MetricWrap>
         <ModuleHeader>
           TOTAL DISTRIBUTED
         </ModuleHeader>
-        <Metric>{formatMoney(creditTransferStats.total_distributed/100, undefined, undefined, undefined, login.organisationToken)}</Metric>
+        <Metric>{formatMoney(creditTransferStats.total_distributed/100, 0, undefined, undefined, login.organisationToken)}</Metric>
       </MetricWrap>
 
       <MetricWrap>
         <ModuleHeader>
           TOTAL SPENT
         </ModuleHeader>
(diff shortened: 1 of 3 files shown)


app/client/components/dashboard/metricsBar.jsx
```diff
@@ -81,6 +81,7 @@ const MetricWrap = styled.div`
 `;
 
 const Metric = styled.p`
+  text-align: center;
```
enjeyw: minor formatting to get the '132222 SARAFU' to centre align like the 'TOTAL DISTRIBUTED' etc labels. We should probably format it so that 'SARAFU' is always on its own line, but that requires a bit more work so meh

app/server/utils/credit_transfer.py
```diff
@@ -43,68 +43,77 @@ def calculate_transfer_stats(total_time_series=False, start_date=None, end_date=
         dateFilter.append(CreditTransfer.created >= start_date)
         dateFilter.append(CreditTransfer.created <= end_date)
 
-    filters = [CreditTransfer.transfer_type == TransferTypeEnum.PAYMENT, CreditTransfer.transfer_subtype == TransferSubTypeEnum.DISBURSEMENT]
+    disbursement_filters = [
```
enjeyw: continual overwriting of the `filters` var (which isn't really a var but more of a const) felt a bit iffy and prone to unexpected errors, and this allows for re-use over multiple queries.

config.py
```diff
@@ -155,6 +155,8 @@
 DATABASE_NAME = specific_parser['DATABASE'].get('database') \
                 or common_parser['DATABASE']['database']
 
+DATABASE_PORT = specific_parser['DATABASE'].get('port') or common_parser['DATABASE']['port']
```
enjeyw: i found myself running multiple localhost postgres servers (different versions) on different ports, so just making the field customisable

enjeyw:
[Fix Dashboard Numbers and graph](https://trello.com/c/Y0HkTh5o/325-fix-dashboard-numbers-and-graph)

app/server/utils/credit_transfer.py
```diff
@@ -77,12 +81,12 @@ def calculate_transfer_stats(total_time_series=False, start_date=None, end_date=
 
     beneficiary_filters = [User.has_beneficiary_role == True]
 
-    total_beneficiaries = db.session.query(User).filter(*beneficiary_filters).filter(*dateFilter).count()
+    total_beneficiaries = db.session.query(User).filter(*beneficiary_filters).filter(*userDateFilter).count()
```
EDsCODE: was using a credittransfer filter on user table which was causing the error @tristanhcole found and noted on slack. 

when a daterange filter is applied it, the usage funnel will filter for newly created users/beneficiaries in the specified period. this may or may not be what we actually want for that section but it will prevent the randomly large number error from appearing

PR #162, opened 2020-02-09
Get Card Balance Fix

enjeyw:
Fixing an issue where cards scanned don't return a balance


enjeyw:
[Scan QR to check card balance error](https://trello.com/c/rTN3uUDT/329-scan-qr-to-check-card-balance-error)

PR #164, opened 2020-02-09
Transfer amount limit message

enjeyw:
Adding error messages when users hit transfer amount limits

Accidently branched off transfer_card fix, so ignore the first commit


enjeyw:
[Limit: Error on sending more than ~5k](https://trello.com/c/c8Jb89QH/301-limit-error-on-sending-more-than-5k)

PR #167, opened 2020-02-11
added simple caching so stats don't continually get recalculated 

EDsCODE:
- uses flask-caching to perform a simple cache on calculate_transfer_stats
- can be extended to cache other repetitive actions

**if we don't want to overload single instance running server we can deploy another redis/memcache to perform the caching


EDsCODE:
[caching everywhere!](https://trello.com/c/mgrQPHB3/331-caching-everywhere)

EDsCODE:
leaving the caching at this for now. will optimize other pages when @michielderoos is done with search and we have a better idea of what the pages are changing into

enjeyw:
> Super cool, I had no idea Flask-Caching was a thing. Would caching potentially return wrong results from `calculate_transfer_stats` if we have multiple organizations though?

Hmmm yeah looks like it _might_ be a problem. Especially because a lot of logic is done in the query itself - means even standard function memoization won't work, because the function isn't pure.

EDsCODE:
i might be mistaken but we could also cache at the api route level so that it caches based on teh "?org=2" parameter

enjeyw:
> i might be mistaken but we could also cache at the api route level so that it caches based on teh "?org=2" parameter

Yeah that'd probably work, though we'd need to be certain that someone couldn't just enter an arbitrary org id and access the cache by bypassing the org membership checks.

Somewhat equivalently, we could pass the org id into `calculate_transfer_stats` so that it's caught in the cache. Then we'd probably want to wrap the whole thing in a class, and make that method private and then provide another public method that checked org membership, to make sure that the cache was never accessed in an unsafe manner.

enjeyw:
> > i might be mistaken but we could also cache at the api route level so that it caches based on teh "?org=2" parameter
> 
> Yeah that'd probably work, though we'd need to be certain that someone couldn't just enter an arbitrary org id and access the cache by bypassing the org membership checks.
> 
> Somewhat equivalently, we could pass the org id into `calculate_transfer_stats` so that it's caught in the cache. Then we'd probably want to wrap the whole thing in a class, and make that method private and then provide another public method that checked org membership, to make sure that the cache was never accessed in an unsafe manner.

@EDsCODE just thought i'd check in on this one!

tristanhcole:
Closing

PR #168, opened 2020-02-11
pin reset message update

tristanhcole:
adding Safaricom and Airtel service codes


tristanhcole:
[pin reset message - service code](https://trello.com/c/Z0eUrhmH/267-pin-reset-message-service-code)

PR #170, opened 2020-02-12
updating view admin tier

tristanhcole:
* live feed only see blockchain addresses (prev. could see names)
* transfer list sender and recipient blockchain address (prev. empty)
* transfer account list now displays as list of blockchain addresses

Bug fixes:
- Cannot read property 'row' of undefined (https://sentry.io/organizations/sempo/issues/1507351531/?project=1228848&query=is%3Aunresolved&sort=priority&statsPeriod=14d)


tristanhcole:
[View only accounts can see Beneficiary live feed on dash board with names](https://trello.com/c/AQ9NZqBN/326-view-only-accounts-can-see-beneficiary-live-feed-on-dash-board-with-names)

app/client/components/pages/transferAccountListPage.jsx
```diff
@@ -41,37 +41,25 @@ class TransferAccountListPage extends React.Component {
...
-
-    if (is_vendor || is_beneficiary) {
-        // just vendors or beneficiaries
-        var transferAccountList = this.props.mergedTransferAccountUserList.filter(transferAccount => {return transferAccount.is_vendor === is_vendor})
-    } else {
-        // all transfer accounts
-        transferAccountList = this.props.mergedTransferAccountUserList
+    let transferAccountList = this.props.mergedTransferAccountUserList;
+
+    if (this.props.login.adminTier === 'view') {
```
tristanhcole: not _super happy_ with string comparison, but we do it all over the codebase for adminTiers so seems OK here.

app/client/components/pages/transferAccountListPage.jsx
```diff
@@ -41,37 +41,25 @@ class TransferAccountListPage extends React.Component {
...
-      } else if (location.pathname.includes(window.BENEFICIARY_TERM_PLURAL.toLowerCase())) {
-          query = {account_type: 'beneficiary'};
-
-      } else {
-          query = {};
-      }
-
-      if (this.props.transferAccounts.loadStatus.lastQueried) {
-        query.updated_after = this.props.transferAccounts.loadStatus.lastQueried;
-      }
```
tristanhcole: `/vendor` and `/beneficiary` are deprecated pathnames no longer used

app/client/components/pages/transferAccountListPage.jsx
michielderoos: I think we could start leaning harder on consts files. I know there's `constants.py`, but that's mostly lists/maps-- I'm taking something literally as simple as `const.VIEW = 'view'`.

app/client/components/navBar.jsx
```diff
@@ -284,6 +284,7 @@ const StandardNavBarHeaderText = styled.p`
   text-decoration: none;
   text-overflow: ellipsis;
   overflow: auto;
+  overflow-x: hidden;
```
tristanhcole: Removes ugly horizontal scrollbar on windows

Screen Shot 2020-02-13 at 2 01 15 pm

app/client/components/pages/transferAccountListPage.jsx
EDsCODE: bigg second on that @michielderoos ^^
i don't mind refactoring that as my next task

tristanhcole: yeah a frontend contestants file would be great

app/client/components/dashboard/analyticsChart.jsx
```diff
@@ -79,7 +79,7 @@ class AnalyticsChart extends React.Component {
 
   render() {
 
-      if (Object.keys(this.props.creditTransferStats).length == 0) {
+      if (Object.keys(this.props.creditTransferStats).length === 3) {
```
enjeyw: would this be safer as `<= 3`?

tristanhcole: good call

PR #172, opened 2020-02-18
Update README.md

michielderoos:


README.md
```diff
@@ -27,7 +27,11 @@ cd app
 ```
 npm install
 ```
-
+    
+### Enable the Simulator (Optional) 
+If you wish to forego installing ganache and redis, you can enable a simulator mode. What this does is bypass the eth_worker and any queued jobs, and instead returns dummy responses to any functions relying on eth_worker. Be warned, this _will_ make your database fall out of sync with any ganache instance you have set up so use this with care, but it is very useful in eliminating dependencies when working on any features in the API or frontend. It also allows you to run `quick_setup_script.py` without additional dependencies.   
+To enable simulator mode, open `/config_files/local_config.ini` and add the line `enable_simulator_mode = true` under the `[APP]` heading.
```
tristanhcole: Can we add `enable_simulator_mode = false` under the `[APP]` heading in SOPS by default?

tristanhcole: Ok, added

PR #175, opened 2020-02-21
default transfer amount

tristanhcole:


tristanhcole:
> Default dispersement amounts are also part of the self signup branch (if this is the same
> idea) https://github.com/teamsempo/SempoBlockchain/pull/144/files#diff-b81908b3c6cb518da625e5ebbbfc4ce7R21

ah right, ok. need to get that merged in. 

I can turn this branch into just the frontend to update default disbursement amounts...

tristanhcole:
[Ability to set organization defaults (i.e. Default Transfer amount)](https://trello.com/c/STfC6Z9e/340-ability-to-set-organization-defaults-ie-default-transfer-amount)

tristanhcole:
Ok going to archive this as being complete in self-sign up

PR #178, opened 2020-02-24
Config Refactor

enjeyw:
Basically, splitting our config files into three parts:
- deployment specific config; things that aren't sensitive, and change on a per deployment basis
- deployment specific secrets; things that ARE sensitive, and change on a per deployment basis
- common secrets; things that ARE sensitive, and can be the same between all deployments
(I also considered adding 'common config': things that aren't sensitive and don't change, but that's pretty much just our consts file, because it's not really config by definition)

Doing means we can now just commit specific config directly to github, which means:
- that SOPS breaks less often
- there's way less configs that a dev without access to sops needs to set up themselves - which means we can just have a script


enjeyw:
TODO:
- get sops running
- get docker tests running

PR #180, opened 2020-02-24
Refactoring single user page to typescript

tristanhcole:
Refactoring single user page to typescript plus...

- changing gender to dropdown
- adding ability to edit business usage category
- show all query string for transfer usages


tristanhcole:
[changing user properties on dev.sempo](https://trello.com/c/KY6UjnQJ/244-changing-user-properties-on-devsempo)

app/server/views/index.py
```diff
@@ -17,59 +17,6 @@ def get_js_bundle_filename():
...
-    return render_template('index.html', js_bundle_main=get_js_bundle_filename())
-
-@index_view.route('/settings/<subroute>')
-def settings_subroute(subroute):
-    return render_template('index.html', js_bundle_main = get_js_bundle_filename())
-
-@index_view.route('/activate-account/')
-def activate_account():
-    return render_template('index.html', js_bundle_main = get_js_bundle_filename())
-
```
enjeyw: Are we sure this works @EDsCODE @tristanhcole? I recall we had only condensed routing back in the day, but we added the above stuff either because of:
a) We did some light server side rendering for some of the pages for SEO reasons (which we don't currently and so if that's the only reason this is here, then it can definitely be removed)
b) React was doing something funky and couldn't find the pages through a catch all during certain actions (like navigating directly to a url, rather than via a link from the index page)

I'm pretty sure it was (a), but just wanna be sure.

tristanhcole: https://github.com/teamsempo/SempoBlockchain/pull/189

Yes, there aren’t any issues with it. We also don’t do any server side rendering anymore so no reason to leave in. See original issue here https://github.com/teamsempo/SempoBlockchain/pull/182#discussion_r390772223

app/client/components/user/EditUserForm.tsx
```diff
@@ -0,0 +1,371 @@
...
+  phone?: string;
+  bio?: string;
+  referredBy?: string;
+  location?: string;
+  businessUsage?: string;
+  usageOtherSpecific?: string;
+  oneTimeCode: number;
+  failedPinAttempts: number;
+  accountType: any[TransferAccountTypes];
+  [key: string]: any;
```
enjeyw: Is there a way we can re-write this to avoid needing to allow basically any key with any value? Feels like it diminished the utility of typescript significantly.

app/client/components/user/SingleUserManagement.tsx
```diff
@@ -0,0 +1,106 @@
...
+  users: any;
+  selectedUser: any;
+  transferUsages: TransferUsage[];
+}
+
+interface OuterProps {
+  userId: number;
+}
+
+type Form = ICreateUserUpdate;
```
enjeyw: `Form` is an identical alias of `ICreateUserUpdate` which is an identical alias of `IEditUser`... can we simplify this a bit?

app/client/components/user/EditUserForm.tsx
```diff
@@ -0,0 +1,371 @@
...
+  publicSerialNumber?: string;
+  phone?: string;
+  bio?: string;
+  referredBy?: string;
+  location?: string;
+  businessUsage?: string;
+  usageOtherSpecific?: string;
+  oneTimeCode: number;
+  failedPinAttempts: number;
+  accountType: any[TransferAccountTypes];
```
enjeyw: is this supposed to just be `accountType: TransferAccountTypes`?
(This would allow `TransferAccountTypes["USER"]` and TransferAccountTypes.USER` as acceptable values)

`any[TransferAccountTypes]` allows "FakeType" as acceptable value

app/client/components/user/EditUserForm.tsx
enjeyw: What's this needed for/is there a way we can refactor this to not require allowing an index signature?

app/client/components/user/SingleUserManagement.tsx
```diff
@@ -0,0 +1,106 @@
+import * as React from "react";
+import { connect } from "react-redux";
+
+import { editUser, resetPin } from "../../reducers/userReducer";
+import { TransferUsage } from "../../reducers/transferUsage/types";
+import { ReduxState } from "../../reducers/rootReducer";
+import EditUserForm, { ICreateUserUpdate } from "./EditUserForm";
+
+interface DispatchProps {
+  editUser: (body: any, path: any) => void;
```
enjeyw: isn't path `string`?

app/client/components/user/SingleUserManagement.tsx
```diff
@@ -0,0 +1,106 @@
...
+  resetPin: (body: any) => void;
```
enjeyw: `body: {user_id: number}` might be better?

app/client/components/user/SingleUserManagement.tsx
enjeyw: also might not hurt to have something like ```
interface IJSONSerialisedUserData {
...
}
```

though i can see how that could be a lot of work

File: app/client/components/user/EditUserForm.tsx [hunk 7]
```diff
@@ -0,0 +1,371 @@
...
+    );
+  }
+}
+
+// TODO: can't figure out the typing here...
+const EditUserFormReduxForm = reduxForm({
+  form: "editUser",
+  validate
+  // @ts-ignore
+})(EditUserForm);
```
enjeyw: Try this?

```
const EditUserFormReduxForm = reduxForm<ICreateUserUpdate, Props>({
  form: "editUser",
  validate
})(EditUserForm);
```

app/client/components/user/SingleUserManagement.tsx
tristanhcole: Updated

app/client/components/user/EditUserForm.tsx
tristanhcole: Fixed I think

app/client/components/user/EditUserForm.tsx
tristanhcole: The reason for `[key: string]: any;` is so that custom_attr_keys can be mapped to the form.

`let custom_attr_keys = customAttributes && Object.keys(customAttributes);
    let attr_dict = {};
    custom_attr_keys.map(key => {
      (attr_dict as attr_dict)[key] = customAttributes[key];
      return attr_dict;
    });`

tristanhcole: We don't know what custom_attributes will be returned. That's why attr_dict and IEditUser need to support.

... although currently the custom_attributes frontend-only supports {key: value} which is 

```
interface attr_dict {
  [key: string]: string;
}
```

tristanhcole: I could update to that?

app/client/components/user/SingleUserManagement.tsx
tristanhcole: mhm I think it's OK as it is

tristanhcole: Once the UserReducer is refactored into Typescript then we can import the `EditUserPayload` from `auth/types.ts`

app/client/components/user/SingleUserManagement.tsx
tristanhcole: good point

app/client/components/user/EditUserForm.tsx
tristanhcole: brilliant thanks!

tristanhcole:
> I haven't got a lot of experience with typescript, so take everything with a huge grain of salt, but a few suggested modifications in comments

Merged in with the latest master. This is good for another review IMO @enjeyw

app/client/components/user/EditUserForm.tsx
enjeyw: mmmmmm doesn't really help much. It could make more sense to have a sub-dict that stores custom attributes, and that can be loosely. Means that the system doesn't need to be clever about whats a custom attribute or not either

app/client/components/user/SingleUserManagement.tsx
enjeyw: oh yeah that's true!

app/client/components/user/EditUserForm.tsx
enjeyw: yeah that looks good

tristanhcole:
> Still not 100% convinced about the index signature - in general I'm not a huge fan of combining business logic elements with loosely-structured data in the same object, but looks good apart from that.

Yeah it’s no ideal. I just think having a pre-defined typed interface on the front end would make custom attributes basically pointless. 

How would you go about it?

PR #182, opened 2020-03-03
Organisations defaults

tristanhcole:
Adding organisation defaults to model:
- ISO 3166-1 alpha-2 country_code
- default_disbursement_wei
- require_transfer card
- default lat
- default lng

Front-end now normalizes organisations to org reducer. This follows the same model that we are doing for all other objects.


app/server/models/organisation.py
```diff
@@ -25,7 +27,13 @@ class Organisation(ModelBase):
     
     _external_auth_password = db.Column(db.String)
 
-    _timezone           = db.Column(db.String)
+    default_lat = db.Column(db.Float())
+    default_lng = db.Column(db.Float())
+
+    _timezone = db.Column(db.String)
+    _country_code = db.Column(db.String)
+    _default_disbursement_wei = db.Column(db.Numeric(27), default=0)
```
tristanhcole: @PhilipWafula I've added default disbursement here

app/migrations/versions/961ab9adc300_.py
```diff
@@ -26,9 +25,15 @@ def upgrade():
     op.add_column('organisation', sa.Column('external_auth_username', sa.String(), nullable=True))
     op.add_column('organisation', sa.Column('_external_auth_password', sa.String(), nullable=True))
 
-    for org in session.query(Organisation).execution_options(show_all=True).all():
+    tcr = sa.sql.table('organisation',
+                       sa.Column('id', sa.Integer, primary_key=True),
+                       sa.Column('external_auth_username', sa.String(), nullable=True),
+                       sa.Column('_external_auth_password', sa.String(), nullable=True),
+                       sa.Column('name', sa.String(), nullable=True))
```
tristanhcole: I wasn't able to update the organisation model as the migration from a clean database was failing here, which makes sense in hindsight.

> The migration defines separate models because the models in your code represent the current state of the database, while the migrations represent steps along the way. Your database might be in any state along that path, so the models might not sync up with the database yet. Unless you're very careful, using the real models directly will cause problems with missing columns, invalid data, etc. It's clearer to explicitly state exactly what columns and models you will use in the migration.
https://stackoverflow.com/questions/24612395/how-do-i-execute-inserts-and-updates-in-an-alembic-upgrade-script/24623979

@michielderoos can you please triple check this works! 🙏 📿

tristanhcole: Also for the user migration https://github.com/teamsempo/SempoBlockchain/blob/50b6113f2ecd01749eb5ffc849299a2cc37f5771/app/migrations/versions/87b723e167d3_.py#L25-L34

app/server/views/index.py
```diff
@@ -52,8 +61,13 @@ def upload():
 def deprecatedVendor():
     return render_template('index.html', js_bundle_main = get_js_bundle_filename())
 
+
```
tristanhcole: to be honest, I think this should all be handled client-side.

right now if you go to a production website `https://dev.withsempo.com/login` it doesn't actually load the bundle and returns a 404 😬

tristanhcole: Fixed here https://github.com/teamsempo/SempoBlockchain/pull/189

app/migrations/versions/961ab9adc300_.py
michielderoos: @tristanhcole Seems so obvious in hindsight!

michielderoos: Will dig in and QA all the permutations here (Fresh DB/ unmigrated DB etc...)

PR #184, opened 2020-03-05
slack fix

tristanhcole:
fixing an error when two ID documents are uploaded


tristanhcole:
Merge in config refactor fixing first

PR #188, opened 2020-03-12
Transaction Deduplicator

enjeyw:


eth_worker/eth_manager/task_interfaces/composite.py
```diff
@@ -279,3 +279,67 @@ def with_log(message, uuid):
...
+                signing_address=task.signing_wallet.address,
+                contract_address=task.contract_address,
+                contract_type='ERC20',
+                func='transferFrom',
+                args=[
+                    orginal_recipient,
+                    orginal_sender,
+                    amount
+                ],
+                gas_limit=8000000,
```
michielderoos: Nit: This shouldn't be hardcoded

PR #189, opened 2020-03-12
simpler route handling

tristanhcole:
Routes are now all handled on client-side. The server now returns default index.html file. Also removed loading spinner from index.html as could be confusing if it fails for some reason

### DIFF
--- a/app/server/views/index.py
+++ b/app/server/views/index.py
@@ -1,13 +1,12 @@
 # -*- coding: utf-8 -*-
 from flask import render_template, Blueprint, current_app
-from server import basic_auth, celery_app
 from server.utils.auth import requires_auth
 import glob, os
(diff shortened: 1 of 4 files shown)


app/server/views/index.py
```diff
@@ -18,55 +17,12 @@ def get_js_bundle_filename():
...
-
-@index_view.route('/activate-account/')
-def activate_account():
-    return render_template('index.html', js_bundle_main = get_js_bundle_filename())
 
-@index_view.route('/reset-password/')
-def reset_password():
-    return render_template('index.html', js_bundle_main = get_js_bundle_filename())
+@index_view.route('/', defaults={'path': ''})
+@index_view.route('/<path:path>')
```
michielderoos: So much cleaner!

PR #195, opened 2020-03-25
fix tfa error

tristanhcole:
fix tfa error caused by typo

### DIFF
--- a/app/client/sagas/authSagas.js
+++ b/app/client/sagas/authSagas.js
@@ -464,8 +464,7 @@ function* watchInviteUserRequest() {
 function* validateTFA({ payload }) {
   try {
     const validateTFAresponse = yield call(ValidateTFAAPI, payload);
-
-    yield call(updateOrganisationStateFromLoginData, activated_account);
+    yield call(updateOrganisationStateFromLoginData, validateTFAresponse);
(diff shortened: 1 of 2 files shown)


app/client/sagas/authSagas.js
```diff
@@ -464,8 +464,7 @@ function* watchInviteUserRequest() {
 function* validateTFA({ payload }) {
   try {
     const validateTFAresponse = yield call(ValidateTFAAPI, payload);
-
-    yield call(updateOrganisationStateFromLoginData, activated_account);
```
tristanhcole: `activated_account ` was undefined, causing the saga to throw an error. We should implement basic frontend tests for all core paths

tristanhcole:
Why is the version wrong?

enjeyw:
> Why is the version wrong?

master is 1.1.4 no?

tristanhcole:
Frontend and backend are on different versioning
https://github.com/teamsempo/SempoBlockchain/blob/master/app/package.json

enjeyw:
> Frontend and backend are on different versioning
> https://github.com/teamsempo/SempoBlockchain/blob/master/app/package.json

ohhh yepyep forgot!

PR #200, opened 2020-03-27
Registering acceptable domains

EDsCODE:
-fixed bug with string matching for accepted domains where the substring comparison wasn't being satisfied and the config file didn't have withsempo at all!

### DIFF
--- a/app/server/api/auth_api.py
+++ b/app/server/api/auth_api.py
@@ -86,9 +86,11 @@ class RegisterAPI(MethodView):
         tier = None
         sempoadmin_emails = current_app.config['SEMPOADMIN_EMAILS']
 
-        if sempoadmin_emails != [''] and email in sempoadmin_emails:
-            email_ok = True
-            tier = 'sempoadmin'
+        for allowed_domain in sempoadmin_emails:
+            if sempoadmin_emails != [''] and allowed_domain in email:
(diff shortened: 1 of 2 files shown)


app/server/api/auth_api.py
```diff
@@ -86,9 +86,11 @@ def post(self):
         tier = None
         sempoadmin_emails = current_app.config['SEMPOADMIN_EMAILS']
```
michielderoos: Does reading config parse the CSV for you?

EDsCODE: yeah, it was getting the values but the condition below `email in sempoadmin_emails` is checking if the entire email is in the accepted domains

EDsCODE: oh wait is this supposed to be working differently than i'm thinking

EDsCODE:
I think i realized my confusion. In trying to restart my entire deployment, I ran into the issue where registering my account wouldn't work because the server would respond with invalid domain. I looked at the PR and noticed you were comparing full emails. @enjeyw with the new PR you did for parsing admin emails, are we supposed to set the config variable ourselves or should we have the email_whitelist table populated when the database initialized during the migrate.py update_data call?

EDsCODE:
@michielderoos clarified the setup. we're good now!

PR #203, opened 2020-04-01
Deployment tweaks

enjeyw:
- secrets bucket can be specified via env file
- local build process also retrieves git-hash tagged docker images


enjeyw:
DON'T MERGE YET! Some docker tag naming conventions being worked out still, so holding off merging this until we're consistent

PR #204, opened 2020-04-01
Correct wrong abi registry lookup key for ERC20

nolash:
`eth_worker/eth_manager` adds abis as a side-effect on inclusion, but does not add the `ERC20Token` key. This key is in use in `app/server/utils/blockchain_task.py`


enjeyw:
Closing - created upstream branch with PR:
https://github.com/teamsempo/SempoBlockchain/pull/208

PR #206, opened 2020-04-04
Wrong name for account type in kyc functional test

nolash:
Api point `/api/v1/kyc_application/` does not process input parameter `type` but expects `account_type`. 

If the parameter name is corrected in the test, the test breaks.

This PR adds `xfail` decorator to the failing tests, and also adds tests for updating kyc_status in preparation for implementing an api client to update this value.


test/functional/app/api/test_kyc_application_api.py
```diff
@@ -1,6 +1,9 @@
 import json, pytest
+import logging
 
+logg = logging.getLogger(__name__)
```
enjeyw, line 4: We should definitely be using more logging! I've been terrible about that lol

test/functional/app/api/test_kyc_application_api.py
```diff
@@ -1,6 +1,9 @@
...
 
+@pytest.mark.xfail(reason='test uses wrong parameter for account type against the api, when replaced with correct the test fails')
```
enjeyw, line 6: Not sure if we should be just xfailing thing rather than working out what's gone wrong?
I'll have a quick squiz but @tristanhcole u might be better placed to diagnose the situation here?

enjeyw: Hmmm yeah i don't know enough about how this api works, but seems to be something to do with `user_id` only being needed if the user is an admin adding another user...?

test/functional/app/api/test_kyc_application_api.py
```diff
@@ -12,19 +15,79 @@ def test_kyc_application_api_post(test_client, complete_admin_auth_token, type,
...
+                                data=json.dumps(dict(account_type=type, user_id=create_transfer_account_user.id, status='VERIFIED')),
+                                follow_redirects=True,
+                                content_type='application/json',
+                                )
+    logg.debug('kyc approve response {}'.format(response.data))
+
+    assert response.status_code == 200
+
+
+@pytest.mark.xfail(reason='due to the limitation that create_transfer_account_user is executed per module only, it is not possible to test with multiple users')
```
enjeyw, line 64: Mmmmm yeah this is annoying. Not sure what the soln is... some factory of some kind?

Either way, this test looks identical to the one above. Can we dry out with the following?

```
@pytest.mark.parametrize('type', [
    ('INDIVIDUAL'),
    pytest.param(
        'BUSINESS',
        marks=pytest.mark.xfail(reason="due to the limitation that create_transfer_account_user is executed per module only, it is not possible to test with multiple users")
    )

])
```

tristanhcole:
See updates here: https://github.com/teamsempo/SempoBlockchain/tree/lash/wrong-param-for-kyc-account-type

test/functional/app/api/test_kyc_application_api.py
tristanhcole: Yeah another way is to just use the two `create_transfer_account_user` and `create_user_with_existing_transfer_account`

tristanhcole: But agree with @enjeyw, the whole point of pytest.mark.parametrize is DRY

test/functional/app/api/test_kyc_application_api.py
tristanhcole: yeah I think it's a bug... should add this code to the KYC API. It needs to be rebuilt as logic is messy

https://github.com/teamsempo/SempoBlockchain/blob/21f89aa4d0ff9e5f0f2a0e2a89b7a9bdf28060a2/app/server/api/kyc_application_api.py#L304-L306

tristanhcole: There's 3 flows currently for the api
1. Mobile User KYC
2. Web KYC flow for an organisation as a superadmin
3. Web KYC flow for a user as another admin user

test/functional/app/api/test_kyc_application_api.py
tristanhcole: Probably should have a `create_transfer_account_user` as a function instead of a module also

tristanhcole:
@nolash can you update this PR with updates?

nolash:
@tristanhcole thanks I'll have a look at this first chance I get.

nolash:
@tristanhcole I see you already amended the code. I have a couple of comments on your commit:

https://github.com/teamsempo/SempoBlockchain/commit/21f89aa4d0ff9e5f0f2a0e2a89b7a9bdf28060a2

tristanhcole:
I don't think I can push to this PR? So I think you'll need to merge in changes and update this PR ...

nolash:
@tristanhcole done but let's amend the comments too please.

PR #215, opened 2020-04-09
Accelerate disbursement calls

michielderoos:
There's some unneeded async stuff. Let's get it outta here!


app/server/utils/blockchain_tasks.py
```diff
@@ -179,17 +179,8 @@ def make_token_transfer(self, signing_address, token,
         :param prior_tasks: list of task uuids that must complete before txn will attempt
         :return: task uuid for the transfer
         """
-
         raw_amount = token.system_amount_to_token(amount)
 
-        balance_wei = self.get_wallet_balance(from_address, token)
```
tristanhcole, line 185: Is this the async stuff? All the other changes seem to be just removing empty lines..

michielderoos: There's also caching the result of the `decimals` call in `app/server/models/token.py`. That did an async call with the same result for every single disbursement-- huge perf hit/ Like I said, it's a simple PR, but the performance difference to the API is pretty big

Issue #217, opened 2020-04-10
Hard Coded KYC Limits can't be changed


WillRuddick:
We need a way to set levels and limits without changing code.
Can we put these variables in a config file or on the db?

See issue here:
https://github.com/GrassrootsEconomics/CIC-Docs/issues/43

tristanhcole:
Closed by https://github.com/teamsempo/SempoBlockchain/pull/219

PR #219, opened 2020-04-13
Make transfer limits configurable

nolash:
Also replaces prints in tests with logging, and outputs all uppercase symbols on debug level:


nolash:
resolves https://github.com/teamsempo/SempoBlockchain/issues/217

tristanhcole:
@nolash thanks for the PR. Transfer limits are hard to change by design. We will think through how this fits into our current roadmap - might be something that is better placed on it's own fork.

nolash:
@tristanhcole I'm not quite sure I understand what you mean. The effect of this PR is exactly the same as before, just that it's moved to the config part. For your own installation only you have access to the config files, the limits will be the same. The limits are set at boot time in both cases. So unless you have an elaborate method to generate proofs of the constant parts of the code to whoever would inspect this, how would you know the difference?

I am trying to avoid forks, because every single deviation potentially adds overhead to future merges.

tristanhcole:
> @tristanhcole I'm not quite sure I understand what you mean. The effect of this PR is exactly the same as before, just that it's moved to the config part. For your own installation only you have access to the config files, the limits will be the same. The limits are set at boot time in both cases. So unless you have an elaborate method to generate proofs of the constant parts of the code to whoever would inspect this, how would you know the difference?
> 
> I am trying to avoid forks, because every single deviation potentially adds overhead to future merges.
> 

After further reflection, I think this is a very reasonable PR and also to limit forks, I shouldn't have been so quick to dismiss this PR and will do better next time.

We don't do any build time verification to ensure limits so I think in the nature of reconfigurability and open source it does make sense to move limits to config.

config.py
```diff
@@ -352,3 +356,35 @@ def address_from_private_key(private_key):
...
+        v = int(config_parser['LIMITS'][k])
+        logg.debug('replacing transfer limit {}:{} with value {}'.format(k.upper(), TRANSFER_LIMITS[k.upper()], v))
+        TRANSFER_LIMITS[k.upper()] = v 
+except KeyError:
+    pass
+    
+l = locals()
+for v in dir():
+    if v.upper() == v:
+        logg.debug('loaded {}: {}'.format(v, l[v]))
```
tristanhcole, line 389: This really emphasises _just how many configs we have._

config.py
```diff
@@ -352,3 +356,35 @@ def address_from_private_key(private_key):
     GE_DB_PORT = ''
     GE_DB_PASSWORD = ''
     GE_HTTP_PROVIDER = ''
+
+TRANSFER_LIMITS = {}
+TRANSFER_LIMITS['0.P7']	= 5000
+TRANSFER_LIMITS['0.P30']	= 10000
+TRANSFER_LIMITS['0.WD7']	= 0
+TRANSFER_LIMITS['0.WD30']	= 0
```
tristanhcole, line 364: hm, these are 0 but are not used anywhere. Could make sense to deprecate `no_transfer_allowed` in favour of simply setting `total_amount` to `0`... what are your thoughts @enjeyw ?

https://github.com/teamsempo/SempoBlockchain/blob/ac12780b7443e3efdea669af9214fa14266dbd38/app/server/utils/transfer_limits.py#L265

enjeyw: that's the way it was originally built, but there were some weird confusing things that happened from the implicit assumption of `total amount == 0` being disabling the limit - for instance the system then didn't know if a transfer was being blocked because the user had their maximum allowable amount or because the transfer type wasn't allowed at all.
Once u start handling that edge case it stops being simple anymore and u may as well just be explicit

PR #225, opened 2020-04-16
Philip/ussd flow cleanup

PhilipWafula:
Implements changes suggested in issue: [Send on pin](https://github.com/GrassrootsEconomics/CIC-Docs/issues/58#issue-593552864)


app/server/utils/ussd/token_processor.py
```diff
@@ -220,7 +220,7 @@ def fetch_exchange_rate(user: User):
             )
 
     @staticmethod
-    def send_token(sender: User, recipient: User, amount: float, reason_str: str, reason_id: int):
+    def send_token(sender: User, recipient: User, amount: float, reason_str: str = None, reason_id: int = None):
```
enjeyw, line 223: Tiny change, but can you please change the typing here to:
`reason_str: Optional[str] = None, reason_id: Optional[int] = None`
https://docs.python.org/3/library/typing.html#typing.Optional

app/server/utils/ussd/ussd_tasks.py
```diff
@@ -11,7 +11,7 @@ def send_directory_listing(user: User, chosen_transfer_usage: TransferUsage):
         processor.send_directory_listing()
 
     @staticmethod
-    def send_token(sender: User, recipient: User, amount: float, reason_str: str, reason_id: int):
+    def send_token(sender: User, recipient: User, amount: float, reason_str: str = None, reason_id: int = None):
```
enjeyw, line 14: same as above

test/helpers/ussd_tasker.py
```diff
@@ -8,7 +8,7 @@ def send_directory_listing(user: User, chosen_transfer_usage: TransferUsage):
         pass
 
     @staticmethod
-    def send_token(sender: User, recipient: User, amount: float, reason_str: str, reason_id: int):
+    def send_token(sender: User, recipient: User, amount: float, reason_str: str = None, reason_id: int = None):
```
enjeyw, line 11: same same lol

app/server/utils/ussd/kenya_ussd_processor.py
```diff
@@ -88,19 +88,25 @@ def custom_display_text(menu: UssdMenu, ussd_session: UssdSession) -> str:
             else:
                 return i18n_for(user, "{}.bio".format(menu.display_key), user_bio=bio_text)
 
-        if menu.name == 'send_token_confirmation':
+        if menu.name == 'send_token_pin_authorization':
```
enjeyw, line 91: Now that this sections handles the pin auth logic for payments, thoughts on removing `if 'pin_authorization' in menu.name` from here in order to tidy things up and make them not confusing? (The `current_pin` logic needs to remain as it's used for pin changes)
https://github.com/teamsempo/SempoBlockchain/blob/360a1c7cc57197527aa02c20ac907f68493e48f8/app/server/utils/ussd/kenya_ussd_processor.py#L131

PhilipWafula: But isn't it matching all other scenarios where there's pin authorization ?
Like `balance_inquiry_pin_authorization` I think it's still in use

enjeyw: oh good catch - maybe change that 'in' match then to be explicit? (see the comment above) It's a bit whacky at the moment  how the `send_token_pin_authorization` technically matches here but just gets caught earlier. Just something like
`if menu.name in ['balance_inquiry_pin_authorization', 'current_pin']` so that in the future a dev looking at it will know it's only for those two cases...

enjeyw:
> good catch on the extra use case of the authorisation, but i still think we should clarify it a bit

Any update on this @PhilipWafula ?

app/server/utils/ussd/kenya_ussd_processor.py
PhilipWafula: > > good catch on the extra use case of the authorisation, but i still think we should clarify it a bit
> 
> Any update on this @PhilipWafula ?I ma

@

> > good catch on the extra use case of the authorisation, but i still think we should clarify it a bit
> 
> Any update on this @PhilipWafula ?



> > good catch on the extra use case of the authorisation, but i still think we should clarify it a bit
> 
> Any update on this @PhilipWafula ?

Hey @enjeyw  makes a lot of sense, I had totally missed this thread in my email. I am terribly sorry for that. Makes a lot of sense, let me push a small alteration for this in a few today

enjeyw: All good - the looks good now!

enjeyw: @PhilipWafula any reason u used:
```
pin_authorization_transitions = ['current_pin', 'pin_authorization']
for transition in pin_authorization_transitions:
    if transition in menu.name:
``` 
instead of just
```
if menu.name in ['balance_inquiry_pin_authorization', 'current_pin']: 
```
?
The former will match if `menu.name = send_token_pin_authorization` which is what we're trying to avoid... unless i'm missing something?

PhilipWafula: The former is what we're going for because these are components of the menu name to look for which will then trigger the pin authorization flow where if the pin is wrong, it will fail and use the retry menu.

I tried the later and it misses some menus, there are quite a number of menus have the `pin_authorization` component in the menu name and all of them need to use that pin flow.
The ones with pin authorization that are separate from these two are apart because the need some data from the ussd processor that the pin flow does not offer

enjeyw: oh that's my mistake, I thought there were only those two transitions lol! Sorry!

In that case, I think the _original_ code here is actually fine - it achieves the same thing and is simple. Sorry for sending u down a rabbit hole.

tristanhcole:
@PhilipWafula @WillRuddick This PR is different from the original spec written on trello https://trello.com/c/3ehIH8AK/124-send-on-pin-entry but does follow the spec on CIC docs https://github.com/GrassrootsEconomics/CIC-Docs/issues/59. Confirming that this PR should remove business categories before merging.

Also NIT @PhilipWafula, if Business Category is being removed then _for None_ should also be removed.
`Successfully received a payment of 1.00 ACME from tristan cole +61401391418 for None on 06/05/2020 at 01:47 PM. New ACME balance is 1.00
`

PR #227, opened 2020-04-17
Updated: Make transfer limits configurable

enjeyw:
Modification to  [https://github.com/teamsempo/SempoBlockchain/pull/219](219) that removes debug logging of secrets in plaintext


app/server/utils/transfer_limits.py
```diff
@@ -255,50 +255,50 @@ def __init__(self,
 LIMITS = [
     TransferLimit('Sempo Level 0: P7', GENERAL_PAYMENTS,
                   is_any_token_and_user_is_not_phone_and_not_kyc_verified, 7,
-                  total_amount=5000),
+                  total_amount=config.TRANSFER_LIMITS['0.P7']),
```
enjeyw, line 258: Configs needs to be accessed in this way, rather than using the `current_app` context because Limits are loaded in during import, before application context is created.

PR #210, opened 2020-04-06
frontend tests: “Zero to Some”

tristanhcole:
Let's get frontend test coverage _above zero!_

- localStorageMock 📁
- redux mock store 📁
- utils.test.js 🔨

also closes https://github.com/teamsempo/SempoBlockchain/issues/224

PR #186, opened 2020-03-07
condense and consolidate types/actions with generic

EDsCODE:
-use generics to reduce code needed to create actions and infer type
-improvement is to reduce amount of declarations needed in the types file (e.g. declaring the action shape with payloads when it can be inferred) also reduces amount of imports needed

PR #228, opened 2020-04-20
typescriptifying

tristanhcole:
Converting additional reducers to typescript

* user 👤
* export

Converting typescript reducers to generics:
* message 💬
* organisation 🏢
* transfer usages

All reducers written in Typescript now use generics/action creators and follow the same pattern setup by @EDsCODE


app/client/sagas/userSagas.js
```diff
@@ -154,17 +158,20 @@ function* resetPin({ payload }) {
...
-      message: reset_response.message
-    });
+    yield put(
+      MessageAction.addMessage({
+        error: false,
+        message: reset_response.message
+      })
+    );
   } catch (fetch_error) {
     const error = yield call(handleError, fetch_error);
```
tristanhcole, line 168: we should consider adding the following code to the `handleError` function in utils.js: `store.dispatch(MessageAction.addMessage({ error: true, message: error.message }))`

tristanhcole:
Includes fix for following error introduced in https://github.com/teamsempo/SempoBlockchain/commit/0b570d6f67f788ec39a0ec2cca92f4bbf433b603#comments PR #210

```
Error: call: argument fn is undefined or null
    at check (io-6de156f3.js?86aa:33)
    at validateFnDescriptor (io-6de156f3.js?86aa:455)
    at call (io-6de156f3.js?86aa:524)
    at showFlashMessage$ (messageSaga.js?6e97:13)
    at tryCatch (runtime.js?f3a2:65)
    at Generator.invoke [as _invoke] (runtime.js?f3a2:303)
    at Generator.prototype.<computed> [as next] (runtime.js?f3a2:117)
    at next (redux-saga-core.esm.js?42a9:1157)
    at currCb (redux-saga-core.esm.js?42a9:1251)
    at eval (redux-saga-core.esm.js?42a9:481)
eval @ index.js?69b4:3680

index.js?69b4:3680 The above error occurred in task onAddFlashMessage
    created by onAddFlashMessage
    created by takeEvery(ADD_FLASH_MESSAGE, onAddFlashMessage)
    created by rootSaga
    created by rootSaga
    created by rootSaga
Tasks cancelled due to error:
takeEvery(ADD_FLASH_MESSAGE, onAddFlashMessage)
```

tristanhcole:
Marking as ready for review as starting to get quite large

app/client/sagas/userSagas.js
enjeyw: though that does presume we necessarily want to flash the error (sometimes it needs interpretation i think?). Guess we could also add that as an input arg...

app/client/components/user/SingleUserManagement.tsx
```diff
@@ -66,15 +74,15 @@ class SingleUserManagement extends React.Component<Props> {
         custom_attributes: attr_dict,
         business_usage_name: businessUsage
       },
-      user_id
+      parseInt(user_id)
```
enjeyw, line 77: This feels weird?
```  
const user_id = this.props.userId.toString();
...
parseInt(user_id)
```

F

app/client/sagas/exportSaga.js
```diff
@@ -9,30 +9,27 @@ import {
...
-} from "../reducers/exportReducer.js";
+import { ExportAction } from "../reducers/export/actions";
+import { ExportActionTypes } from "../reducers/export/types";
 
 import { exportAPI } from "../api/exportAPI";
 
 function* newExport({ payload }) {
   try {
     const result = yield call(exportAPI, payload);
-    if (result.status === "success") {
```
enjeyw, line 24: a great PR would be to
CMD+F `== "`  and remove any string comparisons where possible

app/client/reducers/rootReducer.ts
```diff
@@ -42,7 +42,7 @@ const appReducer = combineReducers({
   spreadsheetUpload,
   datasetSave,
   datasetList,
-  newExportRequest,
+  export: ExportReducer,
```
enjeyw, line 45: wouldn't it make more sense just to follow the common pattern and name the reducer `export`?

enjeyw:
> Marking as ready for review as starting to get quite large

Hahaha yeah there's like 5 seperate PRs in here :S

app/client/components/user/SingleUserManagement.tsx
tristanhcole: haha will fix

app/client/sagas/exportSaga.js
tristanhcole: yeah, I actually think this is the last one in a reducer

app/client/reducers/rootReducer.ts
tristanhcole: i guess so, no strong opinion

tristanhcole:
> I'm getting a bug that I can't repeat on dev... seems to be a generalised version of this:
> #237
> If i try inviting a new user, it adds fine, but then nothing else loads unless i hard-refresh...

Hm ok, at first glance it sounds like it is a generalized problem to do with #237 which was probably introduced in https://github.com/teamsempo/SempoBlockchain/pull/186. I'll take a look and see if I can implement a fix

tristanhcole:
> > I'm getting a bug that I can't repeat on dev... seems to be a generalised version of this:
> > #237
> > If i try inviting a new user, it adds fine, but then nothing else loads unless i hard-refresh...
> 
> Hm ok, at first glance it sounds like it is a generalized problem to do with #237 which was probably introduced in #186. I'll take a look and see if I can implement a fix

Hot fix for this error https://github.com/teamsempo/SempoBlockchain/pull/242 @enjeyw

app/client/reducers/rootReducer.ts
tristanhcole: I think the reason I did it was `export` is not allowed and we've done that flow before for previous typescript reducers;

```
  transferUsages: TransferUsageReducer,
  organisations: OrganisationReducer,

```

enjeyw: oh export not being allowed is a pretty good justification lol!

PR #232, opened 2020-04-23
Transaction Callback

michielderoos:
Make eth_worker call the main application when blockchain tasks change status. This is to allow the app to display current status on the blockchain, and puts any faults/messages into the main app database. This also brings back the old blockchain_status field to the API!  
image


app/server/api/credit_transfer_api.py
```diff
@@ -342,15 +351,13 @@ def post(self, credit_transfer_id):
 
             else:
                 message = 'Transfer Successful' if auto_resolve else 'Transfer Pending. Must be approved.'
-
+                db.session.commit()
```
michielderoos, line 354: Have to commit here since on big disbursements, these rows have to be in the database by the time the callback comes

eth_worker/sql_persistence/models.py
```diff
@@ -258,4 +265,45 @@ def status_code(cls):
...
+        r = requests.post(callback_url,
+            timeout=5,
+            json=post_data,
+            auth=HTTPBasicAuth(config.INTERNAL_AUTH_USERNAME,
+                               config.INTERNAL_AUTH_PASSWORD))
+    except:
+        r = None
+
+    if r and r.ok:
+        obj_table = BlockchainTransaction.__table__
```
michielderoos, line 293: Doing SQL like this as it prevents `after_update` being triggered recursively

michielderoos:
[Transaction Callback](https://trello.com/c/fp97HG4o/446-transaction-callback)

eth_worker/sql_persistence/models.py
```diff
@@ -258,4 +265,45 @@ def status_code(cls):
...
+        connection.execute(
+            obj_table.update().
+            where(obj_table.c.id == target.id).
+            values(is_synchronized_with_app=True)
+        )
+    else:
+        # NOTE: Soft error handling here for now, as incomplete transactions can always be synched later
+        # where is_synchronized_with_app=False
+        # NOTE: Should change eth_worker to logg at some point 
+        print('[\'WARN\'] Could not reach \'APP_HOST\' URL: {callback_url}. Please check your config.ini')
```
enjeyw, line 303: maybe swap to actual python logging now that we've got it set up?

app/server/models/credit_transfer.py
```diff
@@ -47,6 +46,13 @@ class CreditTransfer(ManyOrgBase, BlockchainTaskableBase):
 
     token_id        = db.Column(db.Integer, db.ForeignKey(Token.id))
 
+    # Present status, and time of last update (according to worker) to ensure the present blockchain_status 
+    # is the newest (since order of ack's is not guaranteed)
+    blockchain_status   = db.Column(db.Enum(BlockchainStatus), default=BlockchainStatus.PENDING)
+    blockchain_hash = db.Column(db.String)
+    last_worker_update  = db.Column(db.DateTime)
+    messages = db.relationship('WorkerMessages', backref='credit_transfer', lazy=True)
```
enjeyw, line 54: given that exchanges also use the above logic, wouldn't it be better to add this to the blockchain-taskable base class?

michielderoos:
[Sync eth_worker status to app database](https://trello.com/c/FEoZw4X7/454-sync-ethworker-status-to-app-database)

app/server/api/worker_callback_api.py
```diff
@@ -0,0 +1,52 @@
...
+        blockchain_task_uuid = post_data.get('blockchain_task_uuid')
+        timestamp = datetime.fromtimestamp(post_data.get('timestamp'))
+        blockchain_status = post_data.get('blockchain_status')
+        error = post_data.get('error')
+        message = post_data.get('message')
+        blockchain_hash = post_data.get('hash')
+
+        transfer = CreditTransfer.query.execution_options(show_all=True).filter_by(blockchain_task_uuid = blockchain_task_uuid).first()
+        if not transfer:
+            # Single 2s sleep/retry on failure. Sometimtes when running locally the CreditTransfer object will not be committed yet
```
michielderoos, line 26: Note: This is getting removed in `defer` PR, when it will no longer be needed

app/server/api/worker_callback_api.py
```diff
@@ -0,0 +1,52 @@
...
+    def post(self):
+        post_data = request.get_json()
+        blockchain_task_uuid = post_data.get('blockchain_task_uuid')
+        timestamp = datetime.fromtimestamp(post_data.get('timestamp'))
+        blockchain_status = post_data.get('blockchain_status')
+        error = post_data.get('error')
+        message = post_data.get('message')
+        blockchain_hash = post_data.get('hash')
+
+        transfer = CreditTransfer.query.execution_options(show_all=True).filter_by(blockchain_task_uuid = blockchain_task_uuid).first()
```
enjeyw, line 24: Shouldn't this be a filter on something that also searches for matching `Exchange` objects too?

app/server/api/worker_callback_api.py
enjeyw: lol

app/server/schemas.py
```diff
@@ -162,6 +162,7 @@ def filter_rejected(self, data, many):
     transfer_subtype        = fields.Function(lambda obj: obj.transfer_subtype.value)
     transfer_mode           = fields.Function(lambda obj: obj.transfer_mode.value)
     transfer_status         = fields.Function(lambda obj: obj.transfer_status.value)
+    blockchain_status       = fields.Function(lambda obj: obj.blockchain_status.value)
```
enjeyw, line 165: Out of interest, can this schema just inherit from the blockchain taskable schema?

app/server/api/worker_callback_api.py
```diff
@@ -0,0 +1,52 @@
...
+            transfer.blockchain_hash = blockchain_hash
+        if error or message:
+            transfer.messages.append(WorkerMessages(error = error, message = message, worker_timestamp = timestamp))
+
+        return ('', 204)
+
+
+# add Rules for API Endpoints
+worker_callback_blueprint.add_url_rule(
+    '/worker_callback',
```
enjeyw, line 49: semantics admittedly, but in keeping with REST where possible (rather than RPC), I think this may better viewed as a creation/modification of a BlockchainTaskable object rather than a procedure call....

enjeyw:
oh and frontend tests are failing for no justifiable reason *shrugs*

app/server/api/worker_callback_api.py
michielderoos: @enjeyw Looking at this again, do we actually want the status fields in the `Exchange` model in the first place? Like when you look at the transfer list on the FE, the `blockchain_status` you see  for any given transfer is actually the `blockchain_status` of an exchange's `from_transfer` or `to_transfer`  object.

enjeyw: yeah we still do want the status field - currently it's a case of the Front End not being sufficiently developed to support exchanges.

app/migrations/versions/49e8a333d285_.py
```diff
@@ -0,0 +1,28 @@
...
+# revision identifiers, used by Alembic.
+revision = '49e8a333d285'
+down_revision = 'd7b89a614ae9'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_index('ix_worker_messages_table_blockchain_task_uuid', table_name='worker_messages')
```
michielderoos, line 21: The OG migration seems to have been merged with the defer PR. This is just a generated migration which removes an unused index, a result of changes that happened after that one was merged

app/server/api/worker_callback_api.py
michielderoos: Makes sense. I made it so the status field on _both_ are set now. The redundancy isn't ideal, but  I think it's worth it to have it shown in the front end until it the FE supports exchanges.

app/server/api/blockchain_taskable_api.py
```diff
@@ -0,0 +1,51 @@
...
+        blockchain_task_uuid = post_data.get('blockchain_task_uuid')
+        timestamp = datetime.fromtimestamp(post_data.get('timestamp'))
+        blockchain_status = post_data.get('blockchain_status')
+        error = post_data.get('error')
+        message = post_data.get('message')
+        blockchain_hash = post_data.get('hash')
+
+        blockchain_tasks = []
+        blockchain_tasks.extend(CreditTransfer.query.execution_options(show_all=True).filter_by(blockchain_task_uuid = blockchain_task_uuid).all())
+        blockchain_tasks.extend(Exchange.query.execution_options(show_all=True).filter_by(blockchain_task_uuid = blockchain_task_uuid).all())
```
enjeyw, line 28: clean af

PR #235, opened 2020-04-24
Adding UWSGI Stats Server

enjeyw:
uWSGI docs recommend using the [stats server/uwsgitop ](https://uwsgi-docs.readthedocs.io/en/latest/StatsServer.html) to monitor apps and understand performance issues. See also https://uwsgi-docs.readthedocs.io/en/latest/ThingsToKnow.html

This PR adds this.


enjeyw:
> Nice! Didn't know this was a thing. Should we limit it to staging eventually? I hesitate to open up too many ports (though I suppose we can use AWS firewall rules)

Yeah that did cross my mind, though the UWSGI best-practice doc seemed to indicate that this should be used for monitoring your prod instances as well.

PR #236, opened 2020-04-25
USSD Sim

tristanhcole:
@enjeyw's been hiding a USSD Sim 🎁. So much easier than setting up AT/ngrok everytime.

Also; adding `devtools` directory


enjeyw:
Our directory is getting pretty clogged with random dev tooling files - wondering if we should create a dev_tools or utils directory or similar.

tristanhcole:
> Our directory is getting pretty clogged with random dev tooling files - wondering if we should create a dev_tools or utils directory or similar.

I think that's a good idea. Adding dev tools to the repo is also a good idea imo

tristanhcole:
TODO: update https://github.com/teamsempo/SempoBlockchain/wiki/AWS-Deployment-Guide once merged in

tristanhcole:
I've added changes so all scripts work with following:
- Program Arguments: `path to virtual env`
- Working Directory: `/SempoBlockchain/devtools`

devtools/prepare_aws_deploy.sh
```diff
@@ -1,14 +1,12 @@
 #!/usr/bin/env bash
 
-bash build_docker.sh
-
-# REPOSITORY_URI=290492953667.dkr.ecr.ap-southeast-2.amazonaws.com/blockchaindemo
+#bash build_docker.sh
```
tristanhcole, line 3: commented?

enjeyw: errggg good catch - was retrying code and didn't wanna wait for the build

tristanhcole: 👍

PR #240, opened 2020-04-28
Run external API calls in threads

michielderoos:
Runs calls to external APIs in background threads. This is for two reasons: One is for performance-- the user shouldn't have to wait for (for example) an SMS to be sent when their account is created. Secondly, there are a limited number of app threads and long running jobs (like if a server is not responding) can tie them all up and easily shut down the server. This keeps server threads free for doing real work.


michielderoos:
[Make API calls happen async in separate thread](https://trello.com/c/fHPO6Fy2/456-make-api-calls-happen-async-in-separate-thread)

app/server/__init__.py
```diff
@@ -55,6 +55,7 @@ def create_app():
 
     app.config.from_object('config')
     app.config['BASEDIR'] = os.path.abspath(os.path.dirname(__file__))
+    app.config['EXECUTOR_PROPAGATE_EXCEPTIONS'] = True
```
michielderoos, line 58: Putting this in `__init__.py` and not in `config.py` or `*_config.ini` files because we _never_ want this to be False

enjeyw: 🎶 Forever. Forever-never? Forever-never.

PR #244, opened 2020-04-30
Partially Cache Metrics Results

michielderoos:


app/server/utils/metrics_cache.py
```diff
@@ -0,0 +1,56 @@
...
+
+def _sum_strategy(query, cache_result):
+    cache_result = float(cache_result)
+    return float(query.first().total or 0) + cache_result
+
+def _count_strategy(query, cache_result):
+    cache_result = int(cache_result)
+    return query.count() + cache_result
+
+strategy_functions = { SUM: _sum_strategy, COUNT: _count_strategy }
```
enjeyw, line 56: this is clean

enjeyw:
seems like a solid start (pretty much what i was expecting based on our convo this morning) - as u note, needs some way of handling the fact different queries on the same metric can have different results, but i think we have a way around that....

app/server/utils/metrics_cache.py
```diff
@@ -0,0 +1,89 @@
...
+def _store_cache(key, value):
+    pickled_object = pickle.dumps(value)
+    red.set(key, pickled_object)
+    return True
+
+def _load_cache(key):
+    cached_object = red.get(key)
+    if not cached_object:
+        return None
+    return pickle.loads(cached_object)
```
enjeyw, line 25: I got a corrupted cache, and it caused pickle to die. This allowed it to recover:

```
def _load_cache(key):
    cached_object = red.get(key)
    if not cached_object:
        return None
    try:
        return pickle.loads(cached_object)     
    except pickle.UnpicklingError:
        red.delete(key)
        return None
```

app/server/utils/metrics_cache.py
```diff
@@ -0,0 +1,89 @@
...
+        return None
+    return pickle.loads(cached_object)       
+
+def execute_with_partial_history_cache(metric_name, query, object_model, strategy, disable_cache = False):
+    # disable_cache pass-thru. This is so we don't cache data when filters are active.
+    if disable_cache:
+        return _handle_combinatory_strategy(query, None, strategy)
+
+    # Redis object names
+    CURRENT_MAX_ID = f'{g.active_organisation}_{object_model.__table__.name}_max_id'
```
enjeyw, line 33: thoughts on using `f'org{g.active_organisation.id}...` instead? not really a big deal, but results in a slightly cleaner key

app/server/utils/metrics_cache.py
michielderoos: Good catch. Do you remember the conditions that made that happen? I did make that change though, also had to add the following chunk of code to ensure that a corrupted cache won't lead to undercounted results 
```
    # If there's no cache (either it's a new attribute we're tracking, or the cache is corrupted)
    # then we should pull results starting at id=0
    if cache_result:
        filtered_query = query.filter(object_model.id > highest_id_in_cache)
    else:
        filtered_query = query
```

michielderoos:
@enjeyw 
> turned out to be a massive Pain due to decimal types - got it ostensibly working, but wasn't convinced it was parsing the data correctly so left it lol  
  
That was part of it, but not the whole story! I started by formulating my own little json schema, where every object would be stored with a string connoting its type (I.e. `{"type":"Decimal", "Value":"3.14"}. I'd cast the objects themselves as strings to store, and cast them back to their input formats afterwards. This largely side-stepped the Decimal issue, but in the middle of writing a caster to stringify/destringify a list of tuples of `(datetime, Numeric)` (where Numeric can be int or float or Decimal)  I realized that pickle probably made more sense to use here.

enjeyw:
@michielderoos hahaha yeah the second i realised it was getting complex I was like "hmmmm I have a feeling Michiel's probably been down this path already... should probably quit while I'm ahead" 😆

PR #245, opened 2020-05-01
typescript coverage

tristanhcole:
Small fix to include typescript in frontend coverage

Previously was raising error `TS5056`:
```
error TS5056: Cannot write file '$$ts-jest$$/client/components/AsyncButton.js' because it would be overwritten by multiple input files.
```


app/tsconfig.json
```diff
@@ -3,28 +3,30 @@
...
     // "outFile": "./",                       /* Concatenate and emit output to single file. */
-    // "outDir": "./",                        /* Redirect output structure to the directory. */
+    "outDir": "./",
+    /* Redirect output structure to the directory. */
     // "rootDir": "./",                       /* Specify the root directory of input files. Use to control the output directory structure with --outDir. */
     // "composite": true,                     /* Enable project compilation */
     // "tsBuildInfoFile": "./",               /* Specify file to store incremental compilation information */
     // "removeComments": true,                /* Do not emit comments to output. */
-    // "noEmit": true,                        /* Do not emit outputs. */
+    "noEmit": true,
```
tristanhcole, line 22: we use webpack for compilation and typescript just for type checking

PR #246, opened 2020-05-01
Adding Indices to Association Tables

enjeyw:
Adding Indices to Association Tables, especially the organisation table, significantly speeds up queries as we do a lot of filtering based on organisation ids


enjeyw:
> @michielderoos  Running this against a fully up to date master DB gives a migration error (even when I `merge heads`). I did paste the migration code into a fresh revision and test it there though, and it works! Haven't dug in and done any real perf testing, but loading the dashboard definitely seemed snappier on my local machine.

On aws dev we're getting migrations errors because there's a revision from https://github.com/teamsempo/SempoBlockchain/pull/232 in there... could that be the issue

michielderoos:
> > @michielderoos  Running this against a fully up to date master DB gives a migration error (even when I `merge heads`). I did paste the migration code into a fresh revision and test it there though, and it works! Haven't dug in and done any real perf testing, but loading the dashboard definitely seemed snappier on my local machine.
> 
> On aws dev we're getting migrations errors because there's a revision from #232 in there... could that be the issue?

@enjeyw I don't think so-- this was locally on a freshly rebuilt DB based on `master`

PR #239, opened 2020-04-28
Enable Pusher Batching

michielderoos:
Send pusher requests in batches when doing bulk operations.  
This is ready for review, but should **not** be pushed until #240 is merged

PR #248, opened 2020-05-01
Defer transactions til after request

michielderoos:
Makes transactions run after request, and handles push notifications later as well  
Needs [this](https://github.com/teamsempo/SempoBlockchain/pull/239) PR to go first, then target Master (targeting #239 for now to make diff readable)


michielderoos:
> Tested and makes sense to me! How does this change performance @michielderoos?

@tristanhcole Shouldn't have any impact on perf. It does tidy the pusher code a bit, and it is useful for `transaction callback` since we'll be assured that transactions exist in the database _before_ they're executed by the worker (there was a race condition before this)

app/server/__init__.py
```diff
@@ -118,11 +119,17 @@ def before_request():
...
     @app.after_request
     def after_request(response):
-            # Execute any async celery tasks
+        from server.utils import pusher
 
         if response.status_code < 300 and response.status_code >= 200:
             db.session.commit()
 
+        for transaction, queue in g.pending_transactions:
+            transaction.send_blockchain_payload_to_worker(queue=queue)
```
enjeyw, line 128: `send_blockchain_payload_to_worker` sets the task uuid, but there's no commit (which i think is good - commits in the after_Request have killed us in the past)?

michielderoos: I think we will require a commit in after_request eventually though, just to avoid the race condition in the other PR. I did add that in here, and it seems to work fine though.

michielderoos:
@enjeyw   
> * Pusher doesn't seem to thread?  

It was out of date with master, so pusher threading wasn't merged in yet. Got it up to date now and it works for me.       

> * no commit thing as commented already

Good catch on the commit thing, I just preemptively added the pre-set UUID change we discussed to remove the race condition from callbacks as well.    

> * doesn't handle the `bt.make_liquid_token_exchange` transaction type?   

Do you think we actually have to defer that to after_request? Since the requests only ever come in one at a time, we can generate and commit a UUID before the task is sent off to the worker. With pure credit_transfers they're often bulk, so it's worth deferring (n fewer commits, and make pusher batching a code bit cleaner), but I'm not sure about transfers. What do you think?

app/server/utils/pusher.py
```diff
@@ -10,6 +10,14 @@
 # https://pusher.com/docs/channels/library_auth_reference/rest-api#post-batch-events-trigger-multiple-events-
 PUSHER_MAX_BATCH_SIZE = 10
 
+@executor.job
+def async_pusher_trigger(*args, **kwargs):
+    pusher_client.trigger(*args, **kwargs)
```
enjeyw: wow.                                                                much generic. 
                                                                                                                                multipurpose.
                                           keyword args.

app/server/models/credit_transfer.py
```diff
@@ -112,6 +113,8 @@ def resolve_as_completed(self, existing_blockchain_txn=None, queue='high-priorit
         if self.fiat_ramp and self.transfer_type in [TransferTypeEnum.DEPOSIT, TransferTypeEnum.WITHDRAWAL]:
             self.fiat_ramp.resolve_as_completed()
         if not existing_blockchain_txn:
+            self.blockchain_task_uuid = str(uuid4())
+            db.session.commit()
```
enjeyw, line 117: committing deep inside processes often causes weird side-effects (the org filter can get angry) - is it possible to push this up higher?

enjeyw:
@michielderoos 
> It was out of date with master, so pusher threading wasn't merged in yet
Doi. Not sure how i missed that lol.

> the requests only ever come in one at a time, we can generate and commit a UUID before the task is sent off to the worker
Generally speaking, I _really_ prefer keeping database commits out of deep business logic (it's an [Alchemy Antipattern ](https://docs.sqlalchemy.org/en/13/orm/session_basics.html#when-do-i-construct-a-session-when-do-i-commit-it-and-when-do-i-close-it) and messes up our org level filter real bad.
Given `bt.make_liquid_token_exchange` is currently called from within `exchange_from_amount`. If we want to get the uuid into the database before sending a task, i see three options (all involve supplying the uuid obvs):
-Generate a uuid, and then the db commit from the entrypoint/manager function to inside `exchange_from_amount`, between uuid generation and task dispatch (which i dislike due to said businessLogicCommitPhobia)
- Move task dispatch out of `exchange_from_amount` to something much higher (gross)
- Handle uuid-needing transaction tasks generically using some deferral mechanism that's guaranteed to fire on the way out after commits

app/server/models/credit_transfer.py
enjeyw: @michielderoos just checking in on this one bit

app/server/__init__.py
```diff
@@ -119,11 +121,20 @@ def before_request():
...
+
+        for exchange_args in g.pending_exchanges:
+            bt.make_liquid_token_exchange(**exchange_args)
```
enjeyw, line 133: I'm fully aware that is one is my fault, but I figure now is as good as time as any to fix it lol...

credit transfers and exchanges use a different level of abstractions to send tasks to the blockchain worker:
 credit transfers have a nice `send_blockchain_payload_to_worker` method, where as `pending_exchanges` just has this gnarly bunch of blockchain tasks tied directly into `exchange_from_amount`.

Only really noticed because these two for loops next to each other are totally different. Thoughts on creating a  `send_blockchain_payload_to_worker(queue)` method in Exchange that holds all the blockchain tasks that need to be fired? As an added bonus, it means we could just append all the tasks to `g.pending_transactions` and fire them in a single neat loop.

app/server/models/exchange.py
```diff
@@ -159,19 +159,20 @@ def get_exchange_rate(from_token, to_token):
...
-            transfer_mode=None
+            transfer_mode=None, queue='high-priority'
     ):
         self.user = user
         self.from_token = from_token
         self.to_token = to_token
+        self.from_amount = from_amount
 
-        exchange_contract = self._find_exchange_contract(from_token, to_token)
+        self.exchange_contract = self._find_exchange_contract(from_token, to_token)
```
michielderoos, line 169: Binding this to the exchange object so we don't have to search for it twice

app/server/__init__.py
michielderoos: @enjeyw I think that makes sense. I committed this change for you to have a look at. I think there's a (small) compromise, just in that we have to filter out exchanges in the `after_request` pusher batch.

michielderoos:
@enjeyw Good catch! The tests all run just fine without any modification, so I don't think we'll have to any session-detecting. With just your suggested change to the dev_data script, all the tests pass and `dev_data` seems to run as expected.

PR #249, opened 2020-05-02
Missing default limit on reserve org

nolash:
Workaround for https://github.com/GrassrootsEconomics/CIC-Docs/issues/92 when checking balances


nolash:
wrong repo sorry

PR #250, opened 2020-05-02
Fix key collision for account create ussd menu item

nolash:


nolash:
wrong repo again. sorry again.

PR #251, opened 2020-05-04
refactor(config): deprecating currency_name & currency_decimals

tristanhcole:
Currency Name and Decimals are used by the mobile API, making a small change to enable deprecation from config and just use `default_organisation.token` for now.

Adding `display_decimals` to the `Token` model. Default to 2.


app/server/utils/user.py
```diff
@@ -703,18 +703,17 @@ def send_terms_message_if_required(user):
         user.seen_latest_terms = True
 
 
-
-def send_onboarding_message(to_phone, first_name, credits, one_time_code):
+def send_onboarding_message(to_phone, first_name, amount, currency_name, one_time_code):
```
tristanhcole, line 706: This isn't used anywhere currently

app/server/utils/auth.py
```diff
@@ -354,15 +353,14 @@ def create_user_response_object(user, auth_token, message):
         'server_time': int(time.time() * 1000),
         'ecdsa_public': current_app.config['ECDSA_PUBLIC'],
         'pusher_key': current_app.config['PUSHER_KEY'],
-        # todo: (used on mobile) Deprecate. Currency should be based on active organization/TA account token
-        'currency_decimals': current_app.config['CURRENCY_DECIMALS'],
-        'currency_name': currency_name,
+        'currency_decimals': currency_decimals,
+        'currency_name': currency_symbol,  # todo(mobile-api): rename currency_name to currency_symbol
```
tristanhcole, line 357: Currency Name is actually referencing currency symbol so for now just adding a todo

tristanhcole:
[Deprecate Currency Name, Decimal and Conversion rate from config](https://trello.com/c/Og3q4U3Y/291-deprecate-currency-name-decimal-and-conversion-rate-from-config)

app/server/utils/user.py
```diff
@@ -703,18 +703,17 @@ def send_terms_message_if_required(user):
...
             .format(
             first_name,
             current_app.config['PROGRAM_NAME'],
-            credits if not None else 0,
-            current_app.config['CURRENCY_NAME'],
+            amount if not None else 0,
+            currency_name,
             one_time_code,
-            current_app.config['CURRENCY_NAME']
+            currency_name
```
michielderoos, line 716: Nit: line 716 can be removed, `currency_name` is only used once in formatting this string

tristanhcole: nice catch 👍

app/migrations/versions/8a9941c54185_.py
```diff
@@ -0,0 +1,30 @@
...
+# revision identifiers, used by Alembic.
+revision = '8a9941c54185'
+down_revision = '984d7e9c50af'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('token', sa.Column('display_decimals', sa.Integer(), nullable=True))
```
tristanhcole, line 20: @enjeyw y'all happy with this?

enjeyw:
I don't know if we should necessarily be deprecating this so much as setting it as the default for the master org. Also decimals should be made consistent with the ERC20 token (ie standard 18)

tristanhcole:
> I don't know if we should necessarily be deprecating this so much as setting it as the default for the master org. Also decimals should be made consistent with the ERC20 token (ie standard 18)

Decimals at 18? New column is for displaying on mobile/web frontend. We don't use it for setting default for master org currently and not sure we would necessarily need that

tristanhcole:
@enjeyw did you have any further thoughts on this? I think its fine to merge in

app/server/utils/auth.py
```diff
@@ -323,12 +322,12 @@ def create_user_response_object(user, auth_token, message):
         usd_to_satoshi_rate = None
 
     conversion_rate = 1
-    currency_name = current_app.config['CURRENCY_NAME']
+    currency_symbol = user.default_organisation.token.symbol if user.default_organisation and user.default_organisation.token else None
+    currency_decimals = user.default_organisation.token.display_decimals if user.default_organisation and user.default_organisation.token else None
```
enjeyw, line 326: I don't love this because as a developer coming in, I'd expect that if 
`currency_symbol` is given by `token.symbol`
then 
`currency_decimals` would be given by `token.decimals`, but it's not, which is pretty confusing.

I realise that this naming mess up is somewhat my fault, as it can be traced back to me in Lebanon naming the number of digits displayed behind the decimal place as `display decimals`, but I don't think it makes sense to make the problem any worse.

In an ideal world, I'd suggest that we use currency_decimals everywhere as u've got it here and change and `token.decimals` to `token.precision` or something like that, but we're running up against the ERC20 standard:
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-20.md#decimals

For now i think the best thing to do is call it display_decimals as much as we can on the backend, and only use  `currency_decimals` at the last minute in the backend.

More long term, (though i think this requires spec'ing), we should probably use  proper i18n module (libraries or hand written) that support  proper formatting. Then we can just have a display format column that stores something like `[SYM] ###,###.##`

tristanhcole: Hmmm should this PR just be closed in favour of https://github.com/teamsempo/SempoBlockchain/issues/284 @enjeyw ?

enjeyw: I still think `display_decimals` is a decent change to implement for now, as #284 is a slightly different thing. Unless u wanted to jump straight into i18n, but that would be a much longer PR

tristanhcole:
> got a comment on the text. Wanna acknowledge out once again that the only reason i'm requesting changes on this PR is due to a mess I made lol!

resolved in latest commit. this PR should be merged or closed before #425

enjeyw (approved):
Good to go!

PR #254, opened 2020-05-04
Third Party Transaction Sync

enjeyw:
By far the most common use-case here is:
Watch for all ERC20 transfers to and from wallets registered on the platform, and if the transaction was not initiated by the Sempo platform, record it on the system, updating balances as appropriate.

For me, other major questions (and my current thinking as to answers) are:

_Do we try and handle more than just ERC20 Transfers (eg cross-token conversions)?_
Let's build with ERC20 transfers in mind, but with the expectation of extensibility. 

_If yes (or even if 'no' still), do we route all transactions to a generic transaction endpoint and let the app figure out what to do with the txns and the data, or do we tell the worker how to distinguish between transaction types and then route somewhat parsed transactions to individual `Post` endpoints for the relevant object types?_
Posting a 'create object' request to the relevant endpoint seems most in line with current processes, but it means that we need to include callback information in the watching requests to the worker. I'm undecided about this one still, but fortunately both options have similar-ish overall implementations. 

### Telling the worker what to watch for:
A major challenge here is that I think it makes sense for the blockchain worker to do the transaction watching, but the worker is very dumb - it has no idea of what tokens or exchanges exist, so needs some information about what to look for. The natural move is that this comes from the app, but we wouldn't want to duplicate too much information across the two systems.

Current think is that the app would ask the worker to watch for transactions in similar way to how it currently asks it to make transactions
For this, we'd have a table with:
- Contract Type
- Contract Address
- Event Type
- Event Filter Parameters as either
    - Static
    - Dynamic Across All Registered Addresses (perhaps filtered by address type)
- Callback Endpoint and information about what to return?

### System Watches For New Events
- Either via continuous websocket worker or via periodic beat/cron

### Matched Transactions
Once the worker detects a transaction, it first checks any all transaction hashes* in the transaction table to see if it's a transaction that has been seen before. If it hasn't been seen, it either parses the event data into an app-object friendly form, or just forwards it all to the app blindly (depending on which path we choose).

*So we need to make sure this filter process is efficient

### App processing of parsed Txn
Once the app has a parsed txn, it simply appends it to the relevant table.
One catch here is what to do if one of the parties isn't known by the system; in such a case we simply create a transfer account without any attached user and flag it as an external account. Such an account is viewable by all organisations, since the data is public anyway. The balance for that account would probably be undefined - while we could look it up, it's not useful data right now.


enjeyw:
Thanks @michielderoos!

> we should monitor the sync status of every webhook call on the eth_worker side, and maybe have some kind of retry mechanism.

Can u elaborate on this a little more? Given that the worker is initiating a call to the app when it identifies a completed txn, what is monitoring what exactly? Equally, what does a retry mean in this context.

> If someone does a cross-token conversion, wouldn't that effect their balance of our token? 

Yeah this one was something that caught me a little bit. Fortunately, from memory a cross-token conversion includes two `Transfer` events in it (out of one token, in of the other), so we'll still actually capture these balance changes regardless of whether we explicity model conversion or not

michielderoos:
> Can u elaborate on this a little more? Given that the worker is initiating a call to the app when it identifies a completed txn, what is monitoring what exactly? Equally, what does a retry mean in this context.


@enjeyw Apologies, could have been clearer here. What I'm suggesting is a way to prevent us from missing a transaction in case the app's webhook is down (or otherwise broken somehow). The simplest version of this would be a table in the eth_worker DB just listing failed HTTP requests made to the webhook which we want to examine or retry later.

enjeyw:
> > Can u elaborate on this a little more? Given that the worker is initiating a call to the app when it identifies a completed txn, what is monitoring what exactly? Equally, what does a retry mean in this context.
> 
> @enjeyw Apologies, could have been clearer here. What I'm suggesting is a way to prevent us from missing a transaction in case the app's webhook is down (or otherwise broken somehow). The simplest version of this would be a table in the eth_worker DB just listing failed HTTP requests made to the webhook which we want to examine or retry later.

Ohhhhhhhhh gotcha @michielderoos good catch. Could even just add a flag on the transactions table that lists whether its been syncd...

eth_worker/eth_manager/blockchain_sync/blockchain_sync.py
```diff
@@ -0,0 +1,134 @@
...
+    r = requests.post(config.APP_HOST + '/api/v1/credit_transfer/internal/',
+                        json=body,
+                        auth=HTTPBasicAuth(config.INTERNAL_AUTH_USERNAME,
+                                        config.INTERNAL_AUTH_PASSWORD),
+                        timeout=30
+                    )
+    return r
+
+# Get list of filters from redis. This is the starting point of the synchronization process
+def synchronize_third_party_transactions():
```
jzaki, line 35: Called every 30s by the celery beat, so if execution is delayed, could have concurrency issues (more below).

eth_worker/eth_manager/blockchain_sync/blockchain_sync.py
```diff
@@ -0,0 +1,134 @@
...
+    # the range we want into chunks. Once all the chunk-jobs are formed and loaded into redis,
+    # then trigger process_all_chunks, which will consume those jobs from redis
+    for f in filters:
+        latest_block = get_latest_block_number()
+        # If there's no filter.max_block (which is the default for auto-generated filters)
+        # start tracking third party transactions by looking at the lastest_block
+        max_fetched_block = f.max_block or latest_block
+        number_of_blocks_to_get = (latest_block - max_fetched_block)
+        number_of_chunks = ceil(number_of_blocks_to_get/sync_const.BLOCKS_PER_REQUEST)
+
```
jzaki, line 47: A concurrent call could have overlapping chunks when there a lot of chunks to go through. Eg, initial run with filters starting from epoch, or a scripted migration.

eth_worker/eth_manager/blockchain_sync/blockchain_sync.py
```diff
@@ -0,0 +1,134 @@
...
+        persistence_module.mark_transaction_as_completed(transaction_object)
+    return True
+
+# Gets blockchain transaction history for given range
+# Fallback if something goes wrong at this level: block-tracking table
+def get_blockchain_transaction_history(contract_address, start_block, end_block = 'lastest', argument_filters = None, filter_id = None):
+    # Creates DB objects for every block to monitor status
+    config.logg.info(f'Fetching block range {start_block} to {end_block} for contract {contract_address}')
+
+    persistence_module.add_block_range(start_block, end_block, filter_id)
```
jzaki, line 103: With overlapping calls (from above), it is possible to add multiple identical blocks.

eth_worker/sql_persistence/interface.py
```diff
@@ -444,6 +471,43 @@ def set_wallet_last_topup_task_uuid(self, wallet_address, task_uuid):
...
+
+    def add_block_range(self, start, end, filter_id):
+        for n in range(start, end):
+            block = SynchronizedBlock(block_number = n , status = 'PENDING', is_synchronized = False, synchronization_filter_id = filter_id)
+            self.session.add(block)
+        self.session.commit()
+
+    def set_block_range_status(self, start, end, status):
+        for n in range(start, end):
+            block  = self.session.query(SynchronizedBlock).filter(SynchronizedBlock.block_number == n).first()
```
jzaki, line 506: `.first()` is problematic since it's possible for the same block to be added more than once. (assumption of non-concurrency above)

eth_worker/eth_manager/blockchain_sync/blockchain_sync.py
```diff
@@ -0,0 +1,134 @@
...
+        filter = erc20_contract.events.Transfer.createFilter(
+            fromBlock = start_block,
+            toBlock = end_block,
+            argument_filters = argument_filters
+        )
+        for event in filter.get_all_entries():
+            yield event
+
+        # Once a batch of chunks is completed, we can mark them completed
+        persistence_module.set_block_range_status(start_block, end_block, 'SUCCESS')
```
jzaki, line 118: This only sets the first range, but problematic since it could leave multiple blocks with the initial PENDING status.
Either through concurrency, or if the filters can return overlapping block ranges.

eth_worker/sql_persistence/interface.py
michielderoos: Changed to all(), but also added a lock so there should never be multiple now either

eth_worker/eth_manager/blockchain_sync/blockchain_sync.py
michielderoos: Fixed this through the concurrency route, should no longer see overlapping ranges

eth_worker/eth_manager/blockchain_sync/blockchain_sync.py
michielderoos: Covered by adding a lock

eth_worker/eth_manager/blockchain_sync/blockchain_sync.py
michielderoos: Covered by adding a lock

eth_worker/eth_manager/blockchain_sync/blockchain_sync.py
michielderoos: Added a redis lock, should now strictly run only one at a time

eth_worker/eth_manager/blockchain_sync/blockchain_sync.py
```diff
@@ -133,4 +144,5 @@ def add_transaction_filter(contract_address, contract_type, filter_parameters, f
     if not persistence_module.check_if_synchronization_filter_exists(contract_address, filter_parameters):
         # Set max_block to block_epoch to act as a de-factor zero-point
         config.logg.info(f'No filter found for address {contract_address} with parameters {filter_parameters}. Creating.')
-        return persistence_module.add_transaction_filter(contract_address, contract_type, filter_parameters, filter_type, decimals, block_epoch=block_epoch)
+        persistence_module.add_transaction_filter(contract_address, contract_type, filter_parameters, filter_type, decimals, block_epoch=block_epoch)
+        return True
```
jzaki: All other defs of `add_transaction_filter` return the filter, and uses of it expect the filter. Why is this one now boolean?

eth_worker/sql_persistence/interface.py
```diff
@@ -503,8 +504,9 @@ def add_block_range(self, start, end, filter_id):
 
     def set_block_range_status(self, start, end, status):
```
jzaki, line 505: Possible statuses: PENDING, SUCCESS, FAILED.
`add_block_range` takes a filter param, so assuming filters can return txs in the same block, there will be multiple rows with the same block number, but different filter_ids.
`set_block_range_status` does not filter by filter_id, so the multiple rows being set for a block_number may include FAILED statuses of other filters?

eth_worker/eth_manager/blockchain_sync/blockchain_sync.py
```diff
@@ -77,16 +77,18 @@ def handle_transaction(transaction, filter):
...
-        status = 'SUCCESS',
-        block = transaction.blockNumber,
-        hash = str(transaction.transactionHash.hex()),
-        contract_address = transaction.address,
-        is_synchronized_with_app = False,
-        recipient_address = transaction.args['to'],
-        sender_address = transaction.args['from'],
-        amount = float(transaction.args['value'])/(10**filter.decimals)
-    )
+    if not transaction_object:
```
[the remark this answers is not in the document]
enjeyw: This is @michielderoos's code, but as far as i c it is redundant... it stops transaction object from being created if it already exists (eg transaction_object.is_synchronized_with_app isn't satisfied), but creates one if neither of `transaction_object and transaction_object.is_synchronized_with_app` are satisfied.

eth_worker/eth_manager/blockchain_sync/blockchain_sync.py
michielderoos: I think this is from a previous revision? It's returning normally in current [diff](https://github.com/teamsempo/SempoBlockchain/pull/254/files#diff-53f9cba2470dd5f0cd65a798ff6897a1R147)

eth_worker/sql_persistence/interface.py
michielderoos: Wow, good catch! Fixed

eth_worker/eth_manager/blockchain_sync/blockchain_sync_constants.py
```diff
@@ -1 +1,6 @@
 BLOCKS_PER_REQUEST = 5000
+# How long a blockchain_sync lock can last.
+# This is set to 2 days, longer than any `synchronize_third_party_transactions` should last.
+# The reason for this is just in case the app terminates mid-operation and the lock is never
+# freed, this removes the need to go in manually and remove the lock from the redis server.
+LOCK_TIMEOUT = 172800
```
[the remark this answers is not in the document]
michielderoos: Changed it to a _much_ shorter timeout, but renewing the lock with every fetched chunk.

eth_worker/eth_manager/celery_tasks.py
```diff
@@ -80,7 +80,8 @@ def synchronize_third_party_transactions(self):
 
 @celery_app.task(**low_priority_config)
 def add_transaction_filter(self, contract_address, contract_type, filter_parameters, filter_type, decimals = 18, block_epoch = None):
-    return blockchain_sync.add_transaction_filter(contract_address, contract_type, filter_parameters, filter_type, decimals, block_epoch)
+    blockchain_sync.add_transaction_filter(contract_address, contract_type, filter_parameters, filter_type, decimals, block_epoch)
+    return True
```
jzaki, line 84: @michielderoos This is the filter I was referring to (returning boolean instead of the filter). Is this correct?

jzaki: There are 3 other definitions of `add_transaction_filter` in other files that return the filter.

michielderoos: The reason for this is that `blockchain_sync.add_transaction_filter` returns a `TransactionFilter` object, which isn't JSON serializable, and Celery expects things directly returned from its objects to be serializable.  
I did make a change here though-- now I'm building a serializable object containing the filter object's data. Potentially useful if we ever want to change filter creation to a sync call.

jzaki: Okay thanks, it just stood out as the only function of that name returning boolean.
@michielderoos Will you be adding automated tests to show a blockchain tx updating a balance correctly in the db?
What is left before this PR can be merged?

app/server/api/credit_transfer_api.py
```diff
@@ -440,32 +440,61 @@ def post(self):
...
+                    token=token,
+                    sender_transfer_account=send_transfer_account,
+                    recipient_transfer_account=receive_transfer_account,
+                    transfer_type='PAYMENT',
+                )
+                transfer.transfer_status = TransferStatusEnum.COMPLETE
+                transfer.blockchain_status = BlockchainStatus.SUCCESS
+                transfer.blockchain_hash = blockchain_transaction_hash
+                send_transfer_account.decrement_balance(transfer_amount)
+                receive_transfer_account.increment_balance(transfer_amount)
```
enjeyw, line 488: we'll swap these out for the update process with the new balance thing next pr!

PR #260, opened 2020-05-06
Fix: Adding back curl

tristanhcole:
**Describe the Pull Request**
Curl was removed in PR #231 breaking CodeCov upload. Adding it back in.

### DIFF
--- a/app/Dockerfile
+++ b/app/Dockerfile
@@ -1,7 +1,7 @@
 FROM python:3.6-slim-stretch
 
 RUN apt update \
-  && apt -y install gcc g++ libffi-dev libstdc++-6-dev python3-dev musl-dev libssl-dev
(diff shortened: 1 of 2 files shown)


app/Dockerfile
```diff
@@ -28,6 +28,8 @@ COPY .coveragerc /src
 
 WORKDIR /
 
+RUN apt -y install curl
+
```
tristanhcole, line 32: @enjeyw Was there a reason this was removed or just an accident?

enjeyw: accident, but it should be at the start, otherwise it'll kill our caches. I'll fix

tristanhcole:
> Shifted the location - assuming this runs cov, it should be fine now!

Perfect -thanks!

PR #261, opened 2020-05-06
Fix: codecov relative path for server

tristanhcole:
Relative file paths for backend tests code coverage is broken. This fixes that so that files in the `server` directory are viewable. See error: https://codecov.io/gh/teamsempo/SempoBlockchain/src/b8d1bafd5dcc9b05829a2b2777d7519586f9e161/server/constants.py

**Merge in #260 first**


app/Dockerfile
```diff
@@ -28,6 +28,8 @@ COPY .coveragerc /src
 
 WORKDIR /
 
+RUN apt -y install curl
```
enjeyw, line 31: @tristanhcole should this be here since it was added higher in 260?

tristanhcole: Yeah that’s an accident since was a fork of the other branch, will fix

enjeyw: gone and done it from github :)

tristanhcole:
> Good to go now!

Um this should actually be a draft PR

tristanhcole:
I still need to fix a few things

PR #262, opened 2020-05-06
Hardening Blockchain Worker Transactions

enjeyw:
Objective: Provide the highest guarantee possible that there's a 1:1 mapping between db transaction records and Eth transaction records, without needing to fall back to synchronous transactions.

### Enforce more order between when transactions are created in DB and their nonce/ when they're made to chain.
Currently transaction ordering is only explicitly enforced for sets of transactions that are guaranteed to fail if not submitted in order (for example an Approval **must** occur before a transferFrom).

Complete preservation of order would be very inefficient; if we submit txn1 and txn2 into the same block, there's no guarantee that txn1 won't fail with txn2 succeeding. Thus we'd only be able to make one txn per block.

Some weaker versions are:

####  Enforce submission order
Transactions are submitted to blockchain in order, what happens after then is undefined
Pros:
Assuming master account users transferFrom, and nonces should preserve order in most use cases
Cons:
Doesn't work for exchanges or when transactions revert
Hard to do efficiently with lots of distributed workers (though may be premature optimisation to worry about this)

####  Enforce order between transactions involving common users
TX1: userA -> userB
TX2: userB-> userC
TX3: userA -> userD
TX4: userE->userF
txns 1 to 3 must occur in order, txn4 can occur whenever
Pros:
Works inside existing framework of how we manage txn dependencies
Is probably pretty efficient
Cons:
May be complex to determine the dependency tree (or maybe not - I'll investigate)

EDIT: Pulled out other features into issues:
https://github.com/teamsempo/SempoBlockchain/issues/332
https://github.com/teamsempo/SempoBlockchain/issues/331


michielderoos:
Genarally I like this! I have a ton of thoughts, just because all the complexity here.   
## TX Order
### Any Order + ID Order Fallback
By far the biggest complication is around balancing maintanance of transaction order, and performance/paralellism.  
We discussed the idea of processing the transactions on the worker end in an order-agnostic way, and falling back to Tx ID order upon failure. I've run a simulation to see if that's feasable, and unfortunately, there are many situations in which there is no recovery after transaction order is shuffled and previous transactions have been committed. You can see my work [here](https://gist.github.com/michielderoos/90fef2a8dc91a964446515fa52475328) (I really hope I'm wrong, so I hope you can find a mistake!). The silver lining here is that in that gist is also a model we can use to test different strategies we might come up with.  
  
### Trees
There is also the option of having chains/trees-- specific sets of transactions which _must_ execute in order as they are potentially interdependant. This approach has merit, but as we discussed I have concerns around taking that route. If, for example, there is a user who commits failed transacitons, that failure can propegate virally to any account who touches it (I.e. one unreliable actor could crash a major merchant, and all of that merchant's customers). Employing directionality in the failures can help, but does not fully solve the problem. I am also a bit concerned about the complexity this approach might add, since there are a lot of moving parts (how this works in conjunction with retry logic, permutations of different failure modes, etc...).   
  
### Enforced order
One idea I was toying with a bit was just to copy the way real banks work as much as possible (locking individual accounts until a transaction finishes). It kind of combines your `enforce submission order` and `Enforce order between transactions involving common users` suggestions, but also affords distributed workers. The way we can do this, but keep paralellism/distribution, is as follows:
- Ensure processes START being processed by the worker in order of transaction ID (though we have no assurance they'll be sent to the blockchain in order, or resolve in order)  
- As soon as the worker gets a transaction, it creates a redis lock using the sender's user ID, as well as the recipient's user ID.  
    - We will also have to use some mechanism to ensure that these locks are distributed in the same sequence as the transaciton IDs (so that if a user starts two transactions _almost_ at the same time on app-side, the earlier one will gain the lock on worker-side as well)
- When the transaciton completes, the locks are released    
- If there's an error, we need some mechanism to taxonomize errors and determine fault (E.g. recipient issue(`recipient`), insufficent funds (`sender`), outage (`blameless`))
    - On `blameless`, do normal retries. Might need some branching logic for different failure modes there  
    - On `sender`, lock sender's account, send emergency slack to devs since this should never happen (compromised wallet?)
    - On `recipient`, lock recipient's account, send emergency slack to devs since this should never happen (Note: I can't really think of when this would ever happen)

Pros:  
- Sidesteps need for trees-- errors down-stream are avoided by locking known error'd accounts
- Highly performant under normal use (lots of people doing small transactions with each other)

Cons: 
- The workers can potentially get bunged up with several transactions all waiting for the same lock(s) to release.  
    - I think there's a few different strategies to deal with this (E.g. leapfrog over tasks with locked dependencies, create a nested queue ordered by Tx ID to determine first dibs on the lock)
- Would make bulk disbursements super slow. Might have to make a special case where certian wallets are "trusted" and don't lock.   
  
## App-side Retry  
I agree with this point. To expand a bit, maybe we can make the worker track transaction IDs as well, so if for example, it receives transactions `1, 2, 3`, and then ,`5`, it can go back and say `Hey buddy! Where's 4!?`  
  
## More detailed recording of transaction statuses across all stages  
100%. I think we're almost there already.   
  
## Review of worker-side logic  
Agree. There's probably some stuff which can be simplified quite a bit.  
  
## Unit tests  
Yep!

enjeyw:
Nice work on this @michielderoos. I was pretty certain that an iterative 'retry the oldest txn' strategy would work, but ur research pretty much torpedos that idea.

I used ur code to investigate some concrete failure cases and was actually able to come up with simple example where it's impossible to complete the desired transaction set:

```
CORRECT
Step     Bal              Delta
         A    B    C      A    B    C
---------------------------------------
0        10   10   0      *    *    *

1        0    20   0     -10  +10   *

2        10   10   0     +10  -10   *

3        0    10   10    -10   *   +10

4        0    0    20     *   -10  +10


Out of Order
Step     Bal              Delta
         A    B    C      A    B    C      Possible?
----------------------------------------------------
0        10   10   0      *    *    *      *

3        0    10   10    -10   *   +10     Y

4        0    0    20     *  -10   +10     Y

1        0    0    0     -10  +10   *      N

2        0    0    0     +10  -10   *      N


After steps 3, and 4, both A and B have 0 balance, so there's
no possible way to clear the last two trades.
There is a trade that results the same end state
(in this case no trade is required), but matching the desired txn
set is impossible
```

Gonna have to think a little bit more on mitigations/how to enforce order.

michielderoos (approved):
Good to go!

PR #267, opened 2020-05-11
fix(test): pusher not patched in mock_pusher

tristanhcole:
Pusher not patched in `mock_pusher`


tristanhcole:
Frontend tests are failing. Fix in #268

PR #276, opened 2020-05-19
Add "everyone except"  option to credit_transfer API

michielderoos:
When we add search, the FE will no longer have access to every account ID. This is to add the ability for the API to disburse to everyone _except_ given selected people, so a user will still be able to go through pages and uncheck users they don't want to give to.


app/server/api/credit_transfer_api.py
```diff
@@ -231,16 +238,24 @@ def post(self, credit_transfer_id):
...
             transfer_user_list = []
-            for transfer_account_id in recipient_transfer_accounts_ids:
-                try:
-                    individual_sender_user = None
-                    individual_recipient_user = find_user_with_transfer_account_from_identifiers(
-                        None, None, transfer_account_id)
+            individual_sender_user = None
+
+            if invert_recipient_list:
+                all_accounts_query = TransferAccount.query.filter(TransferAccount.is_ghost != True).filter_by(organisation_id=g.active_organisation.id)
```
michielderoos, line 244: Note: For this to work with search, this query (the set we're inverting) will have to be replaced with the original search query. This is just getting the skeleton in place.

tristanhcole:
[Add ability to disburse to everybody except a given list](https://trello.com/c/X2VYV83t/453-add-ability-to-disburse-to-everybody-except-a-given-list)

app/server/api/credit_transfer_api.py
```diff
@@ -198,6 +200,11 @@ def post(self, credit_transfer_id):
         recipient_transfer_account_id = post_data.get('recipient_transfer_account_id')
 
         recipient_transfer_accounts_ids = post_data.get('recipient_transfer_accounts_ids')
+        
+        # invert_recipient_list will send to everyone _except_ for the users in recipient_transfer_accounts_ids 
+        invert_recipient_list = post_data.get('invert_recipient_list', 'false').lower()
+        invert_recipient_list = False if invert_recipient_list == 'false' else True
```
tristanhcole, line 205: Would this make more sense as a boolean?

michielderoos:
@tristanhcole There wasn't a test yet to validate bulk disbursements (beyond singleton lists + `is_bulk`) so I threw that in, and swapped out the quick String boolean thing also

PR #282, opened 2020-05-26
Docker documentation

gtaschuk:
**Docker development setup suggestions**
Adds some documentation/suggested fixes for some problems I ran into gettnig this up locally with `docker-compose up`

One thing I noticed is this project makes heavy use of startup scripts as opposet to one-off services that (like seeding the database, deploying contracts to ganache) - it took a minute for me to figure out how config is loaded from .ini files on startup

I think the current setup is geared towards CI testing, for local development I like to have a second docker-compose file for CI builds (such as `docker-compose -f docker-compose.yml -f docker-compose.circleci.yml up`) if you need to specify additional vars and commands or mock out services

Cool project - found you on linkedin
Greg


enjeyw:
@gtaschuk 

> I think the current setup is geared towards CI testing, for local development I like to have a second docker-compose file for CI builds

I think that's a really good idea. There's waaaay too much junk in the CI compose file that just shouldn't be there.

Generally speaking, docker-compose dev setup is neglected a bit.

There's some nice stuff in this PR - thanks! Tests aren't passing, but I think that's because we seriously messed up CI last week when we swapped to paid Circle, so we'll get those fixes merged into master first and then re-address here.

enjeyw:
think i worked out what the bug is here - none of the eth workers are running through the db migration process...

devtools/install_python_requirements.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-cd ../app
+cd ./app
```
enjeyw, line 3: Currently all of the devtools scripts are set with the expectation that they're run from the devtools directory. For consistency we should keep it the same. I'm not strongly opinionated whether everything should be from root or devtools. Thoughts @michielderoos ?

docker-compose.yml
```diff
@@ -48,7 +49,7 @@ services:
     image: "eth_worker:${GIT_HASH}"
     environment:
       DEPLOYMENT_NAME: "DOCKER_TEST"
-      CONTAINER_TYPE: "PRIMARY"
+      CONTAINER_TYPE: "HIGH_PRIORITY_WORKER"
```
enjeyw, line 52: This is the part that's currently breaking tests, as it's diverting the worker from the default config (no match) to HIGH_PRIORITY.

Suggest changing it back for now and we review worker splits in a seperate PR...

app/Dockerfile
```diff
@@ -16,7 +16,9 @@ RUN cd / && pip install -r requirements.txt && pip install -r test_requirements.
 RUN apt remove --purge -y gcc g++ libffi-dev libstdc++-6-dev python3-dev musl-dev libssl-dev
 
 
-COPY ./config.py ./config_files/* /src/
+COPY ./config.py /src/
+RUN mkdir /src/config_files
+COPY ./config_files/* /src/config_files/
```
enjeyw, line 21: noice

README.md
```diff
@@ -273,3 +273,12 @@ This will extract all keys into the config_files folder. To encrypt them again f
...
 Instead, if you need to merge in two config files, you need to save the old config, load the new one and merge them by hand.
+
+## To Run with Docker
+
+To run with docker first you must generate local secrets that will be read by the `config.py` script that is copied into the `app`, `worker` and `pgbouncer`
+
+```
+python generate_secrets.py -n docker_test
+docker-compose up
+```
```
enjeyw, line 284: 👍

devtools/install_python_requirements.sh
michielderoos: My hunch is that if tools should be run from some directory, it should be their own. I've worked with tools before which store the app directory in an env variable though, which would make it not matter where it's run from. What if we have it read an env, and fall back to `.. /app` if it's not present?

gtaschuk: makes sense - shall i revert this change and change the README to reflect the preferred usage of running it from the directory?

Side note - I wonder if there's a pattern like npm's "script" manifest, or ruby's rake tool that people use for flask.  I ran into this same issue with another flask project I worked on

enjeyw: yeah updating the readme would be perfect!

> Side note - I wonder if there's a pattern like npm's "script" manifest, or ruby's rake tool that people use for flask.

We use `flask_script Manager` for some things, which allows u to do thinkgs like
`manage.py db upgrade`, but so much of the scripting relies on non-python stuff to get things done.

gtaschuk:
@enjeyw this should be good to merge now

tristanhcole:
> @enjeyw this should be good to merge now

Just bumped version, then this is good to merge!

PR #287, opened 2020-06-01
Threaded Location Lookup

enjeyw:
**Describe the Pull Request**

Basic OSM lookup to convert locations into GPS coordinates. Nothing fancy like retries or anything like that, but does the job. Fixes 

**Todo**

- [X] QA your pull request. This includes running the app, login, testing changes etc.


app/server/utils/executor.py
```diff
@@ -0,0 +1,24 @@
...
+    Utility function for adding executor jobs that need to be submitted AFTER a request has completed
+    (generally speaking, to avoid potential race conditions).
+
+    :param fn: the executor decorated function that will be submitted, eg do_foo NOT do_foo.submit
+    :param args: function arguments
+    :param kwargs: function keyword arguments
+    :return:
+    """
+
+    args = args or []
```
michielderoos, line 21: Could you set these default values on lines 8/9 (in place of `None`)?

app/server/utils/location.py
```diff
@@ -0,0 +1,86 @@
...
+
+    :param location: a string address such as '328 Swanston St, Melbourne'
+     or search query such as 'State Library of Victoria'
+    :return: a (latitude, longitude) tuple if the address is found, or none if there's some issue
+    """
+
+    r = _query_osm(location)
+
+    if r.status_code != 200:
+        # OSM should never hit this under normal operation, even with a unrecognised location, so capture error
```
michielderoos, line 58: <3 OSM

app/server/models/user.py
```diff
@@ -260,20 +260,18 @@ def location(self):
...
-            if self.id is None:
-                raise AttributeError('User ID not set')
-
-            try:
-                task = {'user_id': self.id, 'address': location}
-                task_runner('worker.celery_tasks.geolocate_address', args=(task,))
-            except Exception as e:
-                sentry_sdk.capture_exception(e)
-                pass
+            # Delay execution until after request to avoid race condition with db
```
michielderoos, line 268: What is the nature of the race condition?

app/server/models/user.py
```diff
@@ -260,20 +260,18 @@ def location(self):
...
+            # We still need to flush to get user id though
+            db.session.flush()
+            add_after_request_executor_job(
+                async_set_user_gps_from_location,
+                [self.id, location]
```
michielderoos, line 273: Nit: Do you think we should use favour `kwargs` over `args` when using `add_after_request_executor_job` as a best practice? Just think it might be a bit more readable to see exactly what these arguments are in-line. I.e.   
`{'user_id' = self.id, 'location' = location}` explains something about the function signature without having to look up the actual`async_set_user_gps_from_location` function.

app/test_app/unit/utils/test_location.py
```diff
@@ -0,0 +1,26 @@
...
+
+from server.utils.location import (
+    osm_location_to_gps_lookup,
+    _set_user_gps_from_location
+)
+
+@pytest.mark.parametrize("location, expected_coordinates", [
+    ("fooplace", (-37.81, 144.97)),
+    ("multiple matched place", (12.0, 14.4)),
+    ("not a real place", None)
```
michielderoos, line 11: [Oh really?](https://en.wikipedia.org/wiki/Null_Island) 😉

app/server/utils/location.py
```diff
@@ -0,0 +1,86 @@
...
+    Wrapped version for testing
+    """
+    gps_tuple = osm_location_to_gps_lookup(location)
+    if not gps_tuple:
+        logg.warning(f'GPS for location not found on OSM for user {user_id}')
+        return
+
+    lat, lng = gps_tuple
+
+    print(f"gps_tuple {gps_tuple}")
```
michielderoos, line 32: Should delete this before merging

app/server/utils/executor.py
enjeyw: its_a_trap.jpg
https://florimond.dev/blog/articles/2018/08/python-mutable-defaults-are-the-source-of-all-evil/

app/server/models/user.py
enjeyw: the thread needs to load in the user object from the db based off the user id. If that object hasn't been committed yet it won't exist

app/server/models/user.py
enjeyw: yeah  totally fair, will do.

app/server/utils/location.py
enjeyw: ohh yepp

app/test_app/unit/utils/test_location.py
enjeyw: lol!

app/server/utils/executor.py
michielderoos: Huh!

PR #292, opened 2020-06-07
Separate docker-compose into Local dev and CI docker-files

gtaschuk:
**Describe the Pull Request**

Breaks docker compose down by environment (CI vs local dev) and limits the services req'd to test the workers

I'm running into a problem here where the test:backend app process isn't able to pass tasks off to the workers (confirmed it doesn't see any workers with celery task stats from `app/test_app/functional/conftest.py`) even once the `eth_worker` seems to be up and connected to redis.  Any ideas?

```python
+def celery_worker_is_running():
+    ERROR_KEY = "ERROR"
+    try:
+        from celery.task.control import inspect
+        insp = inspect()
+        d = insp.stats()
+        if d:
+            logg.info('running workers found')
+            return True
+        else:
+            logg.info('no running workers found')
+            return False
+    except IOError as e:
+        logg.info('error connecting to the backend')
+        # Error connecting to the backend
+        return False
+    except ImportError as e:
+        logg.info('import error')
+        return False
+    return False
```


enjeyw:
oh wow I had no idea you could override docker-compose ymls like that!
Regarding worker issues - it may be a different issue, but on the CI reports it looks like the configs aren't being set properly. This can happen when the Circle Docker Layer Caches mess up and accidentally pull old images (or Cache Miss and not find any at all)
This is a known issue with Circle:
https://circleci.com/docs/2.0/docker-layer-caching/#limitations

The suggested fix is to use an registry for docker images, but this would require some careful work for it to run with an open source platform - apparently contexts help with this (https://circleci.com/docs/2.0/contexts). Haven't actually looked into it though

Anyways, two fixes we worked with to prevent this:
- building images with git hashes attached so that it's impossible to pull an old image
- using `docker-compose up` rather than `docker-compose run` to run tests, so that docker compose can rebuild from scratch if there's a cache miss.

enjeyw:
Related thought @gtaschuk - would it make sense to expose all the services inside the the dev docker compose to external access? That way if a user wanted to say run the python code outside of docker, but not need to worry about setting up postgres and redis, they could just launch the docker compose and then point to the db... or something like that...

enjeyw:
ok got it working @gtaschuk!
See here:
https://github.com/teamsempo/SempoBlockchain/tree/nick/separate-ci-docker-secrets
Fixes:
- swapped `docker-compose run` to `docker-compose up` - `run` only launches the specified service, so
`docker-compose -f docker-compose.yml -f docker-compose.ci.yml run app` wasn't able to get to the worker because literally only the app service was running

- Added back in the git hashes on the image tags to prevent later circle builds from accidently accessing the wrong images. Renamed the env variable to TAG as it seemed more generally true.

- Forced Ganache-cli to version 6.7.0 Versions higher than that break it. Not sure why - there's nothing in the [release notes to suggest it ](https://github.com/trufflesuite/ganache-cli/releases/tag/v6.8.1)

docker-compose.ci.yml
```diff
@@ -0,0 +1,55 @@
...
+      CIRCLE_PROJECT_USERNAME: ${CIRCLE_PROJECT_USERNAME}
+      CIRCLE_PROJECT_REPONAME: ${CIRCLE_PROJECT_REPONAME}
+      CIRCLE_REPOSITORY_URL: ${CIRCLE_REPOSITORY_URL}
+      CIRCLE_PR_NUMBER: ${CIRCLE_PR_NUMBER}
+      CIRCLE_SHA1: ${CIRCLE_SHA1}
+      CIRCLE_ARTIFACTS: ${CIRCLE_ARTIFACTS}
+      CIRCLE_TEST_REPORTS: ${CIRCLE_TEST_REPORTS}
+  eth_worker:
+    environment:
+      <<: *CI_ENV
```
enjeyw, line 40: Errrrggg I can't believe i didn't know about this syntax earlier. This is SO MUCH CLEANER

docker-compose.yml
```diff
@@ -103,25 +44,20 @@ services:
...
-    build: proxy
-    image: "proxy:${GIT_HASH}"
-    environment:
-      DEPLOYMENT_NAME: "DOCKER_TEST"
-    ports:
-      - "80:80"
-    depends_on:
-      - app
+    image: pgbouncer
+  #proxy:
```
enjeyw, line 48: we need to leave proxy in - even though CI doesn't use it, it needs to be built so it can be pushed to AWS

enjeyw:
`build_docker.sh` in devtools needs to be updated to match

docker-compose.yml
```diff
@@ -130,26 +66,22 @@ services:
...
       POSTGRES_PASSWORD: "password"
 
   redis:
     image: "redis:alpine"
-    environment:
-      DEPLOYMENT_NAME: "DOCKER_TEST"
 
   ganache:
-    image: "trufflesuite/ganache-cli:v6.7.0"
-    command: "-l 80000000 -i 42 ${GANACHE_ACCOUNT_COMMAND}"
```
enjeyw, line 145: The use of `GANACHE_SEED_DETERMINISTIC` is much nicer, but without this when you launch a dev instance in docker, the wallet generated by `generate_secrets.py` won't have access to any funds, as this is where it's fed in.

Three solutions i see:
1) Feed the private key to GANACHE_ACCOUNT_COMMAND as done previously.
Downside is that it's another step devs setting up need to know about
2) Set the dev wallet pk deterministically.
pros: less dev setup
cons: is slightly weird that there's a single dev secret that is deterministic when all the rest are random
2b)Even more so, could just have deterministic secrets/the same for the entire secrets.ini files for all dev/tests. Gets us even closer to one-click dev setup
Pros:
- fully deterministic tests
- even less dev setup
Cons:
- Could be a security risk if someone tries to use those in prod (seems unlikely)
- Could be a security risk if someone tries accidentally pastes in actual secrets into a file that is no-longer git ignored (slightly more likely, but still unlikely) 
- We still need to set custom db credentials, so some amount of manual config wrangling is always required, though config.py could default to pulling from env variables.
3) Just include a boot script in app/or eth_worker that transfers funds from one of the randomly generated wallets to the master wallet when in dev or test mode
pros: we do this in tests already
cons: another conditional logic piece that needs to fire.

gtaschuk:
@enjeyw 

Thanks!  I think this gets pretty close to one command dev setup.  I rebased your fixes in with my work - Should be good to go now!

I guess I was under the mistaken assumption that `docker-compose run` started requisite services.

As discussed in slack - I think using a deterministic ganache setup makes the tests a lot easier to write, and makes it easier to setup a dev environment config because the contracts will be deployed at a predictable address.

PR #294, opened 2020-06-09
Swapping to generic Sempo T&Cs

enjeyw:
**Describe the Pull Request**

Swapping out grassroots specific T&Cs to the standard Sempo ones

**Todo**


enjeyw:
> 👌
> 
> We should also look at configs/settings table for the service codes etc.

A 'make everything generic' task would be great

PR #296, opened 2020-06-11
Eth worker persistence tests

gtaschuk:
**Describe the Pull Request**

A few rather trivial tests of the persistence interface as I learn how the workers function


gtaschuk:
@enjeyw yeah a walkthrough would be useful at some point

I guess the difficulty is that things operate 3 clocks - the system time, the worker time and the blocktime (and maybe another one if you take finality of chain reorgs into account).  So there are a lot of permutations of inter-dependencies to test

PR #311, opened 2020-06-23
Allow Multi-Org GET requests

michielderoos:
We want to be able to provide multiple organizations when requesting metrics, transfer lists, or user lists.   
For example: `/api/v1/credit_transfer/?get_stats=true&org=1,2,3`   
  
Note: Cross-token addition isn't factored in in this PR-- this will be added in the metrics filtering PR


app/server/models/utils.py
```diff
@@ -127,13 +130,18 @@ def filter_by_org(query):
                     # member_organisations = getattr(g, "member_organisations", [])
                     active_organisation = getattr(g, "active_organisation", None)
                     active_organisation_id = getattr(active_organisation, "id", None)
-                    member_organisation_ids = [active_organisation_id] if active_organisation_id else []
+                    # If we're operating on a query supporting multi_org, AND the application
+                    # context has query_organisations set from the HTTP request, use those  
+                    # organizations. Otherwise, use a singleton of the current active org 
+                    query_organisations = [active_organisation_id] if active_organisation_id else []
+                    if multi_org and getattr(g, 'query_organisations', None):
+                        query_organisations = g.query_organisations
```
enjeyw, line 138:
```
if getattr(g, 'query_organisations', None):
     if not multi_org:
            raise Exception("Boo you suck")
     query_organisations = g.query_organisations
```

michielderoos: 👻

app/server/api/user_api.py
```diff
@@ -25,7 +28,7 @@ def get(self, user_id):
 
             if public_serial_number:
                 user = User.query.filter_by(
-                    public_serial_number=public_serial_number.strip()).first().execution_options(multi_org=True)
```
enjeyw: much delete.

app/server/utils/auth.py
```diff
@@ -186,6 +196,20 @@ def wrapper(*args, **kwargs):
...
 
+                        # Check for query_orgs as well. These are stored in g and used for operations which
+                        # are allowed to be run against multiple orgs. Submitted as a CSV
+                        # E.g. GET metrics, user list, transfer list should be gettable with ?orgs=1,2,3
+                        query_organisations = request.args.get('orgs', None)
+                        if query_organisations:
+                            g.query_organisations = []
+                            try:
+                                query_organisations = [int(q) for q in query_organisations.split(',')]
+                                if set(query_organisations).issubset(set(g.member_organisations)):
```
enjeyw, line 207: set syntax is really nice! I've never used it like this before..

PR #312, opened 2020-06-23
Better dashboard filtering

michielderoos:
All Transfer Data filterable according to sender user/transfer account properties  
All User time series filterable by user data only


tristanhcole:
[Split out map onto own page](https://trello.com/c/guKriJtM/470-split-out-map-onto-own-page)

tristanhcole:
Closed by #393

PR #313, opened 2020-06-23
Add time-series metrics

michielderoos:
Add new time-series metrics to metrics endpoint. We want to be able to get average number of trades per user, and average dollar amount of trades per user. We should also be able to get daily active users, and users created. All of these should be able to be fetched on a daily, weekly, or monthly level


tristanhcole:
Closed by #393

PR #317, opened 2020-06-23
Serverside Search Frontend

michielderoos:
Make the frontend use the backend search endpoint, instead of searching using frontend logic


tristanhcole:
Closed by #615

PR #323, opened 2020-06-29
added metabase to dev docker-compose

gtaschuk:
**Describe the Pull Request**

This adds metabase to the dev docker-compose

Metabase is an open-source BI tool that allows you to save SQL queries and create graphs and tables (that update as new data is added) against your database.  I've found it really useful in past projects to quickly answer questions with SQL and create dashboards to track user behavior.  https://www.metabase.com/docs/latest/getting-started.html

Tristan mentioned the other day that there were some custom reports that you were preparing for some NGOs, and I think metabase is great for this sort of thing, because it doesn't require changes to app code, is quite flexible and generates easy to read charts and graphs that you can expose externally with or without authentication.  You should be able to run a single metabase deployment against any/multiple of the postgres databases so if you start the docker service locally you should be able to connect to prod, and see what it's all about

If anyone wants to jump on a call and play around with this, lmk.  If it proves useful we can get it deployed on EBS so that reports are persisted and can be shared.  Metabase maintains it's own config in file storage, or can be connected to (it's own) postgres.  So we could either have 1 metabase instance connected to all the environment databases, or deploy one for each environment.

**Todo**

### DIFF
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -36,3 +36,8 @@ services:
   pgbouncer:
     environment:
       DEPLOYMENT_NAME: "LOCAL_DOCKER"
+  metabase:
+    image: metabase/metabase:latest
+    restart: always
+    ports:
+      - "3000:3000"


enjeyw:
This is really cool @gtaschuk.

One thing we need for it to be useful though is a way to launch it securely on production, ideally using the same auth that we provide for the main system (so people can just log in using the main sempo credentials). Given we use a JWT-esque spec i think it _may_ be possible given https://www.metabase.com/docs/latest/enterprise-guide/authenticating-with-jwt.html, but I'm not 100% certain as i couldn't see how to pass Auth settings into the docker container (also, it'd probably need to jive with our TFA).... any thoughts?

tristanhcole:
Thanks for the PR @gtaschuk - not needed at present

PR #325, opened 2020-06-29
Use a bootstrap entrypoint task to migrate database

gtaschuk:
**Describe the Pull Request**

Putting this up for discussion. Currently the app docker container applies the migrations to the database each time an app container is started, which, while idempotent isn't strictly necessary, and slows container startup. When the server process is started, it also applies the update_data command (which I also believe is idempotent)

This was causing some problems in my local docker dev environment and decoupling the data migration/creation from app startup seems to make things go smoother.

With this update, the app service generally starts with a single command - it may be more direct/conventional to specify the command in the docker-compose and the AWS EBS config https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/single-container-docker-configuration.html but honestly the current shell script works fine and once I saw what it was doing it makes sense
**Todo**

### DIFF
--- a/app/_docker_app_script.sh
+++ b/app/_docker_app_script.sh
@@ -2,9 +2,6 @@
 echo Container mode: $CONTAINER_MODE
 
 cd src
-echo upgrading database
-python manage.py db upgrade
-
 ret=$?
 if [ "$ret" -ne 0 ]; then
   exit $ret
@@ -27,12 +24,12 @@ if [ "$CONTAINER_MODE" == 'TEST' ]; then
 elif [ "$CONTAINER_MODE" == 'ETH_WORKER_TEST' ]; then
    echo pass
    sleep infinity
-else
-
-  echo upgrading dataset
-
+elif [ "$CONTAINER_MODE" == 'BOOTSTRAP' ]; then
+  echo upgrading database
+  echo bootstrapping dataset
+  python manage.py db upgrade
   python manage.py update_data
-
+else
   uwsgi --socket 0.0.0.0:9000 --protocol http  --processes 4 --enable-threads --module=server.wsgi:app --stats :3031 --stats-http
 fi
 
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -2,11 +2,19 @@
 
 version: "3"
 services:
+  bootstrap:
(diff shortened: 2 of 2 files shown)


enjeyw:
Hmmm not certain about this @gtaschuk... it'd mean we'd need to launch an extra container on AWS just to ensure the database is migrated? Right now we have guarantees that any APP container launched is definitely on a fully migrated db (and updated dataset) due to the idempotency of those two scripts...

What's the issue ur having on local dev exactly?

gtaschuk:
> Hmmm not certain about this @gtaschuk... it'd mean we'd need to launch an extra container on AWS just to ensure the database is migrated?

Good points - I guess this trades one problem for another.  Totally fine to close this PR - maybe it will come in handy later.  

Here's what I was thinking when I was working on it: Decoupling schema upgrades, and data bootstrapping from app container startup may become important as you scale and when you are
- concerned about downtime (these migrations do take time each time you startup the app container, even if there is nothing to do).  Maybe EBS is doing health checks before it rolls things over?
- launching multiple containers starting at the same time (in which case parallel updates can lead to inconsistency and database corruption)

For example if you have a load balancer in front of two app containers, *A* and *B*, and you start up 2 new containers *C* and *D* with a migration then *C* and *D* may try to run the migrations and data bootstrapping simultaneously which could corrupt the database in ways that are hard to recover from.  In the meantime, *A* and *B* are running old code against new migrations before the load balancer switches out which containers service the requests.  Ideally the app code in *A & B* would be able to handle the unmigrated and migrated database, so you never have broken requests or downtime before *C & D* are introduced (though ensuring this is cumbersome in practice)

There may be a half step that makes sense, like decoupling the bootstrapping that only has to happen on a new environment from app startup, to reduce downtime, while keeping the migrations.

> 
> What's the issue ur having on local dev exactly?

https://gist.github.com/gtaschuk/21cef27b781e03d378d3dea24d15bfce <- this. Seems to be some race condition between the app's bootstrap process and the workers starting up (and being able to process the wallet creation task when creating the master organization.)  I'm going to dig in deeper to this and see if there might be a way to mitigate this with persistent local dev database volumes, and work out some networking issues with redis brokers (and see if I can get flower working).  I still don't have the 1 command app startup docker environment that I'm after.

tristanhcole:
Outdated

PR #326, opened 2020-06-29
Refactor Metrics Endpoint

michielderoos:
**Describe the Pull Request**
- Adds unit tests  
- Moves metrics to own endpoint (FE and BE)
- DRYify and Objectify individual metrics
- Add metric categories
- Add a transfer filter
- Update filters endpoint to give participant or transfer filters
- Write some basic unit tests for metrics as they were missing before
  
Note: This breaks some filters on the frontend, since per the new spec, we error when we filter on unsupported attributes now (instead of taking no action) 
**Todo**


app/server/utils/metrics/metrics.py
```diff
@@ -0,0 +1,201 @@
...
+
+        try:
+            data['master_wallet_balance'] = cached_funds_available()
+        except:
+            data['master_wallet_balance'] = 0
+
+    return data
+
+
+def cached_funds_available(allowed_cache_age_seconds=60):
```
michielderoos, line 104: This function is a bit rough, but it's literally just moved over from the old path. Probably a worthwhile todo to rewrite it at some point, but just found it out of scope for here

app/server/api/metrics_api.py
```diff
@@ -0,0 +1,89 @@
...
+        Parameters:
+            - metric_type: (Default: ALL) Allows the user to request filters for `transfer`, `participant`, and `all`
+        """
+
+        metric_type = request.args.get('metric_type', metrics_const.ALL)
+        if metric_type not in metrics_const.METRIC_TYPES:
+            raise Exception(f'{metric_type} not a valid type. Please choose one of the following: {", ".join(metrics_const.METRIC_TYPES)}')
+        METRIC_TYPES_FILTERS = {
+            metrics_const.ALL: ALL_FILTERS,
+            metrics_const.PARTICIPANT: PARTICIPANT_FILTERS,
```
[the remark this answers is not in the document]
michielderoos: Yeah, that makes sense. Fixed!

app/server/utils/transfer_filter.py
```diff
@@ -28,7 +29,13 @@ class TransferFilterEnum:
     DISCRETE        = "discrete"
     BOOLEAN_MAPPING = "boolean_mapping"
```
enjeyw, line 30: side note: It'd be nice to formalise the syntax here a little in a later PR. Eg swap this to "[boolean]"

app/server/api/metrics_api.py
```diff
@@ -30,14 +30,20 @@ def get(self):
         start_date = request.args.get('start_date')
         end_date = request.args.get('end_date')
         encoded_filters = request.args.get('params')
-        disable_cache = request.args.get('disable_cache', False)
+        disable_cache = request.args.get('disable_cache', 'False').lower() in ['true', '1']  # Defaults to bool false
```
enjeyw, line 33: this one was always ending up as True because it was a stringystringstring!

app/server/utils/metrics/filters.py
```diff
@@ -19,24 +36,26 @@ def apply_filters(query, filters, query_table):
 
     for filter_table_name, _filts in filters.items():
         if filter_table_name == TransferAccount.__tablename__:
-                # only join transfer account if table is not transfer account
-                if query_table.__tablename__ == filter_table_name: 
-                    query = _apply_single_column_filter(query, _filts, TransferAccount, None, None) 
-                else:
-                    query = _apply_single_column_filter(query, _filts, TransferAccount, account_join_attribute, None)
```
enjeyw: pulled out the last two terms and this became a bit simpler

app/test_app/functional/api/test_metrics_api.py
```diff
@@ -6,6 +6,74 @@
 from server import db
 import json
 
+
+@pytest.fixture(scope='module')
+def generate_metrics(create_organisation):
```
enjeyw, line 11: this seemed like more of a setup fixture than a test in its own right

app/server/utils/metrics/filters.py
michielderoos: This broke the search tests (which also use metrics filters) so I had to roll some of these changes back

PR #329, opened 2020-07-02
fix: codecov notify after build

tristanhcole:
**Describe the Pull Request**
Codecov sometimes accepts a CI build even if a test fails or comments on a PR if a coverage report doesn't upload correctly. This causes the coverage graph to be lumpy. This PR fixes that.

Screen Shot 2020-07-02 at 4 57 09 pm


tristanhcole:
Hmmm codecov checks are lost in this for some reason

tristanhcole:
> Hmmm codecov checks are lost in this for some reason. Will investigate

Looks like it's working now!

PR #335, opened 2020-07-03
Fix double transaction bug

michielderoos:
**Describe the Pull Request**

This PR fixes the issue of balances not adding up under some circumstances. It also adds the concept of a balance offset, as an app-side analogy to minting coins on the blockchain.  
One remaining TODO: 
These changes work fine for new deployments, but we still need some mechanism to move existing deployments' master wallet balance into balance_offset, while still accurately recalculating everyone else's balance. Not quite sure what the cleanest way to do this would be. Perhaps a migration 
**Todo**


app/server/models/transfer_account.py
```diff
@@ -111,18 +117,13 @@ def balance(self):
 
     @balance.setter
     def balance(self, val):
-        self._balance_wei = val * int(1e16)
+        self._balance_offset = val * int(1e16)
```
enjeyw, line 120: I don't think this should be under the balance setter, because it's _not_ actually setting the balance.

Suggest rather just a plan 'ol method called `set_balance_offset` or something, and leave the balance setter undefined as it's not really allowed unless in tests

app/server/models/transfer_account.py
```diff
@@ -29,6 +29,12 @@ class TransferAccount(OneOrgBase, ModelBase, SoftDelete):
 
     name            = db.Column(db.String())
     _balance_wei    = db.Column(db.Numeric(27), default=0)
+    # The purpose of the balance offset is to allow the master wallet to be seeded at
+    # initial deploy time. Since balance is calculated by subtracting total credits from 
+    # total debits, without a balance offset we'd be stuck in a zero-sum system with no 
+    # mechanism to have initial funds. It's essentially an app-level analogy to minting 
+    # which happens on the chain. 
+    _balance_offset    = db.Column(db.Numeric(27), default=0)
```
enjeyw, line 37: Nit: _balance_offset should be _balance_offset_wei

app/server/models/transfer_account.py
```diff
@@ -29,6 +29,12 @@ class TransferAccount(OneOrgBase, ModelBase, SoftDelete):
 
     name            = db.Column(db.String())
     _balance_wei    = db.Column(db.Numeric(27), default=0)
```
enjeyw, line 31: this can be nuked

enjeyw:
Update. This fails, so it's specifically these numbers:

```
@pytest.mark.parametrize("from_token, to_token, from_amount, status_code", [
    # ("smart_token_2", "smart_token_1", 1e5, 400),
    ("reserve_token", "smart_token_1", 1e2, 200),
    # ("smart_token_1", "reserve_token", 1e-10, 200),
    # ("smart_token_1", "smart_token_2", 1e-10, 200),
])
```

app/server/models/transfer_account.py
michielderoos: `_balance_wei` is still used though, it's just set differently

app/server/utils/contract.py
```diff
@@ -73,8 +73,7 @@ def deploy_cic_token(post_data, creating_org=None):
             }
 
             return response_object, 400
-
-        master_org.org_level_transfer_account.balance -= load_amount
+        #master_org.org_level_transfer_account.balance -= load_amount
```
michielderoos, line 76: Commented this just to see what happens

PR #337, opened 2020-07-13
Functionality to Remove Prior Task Dependencies.

enjeyw:
**Describe the Pull Request**

Companion to https://github.com/teamsempo/SempoBlockchain/pull/262.
New transaction hardening enforces a tree-based order between tasks. This could foreseeably cause issues where a task that isn't completing for whatever reason 'poisons' all downstream posterior tasks that are dependant on it. See: https://github.com/teamsempo/SempoBlockchain/pull/262#issuecomment-628790411

This PR introduces functionally to remove these dependencies manually if need be. Two RPC/Celery Tasks are defined, one to remove a specific prior/posterior task pair dependency, and one to remove all dependencies on a given prior task.

Uses the refactored eth worker, so don't bother reviewing until https://github.com/teamsempo/SempoBlockchain/pull/297 is in, as the diff is yuuggge.

**Todo**


app/migrations/versions/09edd72433d4_.py
```diff
@@ -0,0 +1,28 @@
+"""empty message
```
michielderoos, line 1: Tiny Nit: Should have migration message

app/server/models/credit_transfer.py
```diff
@@ -84,23 +86,87 @@ def transfer_amount(self, val):
...
+            this "more_recent_receives"
+
+        Required priors are all transfers in "more_recent_receives" and "most_recent_out_of_batch_send".
+        For why this works, see https://github.com/teamsempo/SempoBlockchain/pull/262
+
+        """
+
+        # We're constantly querying complete transfers here. Lazy and DRY
+        complete_transfer_base_query = (
+            CreditTransfer.query.filter(CreditTransfer.transfer_status == TransferStatusEnum.COMPLETE)
```
michielderoos, line 129: Could we also filter `BlockchainStatus == SUCCESS` here? Otherwise (if I'm reading this correctly), we'd be sending a bunch of priors for tasks we know for sure are already complete from a tx callback.

app/server/models/credit_transfer.py
```diff
@@ -84,23 +86,87 @@ def transfer_amount(self, val):
...
+        # Lastly, edge case handle: if most_recent_out_of_batch_send is a batch member, the whole batch are priors
+        if most_recent_out_of_batch_send and most_recent_out_of_batch_send.batch_uuid is not None:
+            same_batch_priors = complete_transfer_base_query.filter(
+                CreditTransfer.batch_uuid == most_recent_out_of_batch_send.batch_uuid
+            ).all()
+
+            required_priors = required_priors + same_batch_priors
+
+        # Return the set to remove any possible duplicates
+        return set(required_priors)
```
michielderoos, line 167: Nit: Should this be returned in order?

app/server/models/credit_transfer.py
```diff
@@ -84,23 +86,87 @@ def transfer_amount(self, val):
     def send_blockchain_payload_to_worker(self, is_retry=False, queue='high-priority'):
         sender_approval = self.sender_transfer_account.get_or_create_system_transfer_approval()
         recipient_approval = self.recipient_transfer_account.get_or_create_system_transfer_approval()
+
```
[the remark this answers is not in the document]
enjeyw: Oh both will be merged in seperately - Hardening first and then this one - this one just needs the eth worker refactor to go in first.

app/test_app/helpers/mocks.py
```diff
@@ -46,6 +46,20 @@ def await_task_success(*args, **kwargs):
     def retry_task(*args, **kwargs):
         pass
 
+    @staticmethod
```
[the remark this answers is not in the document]
enjeyw: oh nice

app/server/models/credit_transfer.py
enjeyw: nah, because the common edges for a given node on a dependency graph have no order!

app/server/models/credit_transfer.py
enjeyw: Yeah that's true, but I also do like the idea of having the complete dependency graph written down somewhere.
Otoh it does use up data.

app/migrations/versions/09edd72433d4_.py
enjeyw: done

app/server/models/credit_transfer.py
michielderoos: It could be a separate function to get the complete dependency graph if we need it. My thinking around sending one with `BlockchainStatus == SUCCESS` filtered is that, IMO, we should only send the minimum amount of information the worker needs to complete the task. I think sending less data with this would make the task dependencies more readable and easier to debug, be more efficient, and prevent performance bloat and slow-down on long running instances. An extreme example would be a retail store having to check that a candy somebody bought a few years ago successfully cleared.

enjeyw: Fair point! added here: https://github.com/teamsempo/SempoBlockchain/pull/262/files#diff-a140437659d7c51f9bfbfc2afbf84ad1R208

michielderoos (approved):
LGTM!

tristanhcole:
Outdated

PR #394, opened 2020-07-25
Make transfer usage foreign key

michielderoos:
**Describe the Pull Request**

Makes transfer_usage a foreign key (instead of a JSON column)

**Todo**


app/server/utils/metrics/transfer_stats.py
```diff
@@ -70,15 +70,6 @@ def __init__(self, timeseries_unit = 'day'):
             filterable_by=self.filterable_attributes,
             bypass_user_filters=True))
 
-        transfer_use_breakdown_query = db.session.query(CreditTransfer.transfer_use.cast(JSONB),func.count(CreditTransfer.transfer_use)).group_by(CreditTransfer.transfer_use.cast(JSONB))
```
michielderoos, line 73: Killing this stat since this case will be covered by group_by (no need to keep the legacy one in this case)

app/server/utils/credit_transfer.py
```diff
@@ -257,13 +256,11 @@ def make_payment_transfer(transfer_amount,
...
                 if use:
-                    usages.append(use.name)
+                    transfer.transfer_usages.append(use)
                     if use.is_cashout:
                         make_cashout_incentive_transaction = True
                 else:
-                    usages.append('Other')
-
-        transfer.transfer_use = usages
+                    raise Exception(f'{use_id} not a valid transfer usage')
```
michielderoos, line 263: Should we create a generic `other` usage for this case, or is it fair to count on real usages to be used?

app/server/models/credit_transfer.py
```diff
@@ -45,7 +45,13 @@ class CreditTransfer(ManyOrgBase, BlockchainTaskableBase):
     transfer_subtype    = db.Column(db.Enum(TransferSubTypeEnum))
     transfer_status     = db.Column(db.Enum(TransferStatusEnum), default=TransferStatusEnum.PENDING)
     transfer_mode       = db.Column(db.Enum(TransferModeEnum))
-    transfer_use        = db.Column(JSON)
+    transfer_use        = db.Column(JSON) # Deprecated
```
michielderoos, line 48: Keeping this around so data is never accidentally deleted without migration

app/server/utils/credit_transfer.py
enjeyw: I think it's fair to count on the IDs being real IDs. That being said, doing this
```
use_ids = transfer_use.split(',')  # passed as '3,4' etc.
```
shouldn't be the role of this function, but rather at the top of an API or similar.

michielderoos: @enjeyw Fixed!

PR #399, opened 2020-07-27
Make token api return id

nolash:
On token creation response json id is null, as the change is not being committed.


app/server/api/token_api.py
```diff
@@ -67,6 +67,7 @@ def post(self):
 
         token = Token(address=address, name=name, symbol=symbol, decimals=decimals, token_type=token_type)
         db.session.add(token)
+        db.session.commit()
```
michielderoos, line 70: Would `db.session.flush()` work here instead?

michielderoos, line 70: Would `db.session.flush()` work here instead?

tristanhcole: +1

We already have a commit hook for successful requests.

https://github.com/teamsempo/SempoBlockchain/blob/2bc6849be1cfa76ab30e0f7ef158fe1db91d7592/app/server/__init__.py#L127-L131

enjeyw: Arrrgggg, speaking as the person who implemented the auto commit - I hate it. It's magic and weird and results in weird edge cases where we want to commit on a 400 but the hook doesn't catch it.
I'd much prefer to go down @nolash's path and explicity commit when we want a commit. Fits with PEP20 line 2 as well.

nolash: I believe the flush makes it harder to rollback the db during tests?

enjeyw: commit contains a flush AFAIK, to both make it hard to rollback from a db testing point of view! 
That's why we just reset the entire db in the fixture
https://github.com/teamsempo/SempoBlockchain/blob/master/app/test_app/conftest.py#L475

tristanhcole: Yup @enjeyw, https://docs.sqlalchemy.org/en/13/orm/session_basics.html#committing

michielderoos: > results in weird edge cases where we want to commit on a 400 but the hook doesn't catch it  
  
@enjeyw Should we change the `after_request` to `teardown_request`? It's the same as `after_request`, but [also executes if an exception is thrown](https://pythonise.com/series/learning-flask/python-before-after-request).

enjeyw: I think changing to teardown_request makes sense. Filtering by a responses status code is still something I'm iffy on as it means we still need to do edge-case commits inside api's when we want to save the db changes from a 400.

enjeyw: (Equally often we don't - there's no hard and fast rule).

PR #408, opened 2020-07-31
Fix user type text on creation

tristanhcole:
**Describe the Pull Request**
We changed how we handle user type creation, from a URL param to redux form state. Fixes the text/button so that "Add a [ACCOUNT_TYPE]" is consistent.

Screen Shot 2020-07-31 at 4 29 12 pm

**Todo**


tristanhcole:
[Adding Vendor prompts a message mentioning "Beneficiary"](https://trello.com/c/eVI8IJFQ/475-adding-vendor-prompts-a-message-mentioning-beneficiary)

app/client/components/user/CreateUser.tsx
```diff
@@ -30,16 +32,19 @@ interface OuterProps {
   isVendor: boolean;
 }
 
-declare global {
-  interface Window {
-    BENEFICIARY_TERM: string;
```
tristanhcole, line 35: we should really audit how we are handling this across the software - it's not consistent

enjeyw: Agreed that this needs some thought - assuming that setting the term for a beneficiary is distinct from doing translations, what other handling methods do we have? and where?

tristanhcole: Let's resolve this with https://trello.com/c/bGS17Tyz/511-consistent-and-explicit-labelling-of-user-types-across-entire-platform

tristanhcole:
> So the reason this was a url param rather than redux in the first place is that it allowed u to send a direct link to someone with the correct account type set - good for when u've got low-literacy team members and u just want them to fill in a couple of fields.

So is this deal breaker? I could update so the URL query string updates and handles correctly

enjeyw:
> > So the reason this was a url param rather than redux in the first place is that it allowed u to send a direct link to someone with the correct account type set - good for when u've got low-literacy team members and u just want them to fill in a couple of fields.
> 
> So is this deal breaker? I could update so the URL query string updates and handles correctly

Nah I don't think it is a deal breaker, though am subsequently questioning (to ur point i guess) the extent to which this PR aligns with ( https://trello.com/c/bGS17Tyz/511-consistent-and-explicit-labelling-of-user-types-across-entire-platform)

tristanhcole:
> Nah I don't think it is a deal breaker, though am subsequently questioning (to ur point i guess) the extent to which this PR aligns with ( https://trello.com/c/bGS17Tyz/511-consistent-and-explicit-labelling-of-user-types-across-entire-platform)

I think that is a separate PR/audit

enjeyw:
> > Nah I don't think it is a deal breaker, though am subsequently questioning (to ur point i guess) the extent to which this PR aligns with ( https://trello.com/c/bGS17Tyz/511-consistent-and-explicit-labelling-of-user-types-across-entire-platform)
> 
> I think that is a separate PR/audit

Ok makes sense, let's merge this in then

PR #411, opened 2020-08-01
Fix stack trace in log when access ussd endpoint without auth

nolash:
Currently if called, this will utterly spam the log.


tristanhcole:
Outdated

PR #412, opened 2020-08-02
Make user overridable

nolash:
We maintain some diff patches against your code locally to work around elements that aren't modular or extensible. Doing so creates overhead for merges, and also potentially a source of bugs when we don't cherry-pick the changes properly.

This is one example of such.

Grassroots Economics need to extend the user model. I am not sure if this is the best way to do it, but without these changes so far I have not been able to use our ExtendedUser (which just extends the User class) with User intechangeably where User is the type. I believe sqlalchemy or flask or both does some strict type checking to prevent it.

If this change introduces security concerns for you (after all, it skips some checks that otherwise would protect code integrity) I would like find a different way to do it, and would like to hear your proposals. Thanks.

### DIFF
--- a/app/server/models/organisation.py
+++ b/app/server/models/organisation.py
@@ -45,7 +45,9 @@ class Organisation(ModelBase):
     users               = db.relationship(
         "User",
         secondary=organisation_association_table,
-        back_populates="organisations")
+        back_populates="organisations",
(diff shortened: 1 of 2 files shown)


tristanhcole:
Outdated - reopen if still an issue

PR #420, opened 2020-08-06
Label filters as filters

michielderoos:
**Describe the Pull Request**

Filters weren't called filters before... but now they are!

**Todo**


michielderoos:
[Label Filters as Filters](https://trello.com/c/9o1CKwpf/524-label-filters-as-filters)

tristanhcole:
> So I think this isn't quite what we're looking for here - it should be a label in-front of the drop down, like we're doing with the group-by. @tristanhcole opinions?
> Screen Shot 2020-08-07 at 11 48 52 am

I think that we could go with either a Label (photo) or changing the defaultValue (PR), but we should be consistent in approach. Not strongly opinionated either way.

michielderoos:
> So I think this isn't quite what we're looking for here - it should be a label in-front of the drop down, like we're doing with the group-by. @tristanhcole opinions?
> Screen Shot 2020-08-07 at 11 48 52 am

image
@enjeyw Made the change to look like this!

PR #427, opened 2020-08-11
Make volume clearer

michielderoos:
**Describe the Pull Request**

Renames `Volume` to `Payment Volume`. Not the tooltip part of the ticket here, since that's happening in a separate PR

**Todo**

### DIFF
--- a/app/client/components/pages/dashboardPage.jsx
+++ b/app/client/components/pages/dashboardPage.jsx
@@ -149,7 +149,7 @@ class DashboardPage extends React.Component {
                 defaultTimeSeries="all_payments_volume"
                 filterObject="credit_transfer"
                 timeSeriesNameLabels={[
-                  ["all_payments_volume", "Transaction Volume"],
+                  ["all_payments_volume", "Payment Volume"],
                   ["daily_transaction_count", "Transaction Count"],
                   ["users_who_made_purchase", "Unique Users"],
                   ["transfer_amount_per_user", "Volume Per User"],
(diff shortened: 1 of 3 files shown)


michielderoos:
[Make the "Volume" label clearer](https://trello.com/c/EMZlTUgt/516-make-the-volume-label-clearer)

app/client/components/pages/dashboardPage.jsx
```diff
@@ -149,7 +149,7 @@ class DashboardPage extends React.Component {
                 defaultTimeSeries="all_payments_volume"
                 filterObject="credit_transfer"
                 timeSeriesNameLabels={[
-                  ["all_payments_volume", "Volume"],
+                  ["all_payments_volume", "Payment Volume"],
```
tristanhcole, line 152: Does #429 change this? Does Volume mean all transactions on the system or just internal payments? If only internal payments then I think payment volume is better.

michielderoos: @tristanhcole It is all payments! What verbiage would you suggest to use here?

enjeyw: the tooltips _may_ help with this. Part of the reason this one is hard is that `Volume` on its own is concise and correct.

tristanhcole:
I think we should close this for the meantime as (1) I don't think it necessarily makes it clearer and (2) I think we need [consistent labelling of transfers](https://trello.com/c/VHsuCaQ1/517-consistent-labeling-of-transfer-types-across-platform).

PR #428, opened 2020-08-11
Fix Date Range Bugs

michielderoos:
**Describe the Pull Request**

This PR is actually 3 fixes all contributing to the same [problem](https://trello.com/c/KKk3e3xL/530-dateby-filters-on-dashboard-dont-include-their-end-dates)! All of the issues also mean that these issues only crop up _sometimes_. Fun one to figure out!   

Issue 1:  
One frontend would send an un-truncated date to the backend. For example, if you were requesting metrics at 3:00pm with the start date of January 10, then it would request the metrics starting at January 10 at 3:00pm (I.e. append the current time to the past date you are requesting metrics for). This doesn't error because it's technically a valid query, but you're only getting numbers for part of the first day you're requesting metrics. This is resolved here by sending a truncated date object to the API. 
  
Issue 2:  
The frontend has a mechanism to zero-pad dates which have no metrics (I.e. if there are transactions from January 1-10, then none from 11th-20th, and transactions from the 21st to 30th, zero-out the values for dates from 11th to 20th).   
The issue here is that it only pads dates that are in between the raw data's begin and end (so in this case, it would zero-pad the 11th-20th). This becomes a problem if the user has a `start_date` of the 15th. In that case, the graph would still only show the 21st to the 30th.   
This is fixed in the PR by zero-padding up-to the supplied `start_date` and `end_date`, even if there is no data present on/leading up to those dates.   
  
Issue 3:  
The backend calculated end-date by looking at `ModelName.created <= end_date`. The issue here is that `end_date` is a datetime of the given date _at midnight). We actually want midnight of the following day to get stats for that-day. The solution, therefore, was just to add one day to any end_date. 

**Todo**


michielderoos:
[Dateby filters on dashboard don't include their end dates](https://trello.com/c/KKk3e3xL/530-dateby-filters-on-dashboard-dont-include-their-end-dates)

app/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "SempoBlockchain",
-  "version": "1.2.7",
+  "version": "1.2.10",
```
tristanhcole, line 3: skips some verisons?

michielderoos: Lol I based it off another fix branch, will fix for merge.

PR #431, opened 2020-08-14
Transaction Sync Metrics and Recovery Strategies

michielderoos:
**Describe the Pull Request**

Adds a few useful utils for TX sync  
- Metrics endpoint so you can see if tasks are failing from the API (Either a count summary, or full rundown of failed IDs!)
- Endpoint to refetch any block range
- Endpoint to request specific transactions (by hash) to be resubmitted to the webhook

**Todo**


michielderoos:
[Third Party Sync Hardening/Doc](https://trello.com/c/QFJWll31/551-third-party-sync-hardening-doc)

WillRuddick:
Matching a transaction UUID to a transaction hash (or not finding one) is a really lengthy process when you have 350,000+ transactions to go through. I'm generally really worried that there is no 'load' testing on new platform features. 

below is in brief how I found the 65,000 missing blockchain transactions.

cmd_eth = "SELECT tran.blockchain_task_id, tran.hash, tran.updated FROM blockchain_transaction as tran"
            cmd_eth += " WHERE tran._status = 'SUCCESS' ORDER BY tran.id

cmd_eth = "SELECT task.id, task.uuid FROM blockchain_task as task"
            cmd_eth += " WHERE task.status_text = 'SUCCESS' ORDER BY task.id

first just matching above task.id to the latest matching tran.blockchain_task_id

then going through the credit_transfer table (credit_transfer.transfer_status = 'COMPLETE') for all completed transactions ....
and seeing if there is a uuid that has a hash associated with it.

michielderoos:
> Matching a transaction UUID to a transaction hash (or not finding one) is a really lengthy process when you have 350,000+ transactions to go through. I'm generally really worried that there is no 'load' testing on new platform features.
> 
> below is in brief how I found the 65,000 missing blockchain transactions.
> 
> cmd_eth = "SELECT tran.blockchain_task_id, tran.hash, tran.updated FROM blockchain_transaction as tran"
> cmd_eth += " WHERE tran._status = 'SUCCESS' ORDER BY tran.id
> 
> cmd_eth = "SELECT task.id, task.uuid FROM blockchain_task as task"
> cmd_eth += " WHERE task.status_text = 'SUCCESS' ORDER BY task.id
> 
> first just matching above task.id to the latest matching tran.blockchain_task_id
> 
> then going through the credit_transfer table (credit_transfer.transfer_status = 'COMPLETE') for all completed transactions ....
> and seeing if there is a uuid that has a hash associated with it.

That is essentially what [this](https://github.com/teamsempo/SempoBlockchain/pull/431/files#diff-545ad0cbc9430e4fc81a2bb5ebd80088R578) function does. You can save the n^2 operations by using a SQL join as I did in the `get_failed_callbacks` function. This should execute in just a few milliseconds against a 350k row database.

eth_worker/eth_src/sql_persistence/interface.py
```diff
@@ -508,6 +512,76 @@ def check_if_synchronization_filter_exists(self, contract_address, filter_parame
...
+            .group_by(SynchronizationFilter.contract_address)\
+            .filter(SynchronizedBlock.status == 'SUCCESS')\
+            .join(SynchronizationFilter)\
+            .all()
+
+        max_synchronized_blocks = self.session.query(SynchronizationFilter.contract_address, SynchronizationFilter.max_block).all()
+        
+        last_time_synchronized = self.session.query(SynchronizationFilter.contract_address, SynchronizationFilter.updated).all()
+
+        self.get_failed_callbacks()
```
enjeyw, line 549: what does this do? the return doesn't seem to be captured

app/server/api/synchronization_filter_api.py
```diff
@@ -24,8 +24,77 @@ def post(self):
...
+class SynchronizationFilterRecallWebhookAPI(MethodView):
+    @requires_auth(allowed_roles={'ADMIN': 'admin'})
+    def post(self):
+        """
+        Forces a recall of the webhook for a given transaction
+        Params:
+        transaction_hash - Hash for the transaction you want to resync!
+        """
+        transaction_hash = request.args.get('transaction_hash')
+        bt.force_recall_webook(transaction_hash)
```
enjeyw, line 75: would it make sense to have a 'recall all failed webhook' RPC calls?

app/server/api/synchronization_filter_api.py
```diff
@@ -24,8 +24,77 @@ def post(self):
...
 )
+
+synchronization_filter_blueprint.add_url_rule(
+    '/synchronization_filter/status/',
+    view_func=SynchronizationFilterMetricsAPI.as_view('synchronization_filter_metrics_view'),
+    methods=['GET']
+)
+
+synchronization_filter_blueprint.add_url_rule(
+    '/synchronization_filter/refetch_block_range/',
```
enjeyw, line 91: Should these follow the same pattern as the Blockchain Transaction RPC for consistency?
https://github.com/teamsempo/SempoBlockchain/blob/master/app/server/api/blockchain_transaction_api.py#L10

enjeyw (changes requested):
Useful stuff! Few things I think need to be changed...

app/server/utils/blockchain_tasks.py
```diff
@@ -74,6 +74,21 @@ def add_transaction_sync_filter(self, kwargs):
     def force_third_party_transaction_sync(self):
         return task_runner.delay_task(self._eth_endpoint('synchronize_third_party_transactions'), queue='low-priority').id
 
+    def force_fetch_block_range(self, filter_address, floor, ceiling):
+        return self._execute_synchronous_celery(self._eth_endpoint('force_fetch_block_range'), { 'filter_address': filter_address, 'floor': floor, 'ceiling': ceiling })
+
+    def force_recall_webook(self, transaction_hash):
```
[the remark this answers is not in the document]
michielderoos: Yep! Fixed!

eth_worker/eth_src/sql_persistence/interface.py
michielderoos: Good catch!

app/server/api/synchronization_filter_api.py
michielderoos: I left that out on purpose, just because failed webhook calls should be relatively rare. If they fail and are to be retried, it should be a pretty deliberate action IMO. What do you think?

app/server/api/synchronization_filter_api.py
michielderoos: My thinking here is that these calls don't really have to do with a `blockchain_transaction` per-se. What if we rename such that `/synchronization_filter/recall_webhook` becomes `/synchronization_filter_rpc/recall_webhook`?

eth_worker/eth_src/celery_app.py
```diff
@@ -64,9 +67,10 @@
 app.conf.beat_schedule = {
     'third-party-transaction-sync': {
         'task': celery_utils.eth_endpoint('synchronize_third_party_transactions'),
-        'schedule': 30, # Every 30 seconds
+        'schedule': config.THIRD_PARTY_SYNC_CHECK_PERIOD_SECONDS, # Every 5 seconds by default
```
michielderoos, line 70: @enjeyw This might be too frequent-- The action often takes over 30s to do, so the logs already have a lot of `Skipping execution of synchronizing filter 1, as it is already running in another process`. The endpoint is pretty quick to rate-limit an IP, so hitting it too hard will bog down transactions as well.

app/server/api/synchronization_filter_api.py
enjeyw: oh what i meant was the pattern of 
```

        call = post_data.get('call')

        if call == 'RETRY_TASK':
```

eth_worker/eth_src/celery_app.py
enjeyw: ohhh interesting. I was looking for a default of 5 seconds because that's the typical block time on many ethereum chains, and i noticed that the synchronisation was currently getting multiple blocks per check. Is there an easy way to get the action time down? what's the bottleneck?

app/server/api/synchronization_filter_api.py
enjeyw: I can think of a few cases where a lot of webhook calls could fail together due to a common error (for example callback misconfigured, app goes down etc) - having something to retry them would probably be pretty useful no?

eth_worker/eth_src/celery_app.py
michielderoos: I don't think there is any way to speed it up, since the bottleneck is the actual `filter.get_all_entries` call. It might just be slow on sokol, it might well run faster in prod (would be easy to check this in the logs)

michielderoos:
[Alarm on sync fail](https://trello.com/c/Y7eX9SiN/574-alarm-on-sync-fail)

michielderoos:
[Alarm on sync fail](https://trello.com/c/Y7eX9SiN/574-alarm-on-sync-fail)

eth_worker/eth_src/celery_app.py
enjeyw: from memory we actually found (and cleared) a different bottleneck for this in the last few months, though i can't recall what it is....

eth_worker/eth_src/sql_persistence/interface.py
```diff
@@ -508,6 +512,75 @@ def check_if_synchronization_filter_exists(self, contract_address, filter_parame
             return False
         return filter
 
+    def get_transaction_sync_metrics(self):
```
enjeyw, line 515: we'll need to be careful that these don't blow up RDS with expensive queries

enjeyw (approved):
Looks good, assuming queries are indexed ok....

tristanhcole:
[Validate transactions sent by worker match on-chain transactions](https://trello.com/c/AOhsxc88/797-validate-transactions-sent-by-worker-match-on-chain-transactions)

PR #440, opened 2020-08-25
One slash less

nolash:
Trailing slash in add user to org api endpoint, for consistency


enjeyw (approved):
Thanks!

enjeyw:
@nolash that's gotta be the best branch name I've seen in a while:
`nolash:lash/slash`

PR #443, opened 2020-08-28
Improving typescript recompilation time

enjeyw:
**Describe the Pull Request**

Typescript was taking ages to re-compile - this PR splits node modules out into their own chunk so they don't need to be rebuilt whenever we change client code.

**Todo**


app/tsconfig.json
```diff
@@ -12,13 +13,13 @@
...
     // "outFile": "./",                       /* Concatenate and emit output to single file. */
-    "outDir": "./",
+    //    "outDir": "./",
     /* Redirect output structure to the directory. */
     // "rootDir": "./",                       /* Specify the root directory of input files. Use to control the output directory structure with --outDir. */
     // "composite": true,                     /* Enable project compilation */
     // "tsBuildInfoFile": "./",               /* Specify file to store incremental compilation information */
     // "removeComments": true,                /* Do not emit comments to output. */
-    "noEmit": true,
+    //    "noEmit": true,
```
tristanhcole, line 22: Why did you comment out this? Assuming we _don't_ want to emit outputs

tristanhcole (commented):
Seems to be faster 🚀 , but still taking ~`Time: 28979ms` on initial load?

What are your typescript IDE settings @enjeyw ?

tristanhcole (approved):
🚀

app/tsconfig.json
enjeyw: > Seems to be faster 🚀 , but still taking ~`Time: 28979ms` on initial load?
> 
> What are your typescript IDE settings @enjeyw ?

Yeah the initial load won't be much faster, because it still needs to build the vendor package (node modules) the first time  round - the improvement is that those modules are no longer built on later changes, so ongoing deving should be quicker. It's still not as quick as i'd like mind u

enjeyw: > Why did you comment out this? Assuming we don't want to emit outputs

Nah the webpack now uses typescript directly to compile, so we do need the inputs now. Try adding noemit back in - it'll fail!

PR #441, opened 2020-08-27
Consistent naming of user types

enjeyw:
**Describe the Pull Request**

Ensuring that user types are consistently named:
- All users who are subjects of a program are "Participants"
- All users who administer a program are "Admins"

**Todo**

- [x] Link relevant [trello card](https://trello.com/b/PA2LhlOh/sempo-dev) or [related issue](https://github.com/teamsempo/SempoBlockchain/issues) to the pull request
- [x] Request review from relevant @person or team
- [x] QA your pull request. This includes running the app, login, testing changes etc.
- [ ] Bump [backend](https://github.com/teamsempo/SempoBlockchain/blob/e3eb0f480e86d5a8ef2c1814127b70ff018671c4/config.py#L7) and/or [frontend](https://github.com/teamsempo/SempoBlockchain/blob/e3eb0f480e86d5a8ef2c1814127b70ff018671c4/app/package.json#L3) versions following Semver

PR #447, opened 2020-09-02
Hide User Roles

michielderoos:
**Describe the Pull Request**

Adds functionality to grant a user multiple permissions. Also adds ability to remove applicable roles in org settings
image
image
image

Remaining questions:  
Should we use `user` or `beneficiary`?  
Can we alter OrganisationSchema, or does something count on those commented out lines?  

**Todo**


app/server/schemas.py
```diff
@@ -392,15 +392,17 @@ class OrganisationSchema(SchemaBase):
...
     require_transfer_card = fields.Boolean(default=False)
     default_disbursement = fields.Function(lambda obj: int(obj.default_disbursement))
     country_code = fields.Function(lambda obj: str(obj.country_code))
 
     token               = fields.Nested('server.schemas.TokenSchema')
 
-    users               = fields.Nested('server.schemas.UserSchema', many=True)
-    transfer_accounts   = fields.Nested('server.schemas.TransferAccountSchema', many=True)
-    credit_transfers    = fields.Nested('server.schemas.CreditTransferSchema', many=True)
+    #users               = fields.Nested('server.schemas.UserSchema', many=True)
```
michielderoos, line 403: Is there a reason users, transfer accounts, and credit transfers are returned by this API? Was making it run quite slow on a fully loaded db

michielderoos:
[Org Level Config to hide User Roles that aren't being used](https://trello.com/c/Z0mB6Cls/508-org-level-config-to-hide-user-roles-that-arent-being-used)

michielderoos:
[Org Level Config to hide User Roles that aren't being used](https://trello.com/c/Z0mB6Cls/508-org-level-config-to-hide-user-roles-that-arent-being-used)

michielderoos:
[Redesign participant role assignment process](https://trello.com/c/mI4J5y8A/509-redesign-participant-role-assignment-process)

app/client/components/form/MultipleChoice.tsx
```diff
@@ -0,0 +1,22 @@
+import Select from "react-select";
```
enjeyw, line 1: Can we use ant for this instead to avoid a new package (and to begin the migration to ant)? Something like (doesn't work u get the point):
```
import React from "react";
import { Select } from "antd";
const { Option } = Select;

export const MultipleChoice = (props: any) => {
  const { name, input, options } = props;


  const parsedOptions = options.map((option: string) => (
    <Option key={option} value={option}>
      {option}
    </Option>
  ));

  return (
    <Select
      {...input}
      defaultValue={options[0]}
      onChange={value => input.onChange(value)}
      onBlur={() => input.onBlur(input.value)}
    >
      { parsedOptions }
    </Select>
  );
};
```

app/client/components/user/CreateUserForm.tsx
```diff
@@ -72,7 +72,7 @@ class CreateUserForm extends React.Component<
   componentDidMount() {
     const { defaultDisbursement } = this.props;
     this.props.initialize({
-      accountType: TransferAccountTypes.USER.toLowerCase(),
+      accountTypes: [{ value: "USER", label: "USER" }],
```
enjeyw, line 75: Ohhh this will be need to be changed with:
https://github.com/teamsempo/SempoBlockchain/pull/441/files#diff-8cca8a6af685d1a884e93491c72ced3bR2

Transfer account type of "User" shouldn't have been a valid option

app/client/components/user/CreateUserForm.tsx
```diff
@@ -179,30 +180,33 @@ class CreateUserForm extends React.Component<
...
+        <ModuleHeader>Create an account</ModuleHeader>
 
         <div style={{ padding: "1em" }}>
           <form onSubmit={this.props.handleSubmit}>
-            <SelectField
-              name="accountType"
-              label={"Account Type"}
-              options={accountTypes}
-              hideNoneOption={true}
+            <InputField
```
enjeyw, line 204: I know we can technically filter it out, but I don't think this should ever allow me to create an admin account.

app/client/components/transferAccount/types.ts
```diff
@@ -2,6 +2,6 @@ export enum TransferAccountTypes {
   USER = "USER",
```
enjeyw, line 2: To ur comment, "USER" should be nuked - it's "BENEFICIARY" in this case. Incoming change here:
https://github.com/teamsempo/SempoBlockchain/pull/441/files#diff-8cca8a6af685d1a884e93491c72ced3bR2

app/server/constants.py
```diff
@@ -13,13 +13,16 @@
 POSSIBLE_GROUP_ACCOUNTS = ['group_account']
 
 ACCESS_ROLES = {
+    'USER': [],
```
enjeyw, line 16: shouldn't exist

app/server/models/organisation.py
```diff
@@ -156,7 +160,7 @@ def __init__(self, token=None, is_master=False, **kwargs):
     
         self.external_auth_username = 'admin_'+ self.name.lower().replace(' ', '_')
         self.external_auth_password = secrets.token_hex(16)
-
+        self.valid_roles = list(ACCESS_ROLES.keys())
```
enjeyw, line 163: suggest changing the init to
```
def __init__(self, token=None, is_master=False,  valid_roles=None,**kwargs):
```
and this line to
```
self.valid_roles = valid_roles or list(ACCESS_ROLES.keys())
```

Otherwise if a user provides valid_roles as a kwarg it'lll be overwritten by this and the user will be confused

app/client/components/user/CreateUserForm.tsx
enjeyw: or we could just have it removed from the default i guess.

enjeyw (changes requested):
Oh that's super annoying, Github delete my review comments. Ok adding two more now:
- how do i choose a particular tier for my user (for example I can't work out how to add a subvendor/cashier)
- formatting issue that'll hopefully be fixed by ant:
Screen Shot 2020-09-09 at 5 33 54 pm

app/server/schemas.py
michielderoos: With these fields commented out, hitting an org endpoint is pretty much instant, but when any of them is uncommented the endpoints become unusably slow. Like on my local instance with demo data, when I have just the `transfer_accounts` line uncommented, it takes 23s to load and it returns a 29MB JSON payload containing every single transfer

enjeyw (approved):
Looks good! Merging in

PR #457, opened 2020-09-10
Kobo/Webhook Put Functionality

enjeyw:
**Describe the Pull Request**

Adds an argument on the user POST endpoint data to be submitted in "PUT" mode, where if the user id is already on the system, the form updates the existing user rather than throwing an error. Useful for when we're using a webhook that only supports POSTs. Defaults to this POST only, because we don't want users to be accidentally overwritten if eg, someone uses the same transfer card twice by accident.

Adds an argument on user POST endpoint data for whether to return the raw input data on an error - useful for diagnosing what a webhook actually tried to submit

Also includes slight improvements to submitting location data via GPS - mainly, if a gps location is explicity provided, it wont be overwritten by the (Generally less specific) location provided by text lookup.

**Todo**

- [X] Link relevant [trello card](https://trello.com/b/PA2LhlOh/sempo-dev) or [related issue](https://github.com/teamsempo/SempoBlockchain/issues) to the pull request
- [X] Request review from relevant @person or team
- [X] QA your pull request. This includes running the app, login, testing changes etc.
- [X] Bump [backend](https://github.com/teamsempo/SempoBlockchain/blob/e3eb0f480e86d5a8ef2c1814127b70ff018671c4/config.py#L7) and/or [frontend](https://github.com/teamsempo/SempoBlockchain/blob/e3eb0f480e86d5a8ef2c1814127b70ff018671c4/app/package.json#L3) versions following Semver


enjeyw:
https://trello.com/c/CxcdZUjE

enjeyw:
[KYC info via KoboToolbox API](https://trello.com/c/CxcdZUjE/305-kyc-info-via-kobotoolbox-api)

app/client/components/user/EditUserForm.tsx
```diff
@@ -365,7 +365,7 @@ class EditUserForm extends React.Component<
                 <TopRow>
                   <ModuleHeader>OTHER ATTRIBUTES</ModuleHeader>
                 </TopRow>
-                <Row style={{ margin: "0em 1em" }}>
+                <Row style={{ margin: "0em 1em", flexWrap: "wrap" }}>
```
enjeyw, line 368: when lots of attributes were submitted, they got cut off

app/server/models/user.py
```diff
@@ -260,17 +260,19 @@ def location(self):
 
     @location.setter
     def location(self, location):
-        from server.utils.location import async_set_user_gps_from_location
 
         self._location = location
```
enjeyw, line 264: location now doesn't automagically call the async lookup as we don't want it to always happen. Now call `attempt_update_gps_location` below

app/server/utils/geolocation.py
```diff
@@ -1,34 +0,0 @@
-import googlemaps
```
enjeyw, line 1: turns out this hasn't been used in ages.

app/test_app/functional/api/test_user_api.py
```diff
@@ -34,18 +33,18 @@ def submit(*args, **kwargs):
...
+@pytest.mark.parametrize("user_phone_accessor, phone, use_card, business_usage_name, gps_location, referred_by, "
                          "initial_disbursement, tier, status_code",
                          [
-                             (lambda o: o.phone, None, False, 'Fuel/Energy', '+61401391419', None, 'superadmin', 400),
-                             (lambda o: o.phone, None, True, 'Fuel/Energy', '+61401391419', None, 'superadmin', 200),
-                             (lambda o: o.phone, fake.msisdn(), False, 'Fuel/Energy', fake.msisdn(), 0, 'superadmin', 200),
-                             (lambda o: o.phone, fake.msisdn(), False, 'Fuel/Energy', fake.msisdn(), None, 'superadmin', 200),
-                             (lambda o: o.phone, fake.msisdn(), False, 'Food/Water', fake.msisdn(), None, 'view', 403)
+                             (lambda o: o.phone, None, False, 'Fuel/Energy', None, '+61401391419', None, 'superadmin', 400),
+                             (lambda o: o.phone, None, True, 'Fuel/Energy', '12.02 -15.04 otheruselessdata!##', '+61401391419', None, 'superadmin', 200),
```
tristanhcole, line 40: could be good to add a test that has invalid lat lng coords... i.e. `12.01 banana 12.1`

tristanhcole (approved):
LGTM! 👍

app/test_app/functional/api/test_user_api.py
enjeyw: there's one like that in the unit test :)

PR #464, opened 2020-09-15
Adding Secondary Users to Transfer Accounts and Cards

enjeyw:
**Describe the Pull Request**

We need a way to show secondary user names on touch-to-pay payments. This achieves this by adding the capacity to bind secondary users to transfer accounts, and then by adding those users on the authorised card names

**Todo**

### DIFF
--- a/app/server/api/transfer_card_api.py
+++ b/app/server/api/transfer_card_api.py
@@ -13,7 +13,12 @@ class TransferCardAPI(MethodView):
     @requires_auth
     def get(self):
 
-        transfer_cards = TransferCard.query.all()
+        only_loaded = request.args.get('only_active', 'true').lower() == 'true'
+
+        if only_loaded:
+            transfer_cards = TransferCard.query.filter(TransferCard.amount_loaded_signature != None).all()
+        else:
+            transfer_cards = TransferCard.query.all()
 
         response_object = {
             'message': 'Successfully loaded transfer_cards',
--- a/app/server/utils/user.py
+++ b/app/server/utils/user.py
@@ -692,7 +692,9 @@ def proccess_create_or_modify_user_request(
             first_name=attribute_dict.get('secondary_user_first_name'),
             last_name=attribute_dict.get('secondary_user_last_name'),
             phone=attribute_dict.get('secondary_user_phone'),
-            existing_transfer_account=user.default_transfer_account
+            existing_transfer_account=user.default_transfer_account,
+            is_beneficiary=is_beneficiary, is_vendor=is_vendor,
+            is_tokenagent=is_tokenagent, is_groupaccount=is_groupaccount
         )
 
     if phone:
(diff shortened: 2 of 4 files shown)


app/server/api/transfer_card_api.py
```diff
@@ -13,7 +13,12 @@ class TransferCardAPI(MethodView):
     @requires_auth
     def get(self):
 
-        transfer_cards = TransferCard.query.all()
+        only_loaded = request.args.get('only_active', 'true').lower() == 'true'
+
+        if only_loaded:
+            transfer_cards = TransferCard.query.filter(TransferCard.amount_loaded_signature != None).all()
+        else:
+            transfer_cards = TransferCard.query.all()
```
michielderoos, line 21: This issue wasn't introduced in this PR, but should we switch this to `paginate_query` at some point?

app/server/utils/user.py
```diff
@@ -687,6 +687,16 @@ def proccess_create_or_modify_user_request(
     # Location fires an async task that needs to know user ID
     db.session.flush()
 
+    if attribute_dict.get('secondary_user_first_name'):
+        create_transfer_account_user(
+            first_name=attribute_dict.get('secondary_user_first_name'),
+            last_name=attribute_dict.get('secondary_user_last_name'),
+            phone=attribute_dict.get('secondary_user_phone'),
+            existing_transfer_account=user.default_transfer_account,
+            is_beneficiary=is_beneficiary, is_vendor=is_vendor,
```
michielderoos, line 696: Should also have `roles_to_set` ( from #447 ) IMO

app/server/utils/user.py
```diff
@@ -687,6 +687,16 @@ def proccess_create_or_modify_user_request(
     # Location fires an async task that needs to know user ID
     db.session.flush()
 
+    if attribute_dict.get('secondary_user_first_name'):
+        create_transfer_account_user(
```
michielderoos, line 691: Should the secondary user have an accompanying `set_custom_attributes` like the primary user does?

michielderoos (changes requested):
Looks good! Just a few little notes

app/server/utils/user.py
enjeyw: Oh yep good idea, will add in

app/server/utils/user.py
enjeyw: I decided against it, because the custom attributes can really only be inferred about the primary user (for example we have no idea if user two is female or not).

app/server/api/transfer_card_api.py
enjeyw: Yeah I thought about this too, but I think given the way transfer cards are loaded onto phones is quite different (there's potentially location based sharding etc, it made more sense to do as a seperate PR)

tristanhcole:
[Show multiple users on tap of card](https://trello.com/c/cYw6k0wF/545-show-multiple-users-on-tap-of-card)

tristanhcole:
Closing for now

PR #499, opened 2020-10-16
Sharding for transfer cards

michielderoos:
**Describe the Pull Request**

Adds server-side sharding for transfer cards!

**Todo**


michielderoos:
[Optional Shard Cards by location](https://trello.com/c/48j5Edux/636-optional-shard-cards-by-location)

michielderoos:
[Optional Shard Cards by location](https://trello.com/c/48j5Edux/636-optional-shard-cards-by-location)

enjeyw (changes requested):
Works well overall!

Couple of small comments:

- "Card Shard Distance" could be a little intimidating for users to show on front end. Not sure what the better word is... maaaayyybbee "Auto load cards within"

- Comment in the code to say it's in Km would be helpful

- Alchemy doesn't escape code from memory, so for injection reasons i think it's probably good practice to bind params... i think like this?  https://docs.sqlalchemy.org/en/13/core/tutorial.html#using-textual-sql Alternatively see my pure alchemy formulation below

- would be good to include perfect matched location names in shard, so we can still return even if lookup doesn't work

finally (least important)... i was curious if using a pure sqlalchemy query would make it faster due to not needing the second query... it was, but only ever so slightly... The code is kinda nice though and it gives us baked-in querying... thoughts?

```
@hybrid_method
    def great_circle_distance(self, lat, lng):
        """
        Tries to calculate the great circle distance between
        the two locations in km by using the Haversine formula.

        """
        import math

        return self._haversine(math, self, lat, lng)

    @great_circle_distance.expression
    def great_circle_distance(cls, lat, lng):

        from sqlalchemy.sql.functions import func

        return cls._haversine(func, cls, lat, lng)


    @staticmethod
    def _haversine(lib, selfref, lat, lng):

        return 6371 * lib.acos(
            lib.cos(lib.radians(selfref.lat))
            * lib.cos(lib.radians(lat))
            * lib.cos(lib.radians(selfref.lng) - lib.radians(lng))
            + lib.sin(lib.radians(selfref.lat))
            * lib.sin(lib.radians(lat))
        )

    def get_users_within_radius(self, radius):

        if not (self.lat or self.lng):
            raise Exception('Cannot get users within radius-- User location undefined')

        return db.session.query(User).filter(
            or_(
                User.great_circle_distance(self.lat, self.lng) < radius,
                and_(User.lat==None, User.lng==None),
                and_(self._location is not None, User._location == self._location)
            )
        ).all()
```

michielderoos:
> * "Card Shard Distance" could be a little intimidating for users to show on front end. Not sure what the better word is... maaaayyybbee "Auto load cards within"   

Fixed!
> * Comment in the code to say it's in Km would be helpful  

Fixed!
> * Alchemy doesn't escape code from memory, so for injection reasons i think it's probably good practice to bind params... i think like this?  https://docs.sqlalchemy.org/en/13/core/tutorial.html#using-textual-sql Alternatively see my pure alchemy formulation below  
 
Figured this was fine, since it was taking in an integer direct from the database. Did change to new recipe though!   

> * would be good to include perfect matched location names in shard, so we can still return even if lookup doesn't work
  
Already done! Did add a unit test to confirm though 👍   

> finally (least important)... i was curious if using a pure sqlalchemy query would make it faster due to not needing the second query... it was, but only ever so slightly... The code is kinda nice though and it gives us baked-in querying... thoughts?
> 
> ```
> @hybrid_method
>     def great_circle_distance(self, lat, lng):
>         """
>         Tries to calculate the great circle distance between
>         the two locations in km by using the Haversine formula.
> 
>         """
>         import math
> 
>         return self._haversine(math, self, lat, lng)
> 
>     @great_circle_distance.expression
>     def great_circle_distance(cls, lat, lng):
> 
>         from sqlalchemy.sql.functions import func
> 
>         return cls._haversine(func, cls, lat, lng)
> 
> 
>     @staticmethod
>     def _haversine(lib, selfref, lat, lng):
> 
>         return 6371 * lib.acos(
>             lib.cos(lib.radians(selfref.lat))
>             * lib.cos(lib.radians(lat))
>             * lib.cos(lib.radians(selfref.lng) - lib.radians(lng))
>             + lib.sin(lib.radians(selfref.lat))
>             * lib.sin(lib.radians(lat))
>         )
> 
>     def get_users_within_radius(self, radius):
> 
>         if not (self.lat or self.lng):
>             raise Exception('Cannot get users within radius-- User location undefined')
> 
>         return db.session.query(User).filter(
>             or_(
>                 User.great_circle_distance(self.lat, self.lng) < radius,
>                 and_(User.lat==None, User.lng==None),
>                 and_(self._location is not None, User._location == self._location)
>             )
>         ).all()
> ```

Way cleaner, added this!

enjeyw (approved):
Looks good!

PR #509, opened 2020-10-22
Add country to location lookup

michielderoos:
**Describe the Pull Request**

Adds country to osm location lookup if it's not already there!  

**Todo**


app/server/models/organisation.py
```diff
@@ -81,6 +81,10 @@ def timezone(self, val):
     def country_code(self):
         return self._country_code
 
+    @hybrid_property
+    def country(self):
+        return ISO_COUNTRIES[self._country_code]
```
enjeyw, line 86: in know we do it when setting the country code, but wondering if we should also validate the dict lookup here so the system doesn't 500 if the country code doesn't exist

michielderoos: 👍

enjeyw (approved):
LGTM!

PR #512, opened 2020-10-23
Move search indexes

michielderoos:
**Describe the Pull Request**

Search indexes were being refreshed in a very very inefficient way. This is orders of magnitude faster and cleaner!

**Todo**


app/migrations/env.py
```diff
@@ -26,7 +26,7 @@
 
 # Tables to ignore when running 'manage.py db migrate' are defined here
 def include_object(object, name, type_, reflected, compare_to):
-    if type_ == 'table' and name in ('search_view'):
+    if type_ == 'table' and name in ():
```
michielderoos, line 29: Left the guts here, since it's potentially useful in the future

enjeyw (commented):
Hmmm I've been doing a bunch of research on Postgres Searching and I'm not fully convinced ts_vector is the right way to go.

It seems to me that ts_vector is good for longer bodies of text - sentences, comments or even full docs, where there's no spelling errors - ts_vector only exact matches words (accounting for tense etc), and then calculates weighted rank based on the number of exact matches.

In contract, where as what we're dealing with here is single words, for which we have TONNES of spelling errors (names that are spelt differently etc). So for this, it looks like pg_trgm is a better fit. See:
https://about.gitlab.com/blog/2016/03/18/fast-search-using-postgresql-trigram-indexes/
and
https://stackoverflow.com/questions/43867449/optimizing-a-postgres-similarity-query-pg-trgm-gin-index

I do quite like using the ts_vector to get some weighting, but it is abusing the feature slightly, and given we don't actually need to ts_vector phone numbers or blockchain addresses (the alg does nothing too them), we could probably just use normal pg maths to create the ranks.

Alsoalso, we need to be able to search on custom attributes - currently I'm manually looking up "business name" in the database and then doing the join by hand, so i still think we need a materialized view...

michielderoos:
[Optimize search index](https://trello.com/c/YLoSasZn/656-optimize-search-index)

PR #517, opened 2020-10-26
Support updated mobile pin usage

michielderoos:
**Describe the Pull Request**

The mobile app used the password as pin, but this allows it to use pin instead.  
This also makes one-time-code reset work! 
image

Note: This is required to make [this](https://github.com/teamsempo/SempoMobileApp/pull/92) work, but this PR will not break old mobile versions

**Todo**


app/migrations/versions/faef0f3afb43_.py
```diff
@@ -0,0 +1,24 @@
...
+
+# revision identifiers, used by Alembic.
+revision = 'faef0f3afb43'
+down_revision = '2f8c71039480'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.execute('update "user" set pin_hash = password_hash where pin_hash is null;')
```
tristanhcole, line 20: ⚡

tristanhcole (commented):
Migration code makes sense to me.

One caveat - while running https://github.com/teamsempo/SempoMobileApp/pull/92/files on mobile, I couldn't get the "Reset PIN" functionality on the dashboard to actually work. I'm wondering if this should be in scope of this PR or it would be separate?

Finally, I think @enjeyw should also review this PR as core to the auth flow and don't want to cause issues.

michielderoos:
> Migration code makes sense to me.
> 
> One caveat - while running https://github.com/teamsempo/SempoMobileApp/pull/92/files on mobile, I couldn't get the "Reset PIN" functionality on the dashboard to actually work. I'm wondering if this should be in scope of this PR or it would be separate?
> 
> Finally, I think @enjeyw should also review this PR as core to the auth flow and don't want to cause issues.

@tristanhcole Yea-- I thought the "Reset PIN" fix was [this](https://trello.com/c/qIrO1HyA/616-add-pin-reset-for-android-on-admin-dashboard) ticket (out of scope)

michielderoos:
[Add PIN reset for android on admin dashboard](https://trello.com/c/qIrO1HyA/616-add-pin-reset-for-android-on-admin-dashboard)

michielderoos:
@tristanhcole I scoped [this](https://trello.com/c/qIrO1HyA/616-add-pin-reset-for-android-on-admin-dashboard) into this ticket, ready for fresh 👀

app/server/api/user_api.py
```diff
@@ -210,10 +210,13 @@ def post(self, user_id):
...
             user = User.query.get(reset_user_id)
-
             if user is None:
                 return make_response(jsonify({'message': 'No user found for ID: {}'.format(reset_user_id)})), 404
 
+            user.set_one_time_code(None)
+            user.password_hash = None
+            user.pin_hash = None
+
             UserUtils.admin_reset_user_pin(user)
```
tristanhcole, line 220: I think this new code should be in `UserUtils.admin_reset_user_pin` function

tristanhcole: Additionally, I think it would be nice to have the correct SMS one-time code send to the user if SMS is enabled on the instance.

Right now it sends `Dial *384*96# Safaricom or *483*46# Airtel to change your PIN`, which is valid for USSD users. Not sure how we would specify that maybe @enjeyw has some thoughts?

tristanhcole (approved):
Thanks @michielderoos. Core flow and migrations works and LGTM!

Given it's core to the auth flow, I think a second pair of eyes on this would be good (paging @enjeyw ).

app/test_app/functional/api/test_mobile_auth_api.py
```diff
@@ -43,7 +44,7 @@ def test_request_api_token_golden_path_success(
 
     set_pin_response = test_client.post('/api/v1/auth/reset_password/',
                                         data=json.dumps(
-                                            dict(phone=user_phone, new_password='1234', one_time_code=code)),
+                                            dict(phone=user_phone, new_pin='1234', one_time_code=code)),
```
enjeyw, line 47: could we pls add one more step to the test here to verify that the new pin works?

app/server/api/user_api.py
enjeyw: Yeah agree with tristan that the lines above should just be shifted inside! TBH I think just disabling the sms is fine for now.

app/server/api/user_api.py
```diff
@@ -210,10 +210,13 @@ def post(self, user_id):
 
         if reset_user_id is not None:
             user = User.query.get(reset_user_id)
-
             if user is None:
                 return make_response(jsonify({'message': 'No user found for ID: {}'.format(reset_user_id)})), 404
 
+            user.set_one_time_code(None)
```
enjeyw, line 216: i thought resetting pin was re-applying a new on-time-code?

app/server/api/user_api.py
```diff
@@ -210,10 +210,13 @@ def post(self, user_id):
...
+            user.password_hash = None
```
enjeyw, line 217: probably shouldn't reset the pw hash...

app/server/api/user_api.py
```diff
@@ -210,10 +210,13 @@ def post(self, user_id):
 
         if reset_user_id is not None:
```
enjeyw, line 211: this endpoint is about to become a lot more critical - would be goood to have a test for it

enjeyw (commented):
few things that i think need tweaking...

app/server/api/user_api.py
michielderoos: It does-- when supplied_one_time_code is `None`, it generates a new one

app/test_app/functional/api/test_mobile_auth_api.py
michielderoos: Yep!

app/server/api/user_api.py
michielderoos: Beefed up the test for this a bit more!

enjeyw (changes requested):
Now that this isn't resetting the password hash, it's no longer trying lines
```
  if post_data.get('phone') and user and user.one_time_code and (not user.is_activated or no_password_or_pin_hash):
            # vendor sign up with one time code or OTP verified
            if user.one_time_code == password:
 ```

but instead failing on
```
            if not (user and (pin and user.verify_pin(pin) or password and user.verify_password(password))):
```
with a non-existant pin

**Review by michielderoos — COMMENTED** [review]
@enjeyw I can't replicate that exact bug, but I did get something similar which I've resolved when dealing with a non-existent account. Which version of the FE were you using?

**Comment by enjeyw** [comment]
@michielderoos i just tested with new branch. i still get it. To clarify the issue that im seeing:
because on pin reset the flow `should` hit 
```
            # vendor sign up with one time code or OTP verified
            if user.one_time_code == password:
```
to verify if the otc and the password are the same, but instead because 
` (not user.is_activated or no_password_or_pin_hash)` doesn't trigger (the user is activate, and they still have a password has), it instead behaves as if it's expecting a password, and then fails with this:

  File "/Users/Nick/PycharmProjects/SempoBlockchain/app/server/models/user.py", line 454, in check_salt_hashed_secret
    hashed_password = f.decrypt(hashed_password.encode())
AttributeError: 'NoneType' object has no attribute 'encode'

this somewhat fixes it, but i still get a missing token error

```
        if post_data.get('phone') and user and user.one_time_code and (not user.is_activated or not user.pin_hash):
            # vendor sign up with one time code or OTP verified
            if user.one_time_code == pin:
                response_object = {
                    'status': 'success',
                    'pin_must_be_set': True,
                    'message': 'Please set your pin.'
                }
                return make_response(jsonify(attach_host(response_object))), 200
```

enjeyw (approved):
sweet works for me! mobile too :)

PR #521, opened 2020-10-27
Fix export api

michielderoos:
**Describe the Pull Request**

Would load forever trying to get `g.user.transfer_account`, I think this logic was created before a user could have multiple transfer accounts. Changed to make it get the primary transfer account!

**Todo**


tristanhcole:
[Change me_export_api to use right transfer account](https://trello.com/c/J6q6hWfr/660-change-meexportapi-to-use-right-transfer-account)

tristanhcole (approved):
LGTM!

enjeyw (changes requested):
```
            for index, credit_transfer in enumerate(credit_transfer_list):
```
Just loading this line (not even the loop) still takes fooorrreeevvvveerrrr on my computer if i load in lots of data! I think it's what's timing out on the server.

Setting ```.enable_eagerloads(False)``` on the previous lines makes things better, but I haven't checked if there's any further optimisations by loading in transfer usages....

michielderoos:
> ```
>             for index, credit_transfer in enumerate(credit_transfer_list):
> ```
> 
> Just loading this line (not even the loop) still takes fooorrreeevvvveerrrr on my computer if i load in lots of data! I think it's what's timing out on the server.
> 
> Setting `.enable_eagerloads(False)` on the previous lines makes things better, but I haven't checked if there's any further optimisations by loading in transfer usages....

@enjeyw I found it about 3x faster with `.enable_eagerloads(True)` rather than ` .enable_eagerloads(False)`. One thing which I think will make it way faster though, is #525  since it adds indexes on `recipient_transfer_account_id` and `sender_transfer_account_id`

app/server/api/export_api.py
```diff
@@ -451,19 +452,19 @@ def post(self):
...
 
         if credit_transfer_list is not None:
-            file_url = export_workbook_via_s3(wb, workbook_filename, email)
+            #file_url = export_workbook_via_s3(wb, workbook_filename, email)
 
             response_object = {
                 'status': 'success',
                 'message': 'Export file created.',
-                'file_url': file_url,
+                'file_url': 'a',
```
enjeyw, line 467: doi

enjeyw (commented):
That line is still craaazzzy slow for me if i use a large dataset... i think it's cos credit_transfers joins on transfer accounts and a bunch of other things we don't use?

enjeyw (approved):
LGTM!

PR #541, opened 2020-11-09
[Spec] Add comments to Transfer Accounts

michielderoos:
# Goals: 
- Ability to leave individual comments on transfer accounts   
- Be prompted to add a comment when enabling/disabling a transfer account. We want to be able to look at a transfer account and have a record of why it's been disabled. 
- Ability to view list of comments on the Transfer Account page  
  
# Means:  
### Data Model
- We want a one-to-many relationship between an individual transfer account, and comments  
    - The comment should have the following columns:  
        - Created/Updated
        - Comment Text - The actual comment!  
        - Comment Type - This would be an enum of what the comment connotes. This is to distinguish between standard text comments, and comments associated with account enabling/disabling. This could also be extended when card disabling is added.  
### API  
- We should modify the Transfer Account API to add a comment when an account is enabled or disabled. If no comment is provided, leave a blank comment. This way the comments list also serves as a chronology of when a transfer account was enabled and disabled  
- The Transfer Account API should include a list of comments  
- There should also be a dedicated endpoint which lets admins add comments to a users' account. This should only allow standard text comments, and not ones with comment types associated with enabling/disabling accounts  
  
### Transfer Accounts Page  
- Transfer accounts page should use ANT Design's [comments module](https://ant.design/components/comment/) to show a list of comments, as well as a text field allowing admins to add new comments to an account.  
- When a transfer account (or multiple transfer accounts) are selected and Approve or Unapproved are clicked, instead of immediately doing the action, a module should appear above the textbox (similar to when making a disbursement) giving the user the option to write a comment to attach to the account(s) they're modifying. This same action should also be done when enabling/disabling an account from an individual transfer accounts' page.

### DIFF
--- a/config.py
+++ b/config.py
@@ -5,7 +5,7 @@ env_loglevel = os.environ.get('LOGLEVEL', 'DEBUG')
 logging.basicConfig(level=env_loglevel)
 logg = logging.getLogger(__name__)
 
-VERSION = '1.7.18'  # Remember to bump this in every PR
+VERSION = '1.7.19'  # Remember to bump this in every PR
 
 logg.info('Loading configs at UTC {}'.format(datetime.datetime.utcnow()))
 


michielderoos:
[Scope comments to transfer accounts](https://trello.com/c/5lPuv4bq/670-scope-comments-to-transfer-accounts)

PR #553, opened 2020-11-19
Add URL encoding for dashboard filters

tristanhcole:
**Describe the Pull Request**
Adds URL encoding for dashboard filters. Based off @enjeyw's original encoding function in `utils.js`. 

Example query string: `?credit_transfer.params=public_transfer_type(IN)(PAYMENT)&credit_transfer.group_by=location&credit_transfer.start_date=2020-11-01&credit_transfer.end_date=2020-11-30&user.group_by=ungrouped`

Feedback on URL encoding schema appreciated.

**Todo**


tristanhcole:
[When a page is refreshed, filters should remain](https://trello.com/c/lfzxmVnE/548-when-a-page-is-refreshed-filters-should-remain)

tristanhcole:
Need to think about how to easily "clear all filters" as now you can't just CMD+R to refresh (might want to break into another ticket?)

app/client/utils.js
```diff
@@ -306,6 +319,163 @@ export const processFiltersForQuery = filters => {
...
+        metric_type: mt
+      };
+      store.dispatch(LoadMetricAction.loadMetricRequest({ query: params }));
+      store.dispatch(
+        MessageAction.addMessage({
+          error: true,
+          message: "Something went wrong: " + e.toString()
+        })
+      );
+      buildQueryString(params);
```
tristanhcole, line 396: If the URL is invalid, we want to load _both_ metric types

app/client/__tests__/unit/utils.test.js
```diff
@@ -141,3 +141,105 @@ test("get_zero_filled_values", () => {
     utils.get_zero_filled_values("value", value_array, date_array)
   ).toStrictEqual([11, 0, 1, 0, 0, 1]);
 });
+
+describe("filterEncoding", () => {
```
tristanhcole, line 145: significantly beefs up testing in the utils file

tristanhcole:
Also, I chose the `.` as delimiter as it's similar to how nested objects work in JS and dots are supported https://en.wikipedia.org/wiki/Query_string#URL_encoding

michielderoos (approved):
Tested and looks good!

app/client/utils.js
```diff
@@ -306,6 +319,158 @@ export const processFiltersForQuery = filters => {
   return encoded_filters;
 };
 
+export const parseEncodedParams = (allowedFilters, params) => {
```
enjeyw, line 322: So i could be understanding this wrong, but what I don't love here is that it seems like inputting allowedFilters is being heavily relied upon to do the parsing, when in theory there should be everything we need to parse correctly in the param string.
I'm my mind this increases the complexity of the parser significantly, and makes it less flexible, since we always need to have an opinion on allowedFilters.

The parser becomes pretty small:
```
const parseEncodedParams = (params) => {
  let inequalityMapper = {
    'EQ': '=',
    'LT': '<',
    'GT': '>',
  };

  return params
    .split(":")
    .filter(item => item)
    .map((param, i) => {
      //Tokenize across parentheses
      let [attribute, comparator, value] = param.split(/[()]+/).filter(e => e);

      if (comparator === 'IN') {
        return {
          type: 'discrete',
          attribute: attribute,
          allowedValues: value.split(','),
          id: i+1
        }
      }

      return {
        type: inequalityMapper[comparator] || inequalityMapper['EQ'],
        attribute: attribute,
        threshold: value,
        id: i+1
      }
  })
};
```

I think passing in allowedFilters  does have some merit in using it to validate the param string (ie throwing an error if there's an `EQ` where there's supposed to be an `IN`), but even then, I'd think I think i'd be simpler to write that over two seperate functions, which each one having a specific task: one to parse into a dict, and then one to validate whether the filter is actually valid or not. Right now we just get TypeErrors and the like, when it'd be pretty simple to validate properly.

enjeyw, line 322: So i could be understanding this wrong, but what I don't love here is that it seems like inputting allowedFilters is being heavily relied upon to do the parsing, when in theory there should be everything we need to parse correctly in the param string.
I'm my mind this increases the complexity of the parser significantly, and makes it less flexible, since we always need to have an opinion on allowedFilters.

The parser becomes pretty small (this is just a direct transposition of the python-side version that I threw together):
```
const parseEncodedParams = (params) => {
  let inequalityMapper = {
    'EQ': '=',
    'LT': '<',
    'GT': '>',
  };

  return params
    .split(":")
    .filter(item => item)
    .map((param, i) => {
      //Tokenize across parentheses
      let [attribute, comparator, value] = param.split(/[()]+/).filter(e => e);

      if (comparator === 'IN') {
        return {
          type: 'discrete',
          attribute: attribute,
          allowedValues: value.split(','),
          id: i+1
        }
      }

      return {
        type: inequalityMapper[comparator] || inequalityMapper['EQ'],
        attribute: attribute,
        threshold: value,
        id: i+1
      }
  })
};
```

I think passing in allowedFilters  does have some merit in using it to validate the param string (ie throwing an error if there's an `EQ` where there's supposed to be an `IN`), but even then, I'd think I think i'd be simpler to write that over two seperate functions, which each one having a specific task: one to parse into a dict, and then one to validate whether the filter is actually valid or not. Right now we just get TypeErrors and the like, when it'd be pretty simple to validate properly.

app/client/utils.js
```diff
@@ -306,6 +319,158 @@ export const processFiltersForQuery = filters => {
...
+  } catch (e) {
+    console.log("Something went wrong", e);
+    allowedFilters = null;
+    metric_types.map(mt => {
+      params = {
+        disable_cache: true,
+        group_by: "ungrouped",
+        metric_type: mt
+      };
+      store.dispatch(LoadMetricAction.loadMetricRequest({ query: params }));
```
enjeyw, line 389: feels like a kinda strange place to have side effects to me. Is it unweildly to have the catch somewhere else to keep this function pure?

app/client/utils.js
```diff
@@ -306,6 +319,158 @@ export const processFiltersForQuery = filters => {
...
+      message.error("URL Invalid");
+      buildQueryString(params);
+    });
+  }
+  return filters;
+};
+
+export const flattenObject = obj => {
```
enjeyw, line 397: nice

app/client/utils.js
enjeyw: Edit:

Errrgg just realised that `IN` can also be a boolean mapping (though they're treated the same everyhwere). That's annoying. Adding this still is a lot easier to understand as an overall function - no tripple-nested `if` statements (though just my opinion!)

```
     let matchedFilter = allowedFilters[attribute];
      if (!matchedFilter) {
        throw `Allowed Filter not found for attribute ${attribute}`
      }

      if (comparator === 'IN') {
        return {
          type: matchedFilter.type,
          attribute: attribute,
          allowedValues: value.split(','),
          id: i+1
        }
      }
```

app/client/utils.js
```diff
@@ -89,7 +94,15 @@ export const parseQuery = queryString => {
     var pair = pairs[i].split("=");
     query[decodeURIComponent(pair[0])] = decodeURIComponent(pair[1] || "");
   }
-  return query;
+  if (
```
enjeyw, line 97: i'm unclear what this outer `if` is doing - won't it always be true?

enjeyw (commented):
looks good overall, a few thoughts though!

tristanhcole:
No longer needed

PR #580, opened 2020-12-15
Transfer Card Fetching Improvements

enjeyw:
**Describe the Pull Request**

Adds some improvements to how we fetch transfer cards, allowing for flexibility than a simple bulk get.

**Todo**

- [ ] QA your pull request. This includes running the app, login, testing changes, checking [accessibility](https://www.a11yproject.com/checklist/) etc.


app/server/models/utils.py
```diff
@@ -41,52 +41,6 @@ def get_authorising_user_id():
     else:
         return None
 
-
-def paginate_query(query, queried_object=None, order_override=None):
```
enjeyw, line 45: needs to be moved to avoid an import error

app/test_app/functional/api/test_transfer_card_api.py
```diff
@@ -135,7 +136,7 @@ def bind_user(id, serial_number):
     bind_user(user9.id, 888888)
 
     # Make sure the distance filters are working!
-    distances = [(None, 9), (0, 9), (0.001, 7), (2, 7), (10, 8), (100, 9), (100000, 10)]
```
enjeyw, line 138: actually a bug in the previous code which was picking up an location with none or something (can't remember exactly what, been a while since i wrote this)

app/server/utils/auth.py
```diff
@@ -406,6 +406,8 @@ def create_user_response_object(user, auth_token, message):
         'request_feedback_questions': request_feedback_questions(user),
         'default_feedback_questions': current_app.config['DEFAULT_FEEDBACK_QUESTIONS'],
         'transfer_usages': transfer_usages,
+        'forgiving_deduct': current_app.config['FORGIVING_DEDUCT'],
```
enjeyw, line 409: adding config that turns some of the features associated with this PR on on the mobile app

app/server/models/utils.py
```diff
@@ -289,3 +243,81 @@ def organisations(cls):
...
+
+        paginated = query.paginate(page, per_page, error_out=False)
+
+    if len(paginated.items) > 0:
+        new_last_fetched_obj = paginated.items[-1]
+        new_last_fetched = getattr(new_last_fetched_obj, sort_attribute.key)
+    else:
+        new_last_fetched = None
+
+    return paginated.items, paginated.total, paginated.pages, new_last_fetched
```
michielderoos, line 322: (Willing to be convinced otherwise)  
I get the logic of having a last_fetched input parameter, but I'm not sure if we should return a `new_last_fetched` in the output. Just seems redundant if you can figure out what the `last_fetched` element is from the actual items.

devtools/quick_setup_script.sh
```diff
@@ -129,11 +129,11 @@ echo ~~~Setting up Contracts
 cd ../
 python -u devtools/contract_setup_script.py
 
-if [[ "mockData" != 'none' ]]; then
-    echo ~~~Creating mock data
-    cd ./app/server/utils
-    python -u mock_data.py ${testdata}
-fi
+#if [[ "mockData" != 'none' ]]; then
```
michielderoos, line 132: I know it's broken, but should it be nuked in this PR?

michielderoos (changes requested):
Just a few notes!

devtools/quick_setup_script.sh
enjeyw: oh nope that was a mistake!

app/server/models/utils.py
enjeyw: Few reasons why I think it's good:
- more conformant with restful pagination process of including metadata in the respons. see (https://developer.atlassian.com/server/confluence/pagination-in-the-rest-api/)
- less work for for client side - only needs to be implemented one place on server side, and it means client doesn't need to remember what key is used to identify the last fetched and extract it

app/server/api/credit_transfer_api.py
```diff
@@ -98,7 +98,7 @@ def get(self, credit_transfer_id):
                         or_(CreditTransfer.recipient_transfer_account_id.in_(parsed_transfer_account_ids),
                             CreditTransfer.sender_transfer_account_id.in_(parsed_transfer_account_ids)))
 
-            transfers, total_items, total_pages = paginate_query(query, CreditTransfer)
+            transfers, total_items, total_pages, new_last_fetched = paginate_query(query)
```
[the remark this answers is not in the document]
enjeyw: fair point. went one better and actually just returned it so that the paginations are now all standardised

enjeyw:
[Transfer Card Single Card Get Endpoint](https://trello.com/c/gv78lr2Q/700-transfer-card-single-card-get-endpoint)

app/server/models/user.py
```diff
@@ -407,18 +407,18 @@ def _haversine(lib, selfref, lat, lng):
...
-            or_(
-                and_(User.lat==None, User.lng==None),
-                and_(User.lat==self.lat, User.lng==self.lng),
-                User.great_circle_distance(self.lat, self.lng) < radius,
-                and_(self._location is not None, User._location == self._location)
-            )
-        ).all()
+        return db.session.query(User).filter(self.users_within_radius_filter(radius)).all()
+
+    def users_within_radius_filter(self, radius):
```
michielderoos, line 415: Nice-- this is now easily added to the filters module too (in theory). Neat!

app/server/models/utils.py
michielderoos: I have been convinced otherwise! 👍

michielderoos (approved):
I added a fix to what was breaking tests (was a routing thing-- the test itself was fine). Looks good now!

PR #581, opened 2020-12-15
Search 2.0

michielderoos:
**Describe the Pull Request**

Makes search use trigrams and filters!  
It's also crazy fast, especially if you keep the `per_page` around 20. 
Also much less code, and with that also makes tests much better.  
  
Todo (not in this PR):  
Make it search credit transfers (easy, but less important)  
Make it search custom attributes (harder, but more important)  

**Todo**

- [x] Link relevant [trello card](https://trello.com/b/PA2LhlOh/sempo-dev) or [related issue](https://github.com/teamsempo/SempoBlockchain/issues) to the pull request
- [x] Request review from relevant @person or team
- [x] QA your pull request. This includes running the app, login, testing changes, checking [accessibility](https://www.a11yproject.com/checklist/) etc.
- [x] Bump [backend](https://github.com/teamsempo/SempoBlockchain/blob/e3eb0f480e86d5a8ef2c1814127b70ff018671c4/config.py#L7) and/or [frontend](https://github.com/teamsempo/SempoBlockchain/blob/e3eb0f480e86d5a8ef2c1814127b70ff018671c4/app/package.json#L3) versions following Semver

PR #583, opened 2020-12-17
Revised Bulk Disbursement Spec

michielderoos:
# Bulk Disbursement Spec  

### Requirements  
There are a few base requirements we need to meet for the new disbursement flow
- Ability to review transfer before sending  
- Ability to revise transfer while reviewing  
- Ability to select a beneficiary pool

### Current Issues  
There are a few issues with the current flow which we should address with the new disbursement flow
- With the current flow, it is easy to accidentally disburse twice (by hitting the _Create Disbursement_ button more than once)  
- The HTTP request to create a disbursement can take a long to resolve when sending to a very large number of beneficiaries.  
- With the current flow, it is required that the client has every transfer account loaded. This causes performance issues when there is a large pool of beneficiaries.   

### Considerations  
- In the future, we may want to be able to save a disbursement and modify/execute it again  
- In the future, we may want to schedule disbursements to execute in the future, or allow one disbursement recur on a schedule  

### Backend  
The core of this solution is based on the revised search endpoint (see PR #581). At present, every beneficiary is loaded by the frontend, and to do a bulk disbursement the backend requires a list of each user ID to be disbursed to. With this solution, the goal is to allow the user to narrow down who they want to disburse to using the search and filtering functionality, while maintaining the freedom to manually select specific users who they want to include-only or exclude from the disbursement.   
Below is a table with a sample-set of users, to be used in examples throughout the rest of the document 
  
| ID | Name              | Location      | Gender        |
| -- | ----------------- | ------------- | ------------- |
| 1  | Barney Rubble     | Rock Creek    | Male          |
| 2  | Fred Flinstone    | Bedrock       | Male          |
| 3  | Betty Rubble      | Rock Creek    | Female        |
| 4  | Pebbles Flinstone | Bedrock       | Female        |
| 5  | Wilma Flinstone   | Bedrock       | Female        |

**Phase 1: Part 1**  
**The goal of Phase 1 is to generate a list of users to disburse to.**  

The first thing the backend will see from a user creating a new disbursement is a search query (including a filters object), a list of user IDs, and a flag stating whether 'everyone including' or 'everyone except' the users on that list is being disbursed to. How this plays out is best explained through some simple examples using the sample-set of users seen above.  
Note that these are just sample queries-- there are other permutations of settings which can create the same end-result.   
Example 1:  
An admin wants to **disburse to everyone in Bedrock, except for Pebbles Flinstone**. A Phase 1 disbursement request body might look like this:  
```
{
    // Blank query string will return everyone
    "query": "", 
    // This filter string will remove everyone not from Bedrock
    "filters":  "location(in)(Bedrock)",
    // These two lines mean to exclude user with ID 3
    "users": [3], 
    "except_or_including": "EXCEPT",
}  
```
Example 2:
An admin wants to **disburse to every female Flinstone**. A Phase 1 disbursement request body might look like this:  
```
{
    // Narrow down to people matching search query Flinstone
    "query": "Flinstone", 
    // This filter string will remove every non-female
    "filters":  "gender(in)(Female)",
    // Don't exclude anyone, search and filters already got us down to the users we need!
    "users": [], 
    "except_or_including": "EXCEPT",
}  
```
Example 3A: 
An admin wants to **disburse to only Wilma Flinstone**. A phase 1 disbursement request body might look like this:  
```
{
    // Narrow down to people matching search query Flinstone
    "query": "Flinstone", 
    // No filters in this case!
    "filters":  "",
    // Include only Wilma's ID
    "users": [5], 
    "except_or_including": "INCLUDE_ONLY",
}  
```
One note on this one. The `INCLUDE_ONLY` flag overrides everything else. There _is_ a query string here, but that's just to show that the disbursing party just used the search to find Wilma's name to check off.   

Example 3B: 
Another possible valid payload for this same disbursement request body would be something like this:    
```
{
    // Narrow down to people matching search query Flinstone
    "query": "Flinstone", 
    // No filters in this case!
    "filters":  "",
    // Exclude Fred and Pebbles' IDs 
    "users": [4, 2], 
    "except_or_including": "EXCLUDE",
}  
```
Both of these would result in only Wilma getting disbursed to, but in this case the `query` string _does_ matter. If it were left blank, users 1,3, and 5 would all be disbursed to.   
**Phase 1: Part 2**  
Phase 1 Part 1 generated a list of users, but we haven't sent them any cash yet! We now want to generate a `disbursement` object in the database (in a `PENDING` state), and store the list of transfer accounts associated with that newly created disbursement. A paginated* list of transfer accounts would be returned. With this being created, the list of intended recipients has been cemented. 
That is to say, while the user is building their query, it's possible that new users get added, or existing ones change their attributes to enter or exit the disbursement set criterion. At this point though, the set is represented by a real list of transfer accounts, rather than instructions on how to build that set.  
Note: In the future if we want to make disbursements scheduled or recurring, we might want to add the option to make these objects dynamic instead as well. I.e. as proposed now, if you create a `disbursement` object to send to each Flinstone a disbursement and execute it later on (after Bam-Bam Flinstone joins the platform) only Fred, Wilma, and Pebbles will receive the disbursement.
  
*Page 1 would be returned. Additional pages would be viewed by `GET`ting any given page from a new `disbursement` API endpoint   

**Phase 2**
**The goal of Phase 2 is to have user verify their intention to execute the disbursement**.  
Phase 2 is quite simple. The user who created the disbursement object in Phase 1 essentially just says "Yes" to the list of users it returned. They also have the option to do a simple manipulation to the set (remove users from the list)  
Example 1:  
```
{
    "disbursement_id": 1, 
    "action": "EXECUTE",
    "exclude": [2]
}
```
### Frontend  
From the Accounts or Transfers page, the user clicks the `Build Disbursement` button. This will bring them to a new page, resembling the current `Accounts` page, but revised using antdesign components, and with the updated filters module used by Metrics.
image
- The _Search..._ box will populate the `query` field from _Backend Phase 1_.    
- The _Filters:_ bar will populate the `filters` field from _Backend Phase 1_ with encoded filter strings.   
- The top-most checkbox (next to _STATUS_) will toggle the `except_or_including` field in _Backend Phase 1_.  
  - If the top-most box is **checked**, then `except_or_including` will be `EXCEPT` un-checking users will add them to the `users` field from _Backend Phase 1_.    
  - If the top-most box is **unchecked**, then `except_or_including` will be `INCLUDE_ONLY` checking users will add them to the `users` field from _Backend Phase 1_.    
  

The bottom of the page will have a "Continue" button, which invokes sends the request to the endpoint described in `Phase 1`. That will bring you to a verification screen, similar to this:  
image

### DIFF
--- a/config.py
+++ b/config.py
@@ -5,7 +5,7 @@ env_loglevel = os.environ.get('LOGLEVEL', 'DEBUG')
 logging.basicConfig(level=env_loglevel)
 logg = logging.getLogger(__name__)
 
-VERSION = '1.7.40'  # Remember to bump this in every PR
+VERSION = '1.7.41'  # Remember to bump this in every PR
 
 logg.info('Loading configs at UTC {}'.format(datetime.datetime.utcnow()))
 


enjeyw (commented):
Just for completeness adding the two suggested changes from the initial review:

- rather than having an "except_or_including" flag, just have two lists for users ie `include_users: [..], except_users: [...]`
- don't allow modification during phase two - just require people to restart

tristanhcole:
Closed by #595

PR #600, opened 2021-01-26
Asyncify Spreadsheet Import

michielderoos:
**Describe the Pull Request**
Makes spreadsheet import run in a thread. This does remove `diagnostic`, since it's impossible with this approach. In the future (out of scope here IMO) we should make an `fetch async output` endpoint to fetch results from actions like this. 
  

**Todo**


michielderoos:
[Async Excel Important so that bad things don't happen with 1000 users to import](https://trello.com/c/M7Sbwb8R/726-async-excel-important-so-that-bad-things-dont-happen-with-1000-users-to-import)

app/client/components/uploader/uploadedTable.jsx
```diff
@@ -315,42 +315,14 @@ class uploadedTable extends React.Component {
       );
     }
 
-    if (
-      this.props.saveState.saved &&
-      this.props.saveState.diagnostics.length > 0
-    ) {
+    if (this.props.saveState.saved) {
       var added = 0;
```
enjeyw, line 319: nit-nat-patty-whack: are these vars used?

app/server/models/user.py
```diff
@@ -425,7 +425,7 @@ def users_within_radius_filter(self, radius):
 
     def get_transfer_account_for_organisation(self, organisation):
         for ta in self.transfer_accounts:
-            if ta in organisation.transfer_accounts:
+            if ta.organisation.id == organisation.id:
```
enjeyw, line 428: presumably something was breaking about the old check?

enjeyw (approved):
looks good!

app/server/models/user.py
michielderoos: I just thought it might be faster checking for an ID match than a list membership check

PR #607, opened 2021-02-01
fix: link existing user to org on invite

### LINKED ISSUE #305: Associate existing user to organization when they're invited
**Describe the bug**
User invitation to an organization only works for brand new accounts. Existing accounts don't get updated/associated to the new organization.

**To Reproduce**
1. Create a User A under ORG1
2. Invite User A to ORG2
3. User A will not be able to join ORG2

**Expected behavior**
If a user has an account, invited to an organization they should join that organization.

tristanhcole:
**Describe the Pull Request**
When an existing user is invited they will now be added to the new organisation. Also adds a lot more tests for the permission API. Closes #305 

**Todo**

### DIFF
--- a/app/server/api/auth_api.py
+++ b/app/server/api/auth_api.py
@@ -13,7 +13,8 @@ from server.utils.auth import requires_auth, tfa_logic, show_all, create_user_re
 from server.utils.access_control import AccessControl
 from server.utils import user as UserUtils
 from server.utils.phone import proccess_phone_number
-from server.utils.amazon_ses import send_reset_email, send_activation_email, send_invite_email
+from server.utils.amazon_ses import send_reset_email, send_activation_email, send_invite_email, \
+    send_invite_email_to_existing_user
 from server.utils.misc import decrypt_string, attach_host
 from server.utils.multi_chain import get_chain
 from sqlalchemy.sql import func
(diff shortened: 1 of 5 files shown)


tristanhcole:
[Associate existing user to organization when they're invited](https://trello.com/c/pRUgznwR/289-associate-existing-user-to-organization-when-theyre-invited)

app/server/api/auth_api.py
```diff
@@ -768,12 +769,31 @@ def post(self):
...
-        if email_exists:
-            response_object = {'message': 'Email already on whitelist.'}
+        email_exists_for_org = EmailWhitelist.query.filter(func.lower(EmailWhitelist.email)==email).first()
+        if email_exists_for_org:
+            response_object = {'message': 'Email already on organisation whitelist.'}
             return make_response(jsonify(response_object)), 400
 
+        email_exists = EmailWhitelist.query.filter(func.lower(EmailWhitelist.email)==email)\
+            .execution_options(show_all=True).first()
+        if email_exists and not email_exists.used:
```
enjeyw, line 780: what bad thing happens if we remove this check and a user can be added to a whitelist for more than one org? Presumably it breaks the signup if the user tries to use both invites, but we could fix that. feels better than having one org depend on another org's behaviour (can't resend invites for the user etc etc). Also means it doesn't allow admins discover who's been added to another org.

enjeyw (commented):
Got some thoughts re behaviour when an invite exists but hasn't been used.

app/server/api/auth_api.py
tristanhcole: it won't sign up. the user will be associated to one account. i didn't want to do it as could be complex (need to rewrite part of the auth flow) and burn a day or so on implementing, whereas this was a few hours.

EmailWhitelist is currently OneOrgBase.

enjeyw: hmmm i might take a quick squiz at the auth flow and see how complex it'd be. I think the expected behavor from a ux PoV is that an invite shouldn't be blocked by the presence or lack of one on another org, unless you feel otherwise?

tristanhcole: definitely agree it's currently non-ideal UX - just seems pretty edge case at present. that would be great :)

enjeyw: my opinion is weakened by the fact there's a bunch of other dependencies (tiers, deleting etc), but ideally we'd fix those too

enjeyw (approved):
LGTM

PR #611, opened 2021-02-03
New async flow

michielderoos:
**Describe the Pull Request**

This is a superset of [this](https://github.com/teamsempo/SempoBlockchain/pull/595) PR, so merge first for a tidier diff.  
This redoes how we do async tasks, and allows you to get results from ongoing tasks as they progress! For reviewing, I'd read `executor.py` first, and then all the unit tests touched by this PR. Let me know if you have any questions, because this is pretty hairy!

**Todo**


app/server/utils/user.py
```diff
@@ -638,7 +636,7 @@ def proccess_create_or_modify_user_request(
     user = create_transfer_account_user(
         first_name=first_name, last_name=last_name, preferred_language=preferred_language,
         phone=phone, email=email, public_serial_number=public_serial_number, uuid=uuid,
-        organisation=organisation,
+        organisation=organisation if organisation else g.active_organisation,
```
michielderoos, line 639: This being missing was causing a pretty major bug. Maybe this is important enough to be a separate PR?  
Basically what was going on was that `get_transfer_account_for_organisation` would run later in the user-creation flow, but no account would come up because the account wouldn't belong to the org making the account.

app/server/utils/user.py
```diff
@@ -664,7 +662,6 @@ def proccess_create_or_modify_user_request(
     if is_self_sign_up and attribute_dict.get('deviceInfo', None) is not None:
         save_device_info(device_info=attribute_dict.get(
             'deviceInfo'), user=user)
-    send_onboarding_sms_messages(user)
```
michielderoos, line 667: This was being run twice?

app/server/__init__.py
```diff
@@ -270,7 +267,7 @@ class AppQuery(BaseQuery):
 db = SQLAlchemy(
     query_class=AppQuery,
     session_options={
-        "expire_on_commit": not config.IS_TEST,
+        "expire_on_commit": False,
```
michielderoos, line 270: Having this be True blows up threads sometimes

app/server/models/credit_transfer.py
```diff
@@ -179,11 +187,10 @@ def _get_required_prior_tasks(self):
         For why this works, see https://github.com/teamsempo/SempoBlockchain/pull/262
 
         """
-
         # We're constantly querying complete transfers here. Lazy and DRY
         complete_transfer_base_query = (
             CreditTransfer.query.filter(CreditTransfer.transfer_status == TransferStatusEnum.COMPLETE)
-        )
+        ).execution_options(show_all=True)
```
michielderoos, line 193: I realized that running `_get_required_prior_tasks` before the end of the request was MUCH slower than after. Turns out that after request,  the `before_compile` event listener doesn't listen any more, and that it was slowing these particular queries WAY down before adding these `execution_options`.

michielderoos (commented):
Big comment on perf

app/server/api/disbursement_api.py
```diff
@@ -0,0 +1,203 @@
...
+        if not disbursement:
+            return { 'message': f'Disbursement with ID \'{disbursement_id}\' not found' }
+
+        if disbursement.is_executed == True:
+            return { 'message': f'Disbursement with ID \'{disbursement_id}\' has already been executed!' }
+
+        @status_checkable_executor_job
+        def trigger_jobs(transfers):
+            from server.models.credit_transfer import CreditTransfer
+            # Disabled batch_uuid, since executing two sequential bulk disbursements is unacceptably slow
```
michielderoos, line 172: Just making a note on this! I realize being slow might not be a dealbreaker since it's an async task, but it's _really really_ slow.

michielderoos:
[Redo async structure and real-time request monitoring](https://trello.com/c/JZZ6knGJ/742-redo-async-structure-and-real-time-request-monitoring)

enjeyw (approved):
Just realised i never actually clicked "approve" on this!

PR #679, opened 2021-04-23
feat: header with breadcrumbs

tristanhcole:
**Describe the Pull Request**
Adds a header component with a back button and breadcrumbs. Automatically builds based on paths. Also has the capacity for custom routes to be passed from the nav.jsx component.

I'm unsure if we should just have breadcrumbs, or a title and back button also. See images below.
Screen Shot 2021-04-23 at 4 22 23 pm

Screen Shot 2021-04-23 at 4 21 59 pm

**Todo**

- ~~[ ] Add release notes to [unreleased changelog](https://github.com/teamsempo/SempoBlockchain/blob/master/CHANGELOG.md#unreleased)~~


tristanhcole:
[[Medium] (8 hrs) Several screens would benefit from a “cancel/exit/back” button - Breadcrumbs](https://trello.com/c/YZwibMFf/779-medium-8-hrs-several-screens-would-benefit-from-a-cancel-exit-back-button-breadcrumbs)

michielderoos (changes requested):
Mostly looks pretty good! Pretty slick library! Just a few notes and suggestions:  
image
When clicking the "users" breadcrumb from a user page, it takes you to `/users` (which 404s). Should the parent of a user also be a transfer account? (I.e. `Home/Accounts/Single Transfer Account 4701/Single User 4738 `?  
  
Not a technical note, but I'm also not really sure I like the naming scheme-- a title like `Single User 4738` seems awkward as a large header when you could use either `User` or `John Smith` IMO.

tristanhcole:
> Mostly looks pretty good! Pretty slick! Just a few notes and suggestions:
> image
> When clicking the "users" breadcrumb from a user page, it takes you to `/users` (which 404s). Should the parent of a user also be a transfer account? (I.e. `Home/Accounts/Single Transfer Account 4701/Single User 4738 `?

Agree with this in practice, but I think the breadcrumbs should follow URL nav as much as possible. This is hard to implement and out of scope of current implementation IMO. 

> Not a technical note, but I'm also not really sure I like the naming scheme-- a title like `Single User 4738` seems awkward as a large header when you could use either `User` or `John Smith` IMO.

I've changed to just the Singular noun, i.e. `User`. Hard to implement the specific user name as `Page.tsx` is a parent component of the page where the details are loaded into Redux.

tristanhcole:
> > Mostly looks pretty good! Pretty slick! Just a few notes and suggestions:
> > image
> > When clicking the "users" breadcrumb from a user page, it takes you to `/users` (which 404s). Should the parent of a user also be a transfer account? (I.e. `Home/Accounts/Single Transfer Account 4701/Single User 4738 `?
> 
> Agree with this in practice, but I think the breadcrumbs should follow URL nav as much as possible. This is hard to implement and out of scope of current implementation IMO. In the meantime, I've changed it so it doesn't go to an error 404 page at least.

As I think on this more, I think you are right on this. I'm just not sure if it should be in scope of the ticket or not 🤔

michielderoos:
> As I think on this more, I think you are right on this. I'm just not sure if it should be in scope of the ticket or not

@tristanhcole I think it would be, just because `Users` just takes you to the metrics page which I think is a bit confusing.

michielderoos (commented):
Just a couple things-- not sure `Users` should go to the metrics page (per the other comment), and I think it would make sense to rename `Bulk` to `Bulk Disbursements` here:
image  
Looks good otherwise!

michielderoos (approved):
Looks great!

PR #694, opened 2021-05-10
Logout button will use the logout API

michielderoos:
**Describe the Pull Request**

Logout button was just nuking the cookie instead of also hitting the logout API (which invalidates the JWT)

**Todo**


michielderoos:
[Fix logout button](https://trello.com/c/Va1fvchy/832-fix-logout-button)

tristanhcole (changes requested):
Code makes sense but I'm getting a `Status Code: 401 UNAUTHORIZED` when I log out and token isn't being added to the table `blacklist_tokens`
Screen Shot 2021-05-12 at 10 03 32 am

app/server/api/auth_api.py
```diff
@@ -480,6 +480,7 @@ def post(self):
             auth_token = auth_header.split(" ")[0]
         else:
             auth_token = ''
+        auth_token = auth_header.split("|")[0]
```
michielderoos, line 483: Note the BE fix here-- I think the logic here predates our TFA tokens

tristanhcole: that makes sense

app/client/sagas/authSagas.ts
```diff
@@ -177,6 +178,13 @@ function* saveOrgId(
 function* watchSaveOrgId() {
   yield takeEvery(LoginActionTypes.UPDATE_ACTIVE_ORG, saveOrgId);
 }
+
+export function* apiLogout() {
+  yield call(logoutAPI);
+  yield call(removeSessionToken);
+  yield call(removeOrgIds);
```
tristanhcole, line 185: Can you explain the reasoning for why there's a separate `apiLogout` and `logout` function?

tristanhcole (commented):
One small question that would be good to have an answer for, otherwise tested and make sense to me

app/client/sagas/authSagas.ts
michielderoos: Two reasons! One is that we don't necessarily want to blacklist every time we flush the session token (namely when we're refreshing the token, we don't want to invalidate the token we're about to refresh with).   
The other is that we sometimes call `logout` with no token at all, or with an expired token (Like when you go directly to the `/login` page). These cases would just fail and there's no use wasting an API call on them IMO

tristanhcole: That makes sense

tristanhcole (approved):
👍

PR #702, opened 2021-05-19
Send a 400 on limit error

michielderoos:
**Describe the Pull Request**

When the  POST `/me/credit_transfer/` endpoint (which mobile app uses) hits a limit error, it was uncaught and result in a 5xx error.  
This changes it so it _is_ caught, throws a relevant message, and commits the database. The (desired) side-effect of this is that transfers that formerly went lost to the sands of time will now enter the system (and appear as Resolved as Rejected)   
image.  
Note: Transfers which get 4xx errors show as successfully synchronized in the app. This is technically accurate, and maybe the best we can do until we introduce new limit code.

**Todo**


app/server/me_api/credit_transfer.py
```diff
@@ -338,6 +339,22 @@ def post(self):
             }
             return make_response(jsonify(response_object)), 400
 
+        except TransferAmountLimitError as e:
+            db.session.commit()
+            response_object = {
+                'message': "Account limit reached",
```
tristanhcole, line 345: Maybe providing the specific limit in the message would be useful? Not strongly opinionated though.

tristanhcole (approved):
Makes sense to me! Transfers are no longer lost in 🏜️ ⏳ 

I agree that the best course of action right now is to ensure we are capturing the errors until the new mobile limit code is implemented.

PR #815, opened 2022-05-17
Add processing status to disbursement

michielderoos:
**Describe the Pull Request**
Adds a processing status to indicate if all of the transfers are made yet or not!
image  

**Todo**


tristanhcole (commented):
Tested and makes sense to me.

A question that comes to mind is that there is currently no error state. I see that there is an exception statement, but because it's a loop, it could be hit more than once.

What are your thoughts on handling any errors? I.e. a disbursement that has an error in `disbursement.errors`, or a disbursement that doesn't complete for another reason (e.g. what happened to uganda). Right now I don't think the second scenario would be picked up??

https://github.com/teamsempo/SempoBlockchain/blob/5705b8bf68fbc7e2837ccc8e0dbd89491387cd68/app/server/api/disbursement_api.py#L71-L72

michielderoos:
> Tested and makes sense to me.
> 
> A question that comes to mind is that there is currently no error state. I see that there is an exception statement, but because it's a loop, it could be hit more than once.
> 
> What are your thoughts on handling any errors? I.e. a disbursement that has an error in `disbursement.errors`, or a disbursement that doesn't complete for another reason (e.g. what happened to uganda). Right now I don't think the second scenario would be picked up??
> 
> https://github.com/teamsempo/SempoBlockchain/blob/5705b8bf68fbc7e2837ccc8e0dbd89491387cd68/app/server/api/disbursement_api.py#L71-L72

@tristanhcole I think having errors in `disbursement.errors`, but still being complete is fine though. It's "complete with errors", but still finished, and you can see those errors on the disbursement page. I think it's possible that we can figure out a way to detect if it doesn't complete (like if the system shuts down, or is otherwise preempted before it can be completed), but that might be a new PR. The idea of this one for me was basically just to have a quick way to see if it has finished yet, since previously it was a matter of going into the transfer list and seeing if new transfers are still coming in from the disbursement.

tristanhcole (approved):

